PR #1, opened 2015-11-19
Fixed Upload expressions not saving to storage

jkokosar:


tools/expressionmerge.py
```diff
@@ -35,7 +36,7 @@
     base, ext = os.path.splitext(f)
     delimiter = ';' if ext == '.csv' else '\t'
 
-    with gzip.open(f, 'rb') as csvfile:
+    with io.TextIOWrapper(io.BufferedReader(gzip.open(f))) as csvfile:
```
mstajdohar: @kostko, is there a better solution? The problem is that users upload .csv files with windows (/r/n), linux (/n) or mac (/r) line endings and that simply iterating through the file (gzip.open) fails for .csv files generated with Excel on Macs.

tests/test_upload.py
```diff
@@ -41,6 +41,7 @@ def test_upload_expression(self):
         expressions_3 = self.run_processor("import:upload:expression", inputs)
         self.assertFiles(expressions_3, "rc", "00Hr_rc.tab.gz")
         self.assertFiles(expressions_3, 'exp', '00Hr_tpm.tab.gz')
+        self.assertJSON(expressions_3, expressions_3.output['exp_json'], '', 'expression_2.json.gz')
```
mstajdohar: @jkokosar, you should test the three different line endings. We should prepare three small expression test files.

tools/expressionmerge.py
mstajdohar: @kostko, would you suggest a quick linux command that checks line endings and convert them from win or mac to a linux style? One option would be to do this for all files after users upload them to the platform.

mstajdohar: @jkokosar, if we used Python's open command, we could specify the universal (U) flag. The gzip.open does not support the universal flag. Maybe something like this would work as well but look nicer `open(gzip.open(f, 'rb'), 'Ur')`.

jkokosar: @mstajdohar The last suggestion does't work.

I found the committed solution as a workaround for the fact that the gzip.open() does not support universal new line mode.

mstajdohar: Does `io.TextIOWrapper(io.BufferedReader(...)` impact performance? Can we use this for really large files as well? Maybe we should write a wrapper function for `io.TextIOWrapper(io.BufferedReader(gzip.open(f)))` and replace the `gzip.open` in all processes that use that.

kostko: > @kostko, is there a better way? The problem is that users upload .csv files with windows (/r/n), linux (/n) or mac (/r) line endings and that simply iterating through the file (gzip.open) fails for .csv files generated with Excel on Macs.

What do you mean that simply iterating through the file fails? You mean that the CSV parser does not handle different line endings properly?

> @kostko, would you suggest a quick linux command that checks line endings and convert them from win or mac to a linux style? One option would be to do this for all files after users upload them to the platform.

There is the `dos2unix`, which does this conversion.

> Does io.TextIOWrapper(io.BufferedReader(...)) impact performance? Can we use this for really large files as well? 

There certainly is some overhead, but since it is a stream interface, it might be very low. The best way would be to benchmark its impact with some large files using a simple Python script.

mstajdohar: I suppose dos2unit converts \r\n to \n, not \r to \n? We would just have to try.

kostko: It can do both (`-c mac`), but AFAIK not at the same time. We should first check the overhead of the wrappers and only consider other solutions if it is really too high on huge files.

mstajdohar: A quick test showed the buffers actually improve the performance, for almost 2-folds. Could not believe it, so i googled and found [this](http://aripollak.com/pythongzipbenchmarks/). It seems that there might be even better options (with popen and zcat), but I suggest we start by creating a function `gopen(file)` in `utils.py` that returns `io.TextIOWrapper(io.BufferedReader(gzip.open(file)))`. The we should replace all uses of gzip.open with our function. In the future we can improve the function with `zcat`.

tjanez: Very interesting. Yes, we should definitely introduce a utility function (`utils.gzopen`) and use it everywhere. Indeed, `io.BufferedReader` looks very nice performance wise.

tjanez:
@jkokosar, please manually copy the changed files and create a new pull request (GitHub doesn't let us update and reopen this pull request).

PR #3, opened 2015-11-30
Fix tests

mstajdohar:


mstajdohar:
@dblenkus, you should probably review and accept this before you start working on the integration with resolwe.

PR #8, opened 2015-12-11
Feature testintegration

dblenkus:


resolwe_bio/tests/utils.py
```diff
@@ -1,380 +1,10 @@
-"""
-.. autoclass:: server.tests.processor.utils.PreparedData
-   :members:
+try:
+    from resolwe.flow.tests import ProcessTestCase
+except ImportError:
+    from .utils_genesis import ProcessTestCase
```
mstajdohar: I think you suggested this already, move utils_genesis to genesis. Tadej and I also agree on this. The import should then be:
```
from server.tests.utils import ProcessTestCase
# or
from server.tests.utils import ProcessorTestCase as ProcessTestCase
# or
from genesis.server.tests.utils import ProcessorTestCase as ProcessTestCase
```

mstajdohar:
Jenkins, test this please.

resolwe_bio/tests/test_processor.py
```diff
@@ -1,12 +1,15 @@
 # pylint: disable=missing-docstring
 from django.test import TestCase
 
-from server.models import Processor, iterate_schema
+try:
```
mstajdohar: I do not like this. I suggest to implement a ```get_processes``` method in utils or a similar solution. Utils.py should be the only file with references to genesis, so that we have more control and that it will be easy to remove it.

mstajdohar: Instead of ```get_processes``` we can have this try/catch in ```utils.py``` and then import the Process from ```utils.py``` directly here.

resolwe_bio/tests/test_upload.py
```diff
@@ -14,7 +13,14 @@ def test_bam_upload(self):
         self.assertFiles(upload_bam, 'bai', 'alignment_bam_upload_index.bai')
 
         inputs = {"src": "alignment_position_sorted.bam", "src2": "alignment_bam_upload_index.bai"}
-        self.assertRaises(mongoengine.ValidationError, self.run_processor, "import:upload:mapping-bam-indexed", inputs)
+        try:
+            import mongoengine
+            self.assertRaises(
+                mongoengine.ValidationError, self.run_processor, "import:upload:mapping-bam-indexed", inputs)
```
mstajdohar: ```from .utils import ValidationError```

that can either be a ```mongoengine.ValidationError``` or a ```django.core.exceptions.ValidationError``` or some other validation error that we will use for validating the postgres model.

mstajdohar:
Tests fail because ```/files``` are not found. ```self.current_path``` in ```server/tests/utils.py``` points to the ```server/tests``` instead to ```genpackages/resolwe-bio/resolwe_bio/tests```. Do not know how to solve this quickly. When you run tests in ```server/tests```, the ```/files``` path should point there, but for tests in ```genpackages/resolwe-bio/resolwe_bio/tests``` the ```/files``` path should point there...

PR #14, opened 2015-12-23
Fix annotation data type and some bordering cases

mstajdohar:
The progress from cURL download from the tranSMART instance does not work yet. I suggest we accept this PR as is and add this feature later.

### DIFF
--- a/resolwe_bio/processes/transmart/import.yml
+++ b/resolwe_bio/processes/transmart/import.yml
@@ -6,9 +6,9 @@
 ---
 
 - name: import:web:transmart:expressions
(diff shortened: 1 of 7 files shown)


dblenkus:
It looks like that the values that are passed to `update` function are not `current` and `total` as you named them, but `down` and `up`.
I don't know what is the first number of `down`, but the second one is the download progress. 'up' is probably `NA` because there is no upload.
We have to get file size somehow, but I will have to check how.
https://github.com/transmart/RInterface/blob/master/R/RClientConnectionManager.R#L298

PR #18, opened 2015-12-30
Fix merging columns of different types

mstajdohar:
We merge attributes that were binarized by tranSMART into one attribute. Fox example, tranSMART creates 2 attribtes for sex, _i.e._ ```sex_M``` and ```sex_F``` with values ```{'M' or ''}``` and ```{'F' or ''}```, respectively. Sometimes those binary attributes are of different types. For example, ```Non.Allergic.Rhinitis.Active_0``` and ```Non.Allergic.Rhinitis.Active_NA``` have values ```{-1 or 0}``` and ```{'NA' or ''}```, respectively.

This fix converts such mixed binary attributes to strings before they are to be merged.

### DIFF
--- a/resolwe_bio/tools/transmart_utils.py
+++ b/resolwe_bio/tools/transmart_utils.py
@@ -74,7 +74,7 @@ def format_annotations(annfile):
     # create var template
     var_template = []
     attrs_final = sorted(attrs_final)
-    attrs_final_keys = map(utils.escape_mongokey, attrs_final)
+    attrs_final_keys = map(lambda x: utils.escape_mongokey(x).encode('unicode_escape'), attrs_final)
     dtype = dict(annp.dtype.descr)
     dtype_final = []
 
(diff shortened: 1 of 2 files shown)


mstajdohar:
Btw, it is very strange to have this two attributes saved as different types on the tranSMART Platform. Maybe we did something wrong at the data import.

mstajdohar:
Our code probably fails if you select binary attributes of int type. There was no such example to test. Are there any binary attributes of int type? Would this make sense at all? For instance, the attribute Gerd.Diagnosed has floats, but saves them as binarized strings. Two things to note here: binarizing a float attribute is non-sense, there probably are no integer or float binarized attributes, because we converted them to strings on import.

nkran:
I'll have to check how tranSMART saves data on importing and if there are other ways to specify types on  import. We were a bit in a hurry, so I didn't go so much into details.

PR #23, opened 2016-01-15
Add new settings

dblenkus:


tests/.data/README.rst
```diff
@@ -0,0 +1,2 @@
+Resolwe doesn't create `data` and `upload` directories automatically so
```
mstajdohar: We should probably add to Resolwe this 2 dirs and write test for a basic:file field on the process input.

mstajdohar: I'll accept this because we probably want an upload and data dirs here even if we add them to resolwe later.

PR #25, opened 2016-01-19
Re-enabled processor test (upload reads with old QC encoding)

jkokosar:


resolwe_bio/tools/transmart_fetch.R
```diff
@@ -3,7 +3,7 @@ require("transmartRClient")
 require('argparse')
 
 parser = ArgumentParser(description='Fetch data from tranSMART instance.')
-parser$add_argument('--URL', default = "http://52.48.28.218:8080/transmart", help='tranSMART instance URL')
+parser$add_argument('--URL', default = "http://137.117.169.225:8080/transmart", help='tranSMART instance URL')
```
mstajdohar: Is the IP correct?

dblenkus: Yes

PR #27, opened 2016-01-28
Documentation

mstajdohar:


mstajdohar:
We should probably write a [Sphinx extension](http://www.sphinx-doc.org/en/stable/extdev/index.html#dev-extensions) for generating the process catalog.

mstajdohar:
@jkokosar, would you think about pipeline examples?

mstajdohar:
@dblenkus, could you review this quickly? We would need to publish this docs together with the deploy.

dblenkus:
Rebased and merged

PR #38, opened 2016-04-06
Added MACS 1.42 processor

jkokosar:


mstajdohar:
Tests pass but they shouldn't.

```Creating test database for alias 'default'...
    VALIDATION ERROR: transmart
        path:       deque([])
        message:    u'var' is a required property
        validator:  required
        val. value: [u'name', u'label', u'description', u'var']
    VALIDATION ERROR: mus:bcm:kompexpress
        path:       deque([])
        message:    u'var' is a required property
        validator:  required
        val. value: [u'name', u'label', u'description', u'var']
    VALIDATION ERROR: microarray:affy:rochester
        path:       deque([])
        message:    u'var' is a required property
        validator:  required
        val. value: [u'name', u'label', u'description', u'var']
    VALIDATION ERROR: reads
        path:       deque([])
        message:    u'var' is a required property
        validator:  required
        val. value: [u'name', u'label', u'description', u'var']
    VALIDATION ERROR: dictyostelium:bcm
        path:       deque([])
        message:    u'var' is a required property
        validator:  required
        val. value: [u'name', u'label', u'description', u'var']
    VALIDATION ERROR: dictyostelium:bcm:etc
        path:       deque([])
        message:    u'var' is a required property
        validator:  required
        val. value: [u'name', u'label', u'description', u'var']
    VALIDATION ERROR: sample
        path:       deque([])
        message:    u'var' is a required property
        validator:  required
        val. value: [u'name', u'label', u'description', u'var']
    VALIDATION ERROR: assembler:abyss
        path:       deque([])
        message:    u'var' is a required property
        validator:  required
        val. value: [u'name', u'label', u'description', u'var']```

Add a test that fails for this error and then fix the error. You probably have to make the `var` field not required in the process schema on the old platform. @dblenkus can help you with that.

**Comment by mstajdohar** [comment]
This does not work on the Genesis yet. Will we add MACS 1.4 to Genesis or add support to specify some processes and tests are "Resolwe only". @tjanez?

**Comment by tjanez** [comment]
@mstajdohar, I think it will become too much of a burden to also support all new processors on the legacy platform (Genesis).

It is easy to detect whether tests are run on Resolwe or the legacy platform, so we can just skip the new tests on the legacy platform.

Also, in the mean time, while all the current tests don't work on Resolwe yet, we can skip them and then gradually enable them. This will enable us to run CI tests for Resolwe Bioinformatics running on top of Resolwe.

File: docs/CHANGELOG.rst [hunk 1]
```diff
@@ -4,6 +4,19 @@ Change Log
 
 All notable changes to this project are documented in this file.
 
+===========
+Unreleased:
```
tjanez: Please, remove `:`

docs/CHANGELOG.rst
```diff
@@ -4,6 +4,19 @@ Change Log
...
+===========
+
+Changed
```
tjanez: I would prefer if you deleted unused sections (i.e. *Changed* and *Fixed*) and add them if/when they are necessary. Similar to what I've done in genialis/resolwe#49.

docs/CHANGELOG.rst
```diff
@@ -4,6 +4,19 @@ Change Log
...
+-------
+
+Fixed
+-----
+
+Added
+-----
+* MACS (v1.42) processor
```
tjanez: What about just `MACS14 processor` or `Processor for MACS14`? I'm guessing it is not tied to a specific version (v1.4.2) but a the whole series (v1.4.x).

resolwe_bio/tests/test_chipseq.py
```diff
@@ -1,7 +1,7 @@
 # pylint: disable=missing-docstring
+import unittest
 from .utils import BioProcessTestCase
 
-
```
tjanez: Please, don't delete this line, it is required by [PEP8](https://www.python.org/dev/peps/pep-0008/#blank-lines).

mstajdohar:
@tjanez would be great of you fixed and accepted this before sprint planning. Janez will not have time.

mstajdohar:
@tjanez, @dblenkus, how can we specify which processors and tests are for the new platform only? If we do not wish to make the new code ugly, we have to list them somewhere in the old platform (the genialis-genesis repository) and then ignore them in register and testing.

jkokosar:
@tjanez I have updated the PR based on your suggestions. The test is skipped for now. The test doesn't check the output file, since it would require a set of large input files.

tjanez:
@jkokosar, thanks for fixing my comments!

PR #40, opened 2016-04-12
Add initial Tox configuration and a couple of fixes

tjanez:
It would be great if we could include this before the next release.

At least the commit that ensures all source files and supplementary package data are in source distribution.


setup.py
```diff
@@ -30,8 +30,16 @@
...
+                'descriptors/*.yml',
+                'processes/**/*.yml',
+                'tests/files/*',
+                'tools/*.py',
+                'tools/*.R',
+            ]
+        },
         zip_safe=False,
         install_requires=(
             "resolwe>=1.0.0",
```
mstajdohar: Use single quotes.

docs/CHANGELOG.rst
```diff
@@ -4,14 +4,26 @@ Change Log
...
+-------
+* Removed obsolete utilities superseded by resolwe-runtime-utils
+
+Fixed
+-----
+* Included all source files and supplementary package data in sdist
+
 Added
 -----
 * "flow_collection: sample" to processes
```
mstajdohar: Maybe also fix this to ``, since you're at it.

setup.py
tjanez: Hmm, I'm using single quotes?

Do you want me to change all double-quotes in `setup.py` to single quotes? If yes, I would rather do it in a separate commit.

mstajdohar: "resolwe>=1.0.0"

mstajdohar: Oh, I see. Not marked green, sorry.

mstajdohar: So, the general rule is that text printed to the user should be in double quotes and the other in single. You can fix the whole `setup.py` if you wish, but not urgent.

mstajdohar:
LGTM

docs/CHANGELOG.rst
tjanez: I would rather do it tomorrow when we'll prepare a new release and as part of preparing the release, we will also clean up/prettify the CHANGELOG.

The reason I'm against it is that I don't want to fix unrelated things in commits. Making a separate commit for this change seems overkill if the release is so close.

setup.py
tjanez: Ok, will do.

mstajdohar:
LGTM

PR #41, opened 2016-04-12
Add ROSE2 processor

jkokosar:
Requires bamliquidator and ROSE2 tools to be added to the Docker image.


resolwe_bio/tests/test_chipseq.py
```diff
@@ -46,3 +46,23 @@ def test_macs14(self):
...
+        control = self.run_processor("import:upload:mapping-bam", inputs)
+
+        inputs = {'src': 'peaks.bed'}
+        macs_peaks = self.run_processor('import:upload:bed', inputs)
+
+        inputs = {"g": 'HG19',
+                  "i_upload":  macs_peaks.id,
+                  "r": bam.id,
+                  "c": control.id,
+                  "t": 2500,
```
jkokosar: Enters processor test mode. It runs rose2 (and bamliquidator dependancy) without any parameters, to check if they are contained in the runtime. Running the actual analysis without crashing would require test input data in the range of 150MB, so this is not feasible.

PR #43, opened 2016-04-13
Renamed MACS14 processor; fix for sample descriptor

jkokosar:


tjanez:
@jkokosar, please create a separate commit for the second unrelated change (fix for sample descriptor). Also, is the fix worthy of a CHANGELOG entry?

PR #46, opened 2016-04-15
Change sample viewset to filter based on query params

### LINKED ISSUE #39: Sample API resource is inconsistent with the rest of the API
It would be more consistent with the rest of the API if the `Sample` resource was accessible via `/api/sample` and `annotated` would be a filter query argument. So you would then do queries:
* `/api/sample?annotated=1` instead of `/api/sample/annotated`
* `/api/sample?annotated=0` instead of `/api/sample/unannotated`

This would make it more consistent for use in the GenJs frontend API.

dblenkus:
Fixes #39

### DIFF
--- a/resolwe_bio/api/views.py
+++ b/resolwe_bio/api/views.py
@@ -1,35 +1,25 @@
-from django.db.models import Max, Q
+from __future__ import absolute_import, division, print_function, unicode_literals
 
-from rest_framework.decorators import list_route
-from rest_framework.response import Response
+from django.db.models import Max, Q
 
 from resolwe.flow.models import Collection
 from resolwe.flow.views import CollectionViewSet
 
 
 class SampleViewSet(CollectionViewSet):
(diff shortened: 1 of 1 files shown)


resolwe_bio/api/views.py
```diff
@@ -1,35 +1,25 @@
...
-        ).filter(
-            ~Q(descriptor__has_key='geo') | ~Q(descriptor__geo__has_key='annotator')
-        ).order_by('-latest_date')
-        serializer = self.serializer_class(queryset, many=True, context={'request': request})
-
-        return Response(serializer.data)
+
+    def get_queryset(self):
+        queryset = Collection.objects.annotate(
+                latest_date=Max('data__created')
```
kostko: Strange indentation?

kostko:
One minor style fix, otherwise LGTM.

resolwe_bio/api/views.py
dblenkus: Thanks, fixed.

PR #47, opened 2016-04-17
Fix Rose2 process

dblenkus:


resolwe_bio/processes/chip_seq/rose2.yml
```diff
@@ -16,7 +16,7 @@
     - name: g
       label: Genome
       type: basic:string
-      required: false
+      required: true
```
jkokosar: I think you can just delete this line. If you omit the "required" field, it will default to "required: true"

jkokosar:
@dblenkus Ready to accept.

resolwe_bio/processes/chip_seq/rose2.yml
dblenkus: Fixed.

PR #49, opened 2016-04-21
Update vc-gatk-joint processor

jkokosar:
Requires change in how the Picard tool is called in Genesis.


tjanez:
Please, close this PR and create a new against the new `legacy` branch

PR #52, opened 2016-05-05
Add Sample

dblenkus:


resolwe_bio/__init__.py
```diff
@@ -0,0 +1 @@
+default_app_config = 'apps.BaseConfig'  # pylint: disable=invalid-name
```
dblenkus: @kostko: Is this the right path or should it be `resolwe-bio.apps.BaseConfig`?

kostko: No, it should be `resolwe_bio.apps.BaseConfig`.

tjanez:
Please, rebase and add a CHANGELOG entry

resolwe_bio/signals.py
```diff
@@ -0,0 +1,47 @@
...
+
+from guardian import shortcuts
+
+from resolwe.flow.models import Data, DescriptorSchema, iterate_schema
+from .models import Sample
+
+
+@receiver(post_save, sender=Data)
+def add_post_save_handler(sender, instance, **kwargs):
+    if kwargs['created'] and instance.process.flow_collection:
```
mstajdohar: In the future we should replace this parameter with a more general group of parameters for Resolwe extensions. For example, `instance.process.ext.bio.sample`.

mstajdohar:
LGTM, @dblenkus please accept.

PR #58, opened 2016-05-13
API endpoint for adding Samples to Collections

dblenkus:


mstajdohar:
@nkran, did you test it in practice, does it work?

mstajdohar:
LGTM

PR #59, opened 2016-05-13
Change default PostgreSQL port to avoid a collision with Resolwe

tjanez:
Change default PostgreSQL port to 55433 to avoid a collision with
Resolwe's default PostgreSQL port. If a developer wants to use
docker-compose with each project's default configuration, this is now
possible since the projects' docker-compose settings don't interfere
anymore.


mstajdohar:
LGTM

PR #65, opened 2016-05-20
Refactor processor syntax

jkokosar:
* `name` -> `slug`
* `label` -> `name`
* add `data_name` field
* remove `var`and `static`fields


mstajdohar:
LGTM

PR #75, opened 2016-06-21
Fix 'generate_samples' django-admin command

jkokosar:


mstajdohar:
Use `string.lower`

resolwe_bio/management/commands/generate_samples.py
```diff
@@ -264,9 +271,9 @@ def create_data(self, reads_name='seq_reads', annotated=False, rseed=None):
             sample.descriptor = self.generate_sample_desciptor(d.name)
             sample.presample = False
             sample.save()
-            print "Created sample: {} (id={})".format(sample.name, sample.id)
+            self.stdout.write("Created sample: {} (id={})".format(sample.name, sample.id))
```
mstajdohar: Why did you change?

mstajdohar: You could include `from __future__ import absolute_import, division, print_function, unicode_literals` in the first line and then just use print(...) for Py2 and Py3. Even better would be to use `logger.info`. See and [example of logger usage here](https://github.com/genialis/resolwe/blob/a7a6b1eb22d61dfc385735f997dcaa920676e1f4/resolwe/flow/executors/__init__.py). Domen can tell you more about logger and logging.

mstajdohar:
Should we setup test coverage on codecov.io?

resolwe_bio/management/commands/generate_samples.py
jkokosar: Changed by following Django docs. Output can now be tested and compared by call_command() stringIO redirection.

mstajdohar: I suggest you change this to `logger.info` and check directly in the DB. For instance, you can check if 1 sample was added to the Sample model after you call the command and maybe even check its name. You could check if the sample has the required data objects in the Data model as well. @dblenkus can you comment if this makes sense?

mstajdohar: Also, maybe just writing "XXX samples successfully created" would be more appropriate then a line for each sample. This gets ugly and slow when you create tens of thousands of test sample data. Test it and decide.

mstajdohar: If you use `logger.info` or even `logger.debug` this might not be necessary because it is relatively easy to suppress the output.

docs/CHANGELOG.rst
```diff
@@ -43,6 +43,7 @@ Added
 * Add docker-compose command for PostgreSQL
 * API endpoint for adding ``Samples`` to ``Collections``
 * HISAT2 aligner
+* Test for ``generate_samples`` django-management command
```
tjanez: Change "django-management" to "django-admin" to be consistent with the previous change log entry.

tjanez:
@jkokosar, please improve the commit message. Try to avoid:
- too general statements like "Fix generate samples" or "Add test"
- putting multiple things in the commit message summary (i.e. the first line) separated by ";"
- using rST formatting for the changelog (if you need to put something in quotes, just use single quotes "'")

resolwe_bio/management/commands/generate_samples.py
```diff
@@ -158,23 +164,16 @@ def create_data(self, reads_name='seq_reads', annotated=False, rseed=None):
...
         started = timezone.now()
         d = Data.objects.create(
+            slug='gs-reads',
             name=self.set_name(),
             started=started,
             finished=started + datetime.timedelta(minutes=45),
             descriptor_schema=DescriptorSchema.objects.get(slug='reads'),
             descriptor=self.generate_reads_descriptor(),
-            status='OK',
+            status='PR',
```
dblenkus: Use `Data.STATUS_PROCESSING` instead of `PR`

resolwe_bio/management/commands/generate_samples.py
```diff
@@ -227,7 +241,9 @@ def create_data(self, reads_name='seq_reads', annotated=False, rseed=None):
             name='Expression',
             process=Process.objects.get(slug='upload-expression'),
             contributor=self.set_user(),
-            status='OK',
+            started=started,
+            finished=started + datetime.timedelta(minutes=60),
+            status='PR',
```
dblenkus: `Data.STATUS_PROCESSING`

resolwe_bio/management/commands/generate_samples.py
```diff
@@ -249,7 +265,7 @@ def create_data(self, reads_name='seq_reads', annotated=False, rseed=None):
                 'exp_type': 'FPKM',
                 'exp_json': json_object.id
         }
-
+        exp.status='OK'
```
dblenkus: `Data.STATUS_DONE`

resolwe_bio/tests/unit/test_generate_samples.py
```diff
@@ -0,0 +1,23 @@
+# pylint: disable=missing-docstring
+from django.core.management import call_command
+from django.utils.six import StringIO
```
dblenkus: Remove unused imports

jkokosar:
@dblenkus Fixed as suggested!

resolwe_bio/management/commands/generate_samples.py
```diff
@@ -191,6 +190,15 @@ def create_data(self, reads_name='seq_reads', annotated=False, rseed=None):
         new_fastqc_path = os.path.join(data_dir, str(d.id), 'fastqc', reads_name + '_fastqc.html')
         shutil.copy(old_fastqc_path, new_fastqc_path)
 
+        d.output={
+                'fastq': [{'file': os.path.basename(reads)}],
+                'fastqc_url': [{
+                    'url': 'fastqc/{}_fastqc/fastqc_report.html'.format(reads_name),
+                    'refs': ['fastqc/{}_fastqc'.format(reads_name)], 'name': 'View'}],
+                'fastqc_archive': [{'file': '{}_fastqc.zip'.format(reads_name)}]}
+        d.STATUS_DONE
```
dblenkus: You have to assign this value: `d.status = Data.STATUS_DONE`

resolwe_bio/management/commands/generate_samples.py
```diff
@@ -201,20 +209,26 @@ def create_data(self, reads_name='seq_reads', annotated=False, rseed=None):
...
+                'src2': {'file': 'alignment_position_sorted.bam.bai'}})
 
         os.mkdir(os.path.join(data_dir, str(bam.id)))
         shutil.copy(bam_mapping, os.path.join(data_dir, str(bam.id)))
         shutil.copy(bai, os.path.join(data_dir, str(bam.id)))
 
+        bam.output={
+                'bam': {'file': 'alignment_position_sorted.bam'},
+                'bai': {'file': 'alignment_position_sorted.bam.bai'}}
+        bam.STATUS_DONE
```
dblenkus: Same here

resolwe_bio/management/commands/generate_samples.py
```diff
@@ -249,7 +265,7 @@ def create_data(self, reads_name='seq_reads', annotated=False, rseed=None):
                 'exp_type': 'FPKM',
                 'exp_json': json_object.id
         }
-
+        exp.STATUS_DONE
```
dblenkus: and here

dblenkus:
LGTM, @jkokosar please rebase to master

dblenkus:
Rebased and merged

PR #77, opened 2016-06-23
Add require: false for genome index files inputs

jkokosar:


jkokosar:
Closed in favour of https://github.com/genialis/resolwe-bio/pull/81

PR #78, opened 2016-06-27
Fix adding Samples to Collections

dblenkus:


resolwe_bio/views.py
```diff
@@ -38,48 +38,53 @@ class SampleViewSet(CollectionViewSet):
...
-                    raise exceptions.PermissionDenied()
-                else:
-                    raise exceptions.NotFound()
+            self._check_collection_permissions(collection_id, request.user)
 
         for collection_id in request.data['ids']:
             sample.collections.remove(collection_id)
 
+            collection = Collection.objects.get(pk=collection_id)
+            for data in sample.data.all():
```
mstajdohar: This calls an SQL remove statement for each data obj. How about this?
```
collection.data.remove(*sample.data.all())
```

PR #79, opened 2016-06-27
Improve packaging

tjanez:
- Include all necessary files for running the tests in the source distribution.
- Exclude tests from built/installed version of the package.

For more in depth reasoning for this change, refer to Python Distutils-SIG [discussion from last October](https://mail.python.org/pipermail/distutils-sig/2015-October/027003.html).

As a related thing, I removed the obsolete `tests/run_tests.sh` Bash script.


mstajdohar:
LGTM

PR #86, opened 2016-07-05
Add template tag for sample

dblenkus:


resolwe_bio/templatetags/sample.py
```diff
@@ -0,0 +1,42 @@
...
+    """Return `pk` of `Sample` that given `Data` object belongs to."""
+
+    return get_sample_attr(data, 'pk')
+
+
+@register.filter
+def sample_slug(data):
+    """Return `slug` of `Sample` that given `Data` object belongs to."""
+
+    return get_sample_attr(data, 'pk')
```
jkokosar: `return get_sample_attr(data, 'slug')`

resolwe_bio/templatetags/sample.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+@register.filter
+def sample_name(data):
+    """Return `name` of `Sample` that given `Data` object belongs to."""
+
+    return get_sample_attr(data, 'pk')
```
jkokosar: `get_sample_attr(data, 'name')`

jkokosar:
@dblenkus With above corrections works as expected!

PR #90, opened 2016-07-08
Sphinx autoprocess extension

JureZmrzlikar:
No more corrections anticipated.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -5,7 +5,8 @@ import sys
 import os
 import shlex
 
-base_dir = os.path.join(os.path.abspath(os.path.dirname(__file__)), "..")
+docs_dir = os.path.abspath(os.path.dirname(__file__))
+base_dir = os.path.join(docs_dir, "..")
 
 # Get package metadata from 'setup.py' file
 module_setup = imp.load_source('module_setup', os.path.join(base_dir, 'setup.py'))
@@ -25,6 +26,7 @@ extensions = [
     'sphinx.ext.intersphinx',
     'sphinx.ext.coverage',
     'sphinx.ext.viewcode',
+    'resolwe.flow.utils.docs.autoprocess'
 ]
 
 # The suffix(es) of source filenames.
@@ -56,6 +58,12 @@ pygments_style = 'sphinx'
 
 # If true, `todo` and `todoList` produce output, else they produce nothing.
 todo_include_todos = False
+
+# Parent directory of all process definitions:
+autoprocess_process_dir = os.path.normpath(docs_dir + "./../resolwe_bio/processes/")
+
+# Base of the url to process source code:
+autoprocess_source_base_url = 'https://github.com/genialis/resolwe-bio/blob/master/resolwe_bio/processes/'
 
 # -- Options for HTML output ----------------------------------------------
 
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -20,15 +20,4 @@ Contents
    intro
    proc
    contributing
-
-Process Catalog
-===============
-
-TODO: Automatically import a list of processes from the YML files. Organize
-them in categories.
-
-
-Pipelines Examples
-==================
-
-TODO: Include pipeline examples.
+   catalog
--- a/docs/catalog.rst
+++ b/docs/catalog.rst
@@ -1,0 +1,30 @@
+===============
+Process catalog
+===============
+
+Processes are sorted by the following categories:
+
+    * upload
+    * import
+    * analyses
+    * analyses:alignment
+    * analyses:variants
+    * jbrowse
+    * uncategorized
+
+List of processes is quite long, you can navigate or search them in the menu bar on the left.
+
+
+.. autoprocess:: upload
+
+.. autoprocess:: import
+
+.. autoprocess:: analyses
+
+.. autoprocess:: analyses:alignment
+
+.. autoprocess:: analyses:variants
+
+.. autoprocess:: jbrowse
+
+.. autoprocess:: uncategorized


docs/autoprocess.py
```diff
@@ -0,0 +1,412 @@
...
+            * Process slug - if the content of directive is a valid process
+              slug only corresponding process is returned. Actually, a list
+              with only this process is returned.
+            * If none of this options, then raise a ValueError
+
+        :param content: content of the directive
+        :type content: list
+        :param categories: set of all categories
+        :type categories: set
+        :param processes: dict of all github_url: processes pairs
```
dblenkus: `processes` -> `processes_dict`

docs/autoprocess.py
```diff
@@ -0,0 +1,412 @@
...
+            * None - if there is not content, all processes are returned.
+            * Category - if the content of the directive is a string matching
+              an element in categories list, only processes with this
+              category will be returned.
+            * Process slug - if the content of directive is a valid process
+              slug only corresponding process is returned. Actually, a list
+              with only this process is returned.
+            * If none of this options, then raise a ValueError
+
+        :param content: content of the directive
```
dblenkus: For basic types you can use `:param list content: ...` and omit the second line.

http://www.sphinx-doc.org/en/stable/domains.html#info-field-lists

docs/autoprocess.py
```diff
@@ -0,0 +1,412 @@
...
+        :type content: list
+        :param categories: set of all categories
+        :type categories: set
+        :param processes: dict of all github_url: processes pairs
+        :type processes:dict
+
+        :rtype: list
+        """
+        if not content:
+            return processes
```
dblenkus: `processes` is not defined. Probably `processes_dict`?

docs/autoprocess.py
```diff
@@ -0,0 +1,412 @@
...
+        else:
+            content = content[0]
```
dblenkus: What happens with other elements of the list?

docs/autoprocess.py
```diff
@@ -0,0 +1,412 @@
...
+        :param categories: set of all categories
+        :type categories: set
+        :param processes: dict of all github_url: processes pairs
+        :type processes:dict
+
+        :rtype: list
+        """
+        if not content:
+            return processes
+        else:
```
dblenkus: It will be more readable if you omit `else` and remove one level of indentation.

docs/autoprocess.py
```diff
@@ -0,0 +1,412 @@
...
+            querry_for_slug = dict(pair for pair in processes_dict.items() if pair[1]['slug'] == content)
+            if len(querry_for_slug.items()) == 1:
+                return querry_for_slug
+
+            elif content in categories:
+                if content == 'uncategorized':
+                    filtered_dict = dict(
+                        pair for pair in processes_dict.items() if 'category' not in pair[1])
+                    return filtered_dict
+                else:
```
dblenkus: Remove `else`

docs/autoprocess.py
```diff
@@ -0,0 +1,412 @@
...
+                    filtered_dict = dict(
+                        pair for pair in processes_dict.items()
+                        if 'category' in pair[1] and pair[1]['category'] == content)
+                    return filtered_dict
+
+            else:
```
dblenkus: Remove `else`.

docs/autoprocess.py
```diff
@@ -0,0 +1,412 @@
...
+                raise ValueError("Invalid argument given to autoprocess directive.")
+
+    def get_all_processes(self, process_dir, github_base_url):
+        """
+        Return a dictionary of all processes in `process_dir`
+
+        Dictionary key is an URL's to process source code on resolwe-bio
+        GitHub page and value is process as it is loaded from yaml file.
```
dblenkus: Is this linked to GitHub in any way? If not, `source_base_url` would be better.

docs/autoprocess.py
```diff
@@ -0,0 +1,412 @@
...
+        for yaml_file in all_process_files:
+            processes_in_file = read_yaml_file(yaml_file)
+            for process in processes_in_file:
+
+                # This section finds the line in file where the
+                # defintion of the process starts. (there are
+                # multiple process definition in some files).
+                i = 1
+                found = False
+                with open(yaml_file) as file_:
```
dblenkus: You've already read the file in `read_yaml_file` function. Why you have to do it again?

docs/autoprocess.py
```diff
@@ -0,0 +1,412 @@
...
+                    for line in file_:
+                        if 'slug: ' + process['slug'] in line:
```
dblenkus: There is no guarantee that the `slug` attribute will be in the first line of the process.

docs/autoprocess.py
JureZmrzlikar: Currently, they are discarded.

docs/autoprocess.py
```diff
@@ -0,0 +1,412 @@
...
+
+        For processes with undefined category field (yes, there are some!),
+        there is uncategorized category.
+
+        :param process_list: list of all processes
+        :type process_list: list
+
+        :rtype: list
+        """
+        categories = ['uncategorized']
```
dblenkus: categories = set()
    for process in process_list:
        categories.add(process.get('category', 'uncategorized'))
    return list(categories)

docs/autoprocess.py
JureZmrzlikar: With `read_yaml_file` function I directly turn yaml to dictionary. The second opening is juts to determine the starting line of process definition. I don't know how to do it simultaneously.

docs/autoprocess.py
JureZmrzlikar: Indeed, but that was the best idea that i had. Suggestions?

docs/autoprocess.py
```diff
@@ -0,0 +1,412 @@
...
+        title = nodes.subtitle()
+        title_text = nodes.Text(name)
+        title += title_text
+        section += title
+
+        top_field_list = nodes.field_list()
+
+        slug = process['slug']
+        version = process['version']
+        description = process['description']
```
dblenkus: There is a chance that `description` is not given.

    description = process.get('description', '')

docs/autoprocess.py
```diff
@@ -0,0 +1,412 @@
...
+        inputs = process['input'] if 'input' in process else []
```
dblenkus: process.get('input', [])

same for `output`

docs/autoprocess.py
```diff
@@ -0,0 +1,412 @@
...
+              representing the hierarchical structure of the document. More
+              importantly, the whole point of method run is to return a
+              list of nodes. Theese nodes are than parsed by docutils
+              package into HTML/LaTeX...
+
+        """
+
+        # Get all processes:
+        all_processes_dict = self.get_all_processes(PROCESS_DIR, GITHUB_BASE_URL)
+        _, all_processes = zip(*all_processes_dict.items())
```
dblenkus: Maybe `all_processes = zip(*all_processes_dict.items())[1]`?

docs/conf.py
```diff
@@ -5,6 +5,9 @@
 import os
 import shlex
 
+# Add the folder where autoprocess.py extension is located to python path, so sphinx can locate it:
+sys.path.append(os.path.abspath(os.path.dirname(__file__)))
```
dblenkus: Save `os.path.abspath(os.path.dirname(__file__))` to variable and use here and in `base_dir = ...`

docs/autoprocess.py
dblenkus: Ok, I've figured it out :). Move this part to function, so you can just return `i` when process is found or return `1` at the end.

docs/autoprocess.py
```diff
@@ -0,0 +1,412 @@
...
+            paragraph += nodes.Text(field_body)
+        else:
+            # This is the case when field_body is a complex node:
+            # useful when constructing nested field lists
+            paragraph += field_body
+
+        body += paragraph
+        field += name
+        field += body
+        return field
```
dblenkus: name = nodes.field_name() + nodes.Text(field_name)

    if isinstance(field_body, six.string_types):
        field_body = nodes.Text(field_body)
    paragraph = nodes.paragraph() + field_body

    body = nodes.field_body() + paragraph

    return nodes.field() + name + body

Something like this looks more readable to me.

docs/autoprocess.py
```diff
@@ -0,0 +1,401 @@
...
+        :return: line where the process definiton starts
+        :rtype: int
+        """
+        i = 1
+        found = False
+        with open(fname) as file_:
+            for line in file_:
+                if 'slug: ' + slug in line:
+                    found = True
+                    break
```
dblenkus: You don't need `found` anymore. Just `return i` instead of `break`.

docs/autoprocess.py
```diff
@@ -0,0 +1,401 @@
...
+                # Special formating of `value` is needed if name == 'choices'
+                elif name == 'choices':
+                    bullet_list = nodes.bullet_list()
+                    for choice in value:
+                        list_item = nodes.list_item()
+                        paragraph = nodes.paragraph()
+                        label = nodes.Text(choice['label'] + ': ')
+                        val = nodes.literal(choice['value'], choice['value'])
+
+                        paragraph += label
```
dblenkus: Move `paragraph` definition just before it is used. Or maybe even `paragraph = nodes.paragraph() + label + val`?

docs/autoprocess.py
```diff
@@ -0,0 +1,401 @@
...
+                        paragraph += val
+                        list_item += paragraph
```
dblenkus: Same for `list_item`

docs/autoprocess.py
```diff
@@ -0,0 +1,401 @@
...
+        :param url/string source_base_url: url to process definition
+        :param dict process: process data as given from yaml.load function
+        :rtype: nodes.field
+        """
+
+        name = process['name']
+
+        # Make process name as section title:
+        section = nodes.section()
+        section['ids'] = [str(uuid.uuid4())]
```
dblenkus: Is this id visible to the end user? If yes, it should be nice, i.e. use process slug.

dblenkus:
Try not to use `github` in the code if it is not directly link to the GitHub (i.e. GITHUB_BASE_URL is fine). Something like `source_code` or just `source` would be better.

docs/autoprocess.py
```diff
@@ -0,0 +1,401 @@
...
+
+class AutoProcessDirective(Directive):
+    """
+    Directive used to automatically document resolwe-bio process.
+    """
+    has_content = True
+    required_arguments = 0
+    optional_arguments = 0
+    final_argument_whitespace = False
+    option_spec = None
```
dblenkus: Where are these variables used? Are they required by Sphinx?

docs/autoprocess.py
JureZmrzlikar: Unfortunately it cannot be done this way. The "+" operator for docutils.nodes is same as append operator. Therefore
`name += nodes.Text(field_name)`
is same as 
`name.append(nodes.Text(field_name))`

docs/autoprocess.py
JureZmrzlikar: Not visible to end user, but still, better to put process slug.

docs/autoprocess.py
JureZmrzlikar: Yes, every child of `Directive` has to define these attributes.

docs/autoprocess.py
```diff
@@ -0,0 +1,397 @@
...
+
+        elif content in categories:
+            if content == 'uncategorized':
+                filtered_dict = dict(
+                    pair for pair in processes_dict.items() if 'category' not in pair[1])
+                return filtered_dict
+
+            filtered_dict = dict(
+                pair for pair in processes_dict.items()
+                if 'category' in pair[1] and pair[1]['category'] == content)
```
dblenkus: `if pair[1].get('category', 'uncategorized') == content` and you can omit if above.

docs/autoprocess.py
```diff
@@ -0,0 +1,397 @@
...
+        Find the starting line of process definition - where slug is defined
+
+        :param path/str fname: Filename with processes
+        :param string slug: slug of the process
+        :return: line where the process definiton starts
+        :rtype: int
+        """
+        i = 1
+        with open(fname) as file_:
+            for line in file_:
```
dblenkus: `for i, line in enumerate(file_):` and delete `i = 1` and `i += 1`

docs/autoprocess.py
```diff
@@ -0,0 +1,397 @@
...
+
+import six
+import yaml
+
+from docutils import nodes
+from docutils.parsers.rst import Directive
+
+
+PROCESS_DIR = os.path.normpath(os.path.dirname(__file__) + "./../resolwe_bio/processes/")
+SOURCE_BASE_URL = 'https://github.com/genialis/resolwe-bio/blob/master/resolwe_bio/processes/'
```
dblenkus: Is there an option to pass this values as arguments in `rst` file? This directive will be also used in other packages that contains processes, so `resolwe-bio` shouldn't be hardcoded.

docs/autoprocess.py
```diff
@@ -0,0 +1,397 @@
...
+                # This section finds the line in file where the
+                # defintion of the process starts. (there are
+                # multiple process definition in some files).
+                startline = self.get_process_definition_start(yaml_file, process['slug'])
+
+                # Put together URL to starting line of process definition.
+                cutted_yaml_path = yaml_file.split('/resolwe_bio/processes/')
+                source_base_url = source_base_url + cutted_yaml_path[1] + '#L' + str(startline)
+                chosen_process = [proc for proc in processes_in_file if proc['slug'] == process['slug']]
+                assert len(chosen_process) == 1
```
dblenkus: What happens if this is not `True`? Add message with the description of the error.

docs/autoprocess.py
JureZmrzlikar: Good point: It actually should be done this way... we need to define new config_value. Will do it.

docs/autoprocess.py
```diff
@@ -0,0 +1,397 @@
...
+        processes_to_document = self.filter_processes(self.content, categories, all_processes_dict)
+
+        if len(processes_to_document.items()) == 1:
+            # Only one process:
+            source_url, process = processes_to_document.items()[0]
+            top_node = self.make_process_node(source_url, process)
+        elif len(processes_to_document.items()) > 1:
+            # A category of processes, make a process category section:
+            process1 = processes_to_document.items()[0][1]
+            category = process1['category'] if 'category' in process1 else "uncategorized"
```
dblenkus: `process1.get('category', 'uncategorized')`

docs/autoprocess.py
```diff
@@ -0,0 +1,397 @@
...
+        :param set categories: set of all categories
+        :param dict processes_dict: dict of all source_base_url: processes pairs
+        :rtype: list
+        """
+        if not content:
+            return processes_dict
+        content = content[0]
+        # Make a filtered dict of all process that have name == content
+        querry_for_slug = dict(pair for pair in processes_dict.items() if pair[1]['slug'] == content)
+        if len(querry_for_slug.items()) == 1:
```
dblenkus: You can ommit `.items()`

docs/autoprocess.py
dblenkus: Ok, then just define variables just before they are used. This way it is easier to figure it out where they come from.

docs/autoprocess.py
dblenkus: Now, that I've found out, why this is used for, I think that solution is sufficient. Solution would  be to look for first line of yaml element and return it when you find right slug, but this is overkill for this use-case.

docs/autoprocess.py
dblenkus: Raise `NotImplementedError` if there is more than one element.

docs/index.rst
```diff
@@ -20,6 +20,7 @@ Contents
    intro
    proc
    contributing
+   catalog
 
 Process Catalog
 ===============
```
mstajdohar: Remove the Process Catalog and Pipeline Examples sections.

PR #83, opened 2016-06-30
Add manage.py command for creation of test differential expression data

nkran:
@jkokosar, @dblenkus could you please review this. Thanks. I really hope it's not too messy. :)

PR #91, opened 2016-07-08
Update Cuffdiff inputs and results parsing

jkokosar:
Requires: https://github.com/genialis/resolwe/pull/112


resolwe_bio/tools/parse_diffexp.py
```diff
@@ -0,0 +1,19 @@
...
+args = parser.parse_args()
+
+de_data = pd.read_csv(args.de_file, sep='\t')
+de_data.rename(columns={'Unnamed: 0': 'gene_id'}, inplace=True)
+columns = {}
+
+for column in de_data:
+    columns[column] = list(de_data[column])
+
+data = {'de_data': columns}
```
jkokosar: @nkran @lukaw3d This parses whole Diff Exp results table into separate columns and saves to storage. The idea is to build DE results tables and volcano plot in front-end from this results, and than allow for the changes to the plot and tables to be made from the settings (e.g. which column for the Y axis, what are the labels..). Do you propose some other structure for this JSON? For example, should it include default fields like fc, fdr, gene_ids, axis labels?

docs/CHANGELOG.rst
```diff
@@ -33,6 +33,7 @@ Changed
 * Exclude tests from built/installed version of the package
 * Move testing utilities from ``resolwe_bio.tests.processes.utils`` to
   ``resolwe_bio.utils.test``
+* Updated Cuffdiff processor inputs and results table parsing
```
tjanez: Use present tense, i.e. "Update Cuffdiff ..."

dblenkus:
Tested in https://github.com/genialis/resolwe-bio/pull/83

PR #92, opened 2016-07-21
Update settings to work with the newest resolwe

dblenkus:
Requires https://github.com/genialis/resolwe/pull/116


mstajdohar:
Add to change log, fix tests.

dblenkus:
Tests will pass when PR in resolwe will be accepted.

I don't think that change log entry is needed, as this isn't actual change in resolwe-bio, but just in demo project used for testing. So all changes are covered in change log of resolwe. @tjanez what do you think?

mstajdohar:
Included in PR #93.

PR #93, opened 2016-07-22
Update processes to use the updated run command

jkokosar:
Requires: https://github.com/genialis/resolwe/pull/119


mstajdohar:
LGTM. Accept when we merge the [PR #19 on Resolwe](https://github.com/genialis/resolwe/pull/119).

PR #98, opened 2016-08-16
Update process definition documentation

mstajdohar:
Depends on resolwe [PR #125](https://github.com/genialis/resolwe/pull/125).


resolwe_bio/utils/test.py
```diff
@@ -28,9 +28,9 @@ def skipDockerFailure(reason):
 def skipUnlessLargeFiles(*files):
     """Skip the decoreted tests unless the given large files are available.
 
-    :param list *files: variable lenght files list, where each element
-        represents a large file path relative to the ``TEST_FILES_DIR``
-        directory
+    :param list \*files: variable lenght files list, where each element
+                         represents a large file path relative to the
```
dblenkus: I like the before indentation more. With the new style, you lose a lot of space.

dblenkus:
Remove the part about `__about__.py` in `Preparing release` section in `resolwe-bio/docs/contributing.rst`.

Otherwise looks OK.

resolwe_bio/utils/test.py
mstajdohar: Sphinx does not parse correctly the previous version. We did not notice that, because this method wasn't included in the reference (`ref.rst`).

PR #99, opened 2016-08-23
Support linters

mstajdohar:
I have supported pylint, pycodestyle and pydocstyle. Also, the coverage is now reported to codecov.io if checks pass.

Many of the docstrings I added are trivial, and simply follow the function name. The goal was to support this checks for later incremental improvements, not to refactor the documentation.


docs/conf.py
```diff
@@ -10,7 +10,7 @@
 
 
 docs_dir = os.path.abspath(os.path.dirname(__file__))
-base_dir = os.path.join(docs_dir, "..")
+base_dir = os.path.abspath(os.path.join(docs_dir, ".."))
```
tjanez: While you're correcting the `base_dir`, maybe get rid of `docs_dir` altogether since it is only used once below and it needs the `base_dir` anyway :smile:

tjanez: On a related note, in a Python 3.4+ project one could just do:

```python
from pathlib import Path
base_dir = Path(__file__).parent.resolve()
```

When are we dropping Python 2 :stuck_out_tongue_closed_eyes: ?

mstajdohar:
@tjanez, I have address your comment and removed the docs_dir. On the side note, I look forward to switching to Python3.

jkokosar:
@mstajdohar It looks ok, I didn't see any obvious errors.

docs/conf.py
dblenkus: As soon as all dependencies of genialis-base will run on Python 3 :P

I am already developing Resolwe on py3

docs/conf.py
```diff
@@ -73,7 +72,7 @@
 todo_include_todos = False
 
 # Parent directory of all process definitions:
-autoprocess_process_dir = os.path.normpath(docs_dir + "./../resolwe_bio/processes/")
+autoprocess_process_dir = os.path.join(base_dir, 'resolwe_bio/processes/')
```
tjanez: @mstajdohar, wouldn't `os.path.join(base_dir, 'resolwe_bio', 'processes')` be more portable (cross-platform)?

resolwe_bio/utils/test.py
```diff
@@ -15,18 +15,19 @@
...
     """
     if getattr(settings, 'TESTS_SKIP_DOCKER_FAILURES', True):
         return unittest.skip(reason)
     return lambda func: func
 
 
-def skipUnlessLargeFiles(*files):
-    """Skip the decoreted tests unless the given large files are available.
+def skipUnlessLargeFiles(*files):  # pylint: disable=invalid-name
+    r"""Skip the decoreted tests unless a large file is available.
```
tjanez: Why is a raw string needed?

Also, there can be multiple large files, so maybe it's better to write:

- `Skip the decorated tests unless specified large files are available.` or just
- `Skip the decorated tests unless large files are available.` (Please, also fix the mistake in `decoreted` -> `decorated`)

mstajdohar: We have to escape the * in the param docstring below. I had \\ before, but Pylint warned me not to use double backslashes and suggested that I should use the r" syntax.

setup.cfg
```diff
@@ -1,3 +1,16 @@
...
+max-line-length = 119
+# Ignore E127: checked by pylint
+# E127 continuation line over-indented for visual indent
+ignore = E127
+exclude = migrations,setup.py
+
+[pydocstyle]
+match-dir = (?!tests|migrations|\.).*
+match = (?!setup).*.py
+add-select = D203,D404
```
tjanez: Why enable `D203` and disable `D211`?

Pydocstyle changed the default behaviour in [Oct 2015](https://pydocstyle.readthedocs.io/en/latest/release_notes.html#october-9th-2015) due to [Guido van Rossum's change to the official PEP257](https://hg.python.org/peps/rev/9b715d8246db). Guido changed this since the ["previous version didn't reflect the common practice, including the Python standard library"](http://raspberry-python.blogspot.si/2015/01/significant-pep257-change.html?showComment=1420601718075#c4025816682001290873).

resolwe_bio/utils/test.py
mstajdohar: I had two backslashes before (Github formatted my two backslashes as a single backslash).

setup.cfg
mstajdohar: We had more docstrings in D203. Personally I like the symmetry of one extra line before and one after the class doctoring (the one after the docstring is required). Whereas in functions there are no lines before and after the docstring, so there is an inherent symmetry there. I do not mind switching if you would prefer that.

setup.cfg
```diff
@@ -1,3 +1,16 @@
 [bdist_wheel]
 # code is written to work on both Python 2 and Python 3
-universal=1
+universal = 1
+
+[pep8]
+max-line-length = 119
```
tjanez: I would prefer if we would [follow PEP8](https://www.python.org/dev/peps/pep-0008/#maximum-line-length) and limit the maximum line length to 100.

How much work would it require?

mstajdohar: A lot, we should do it separately.

mstajdohar: We would have to fix about 250 lines.

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py{27,34,35},packaging
+envlist = py{27,34,35},docs,linters,packaging
 skip_missing_interpreters = True
```
tjanez: Please, add a general note why we are not using `deps = .[<extra-requirements>]` Tox option like [in Resolwe](https://github.com/genialis/resolwe/blob/8deae4379cb409b54d7114c74ac2ea191373d11f/tox.ini#L5-L6) and remove the notes from individual Tox environments.

setup.cfg
mstajdohar: If we limit the maximum line length to 100, we should do it on all our projects.

mstajdohar: The reason we have set the limit to 119 is because we have followed [Django's coding guidelines](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/#python-style).

tox.ini
```diff
@@ -24,6 +24,30 @@ setenv =
...
+# run all linters to see their output even if one of them fails
+ignore_errors = True
+setenv =
+    PYTHONPATH = {toxinidir}/resolwe_bio/tools
+commands =
+# install testing requirements
+    pip install --process-dependency-links .[test]
+# run pylint
+    pylint resolwe_bio
+# check pep8
```
tjanez: Maybe even better to put a space in the middle, i.e. PEP 8

tox.ini
```diff
@@ -24,6 +24,30 @@ setenv =
...
+    pycodestyle resolwe_bio
+# check pep247
```
tjanez: Same as above, and its actually PEP 257.

tox.ini
```diff
@@ -24,6 +24,30 @@ setenv =
...
+# build documentation
+# NOTE: After https://github.com/sphinx-doc/sphinx/pull/2649 is accepted, use:
+# python setup.py build_sphinx --warning-is-error
+    sphinx-build -E -W docs/ build/sphinx/html
+
+[testenv:linters]
+# run all linters to see their output even if one of them fails
+ignore_errors = True
+setenv =
+    PYTHONPATH = {toxinidir}/resolwe_bio/tools
```
tjanez: It might be worth adding a comment why setting `PYTHONPATH` is necessary. Something like:
`# enable pylint to find modules in the tools package even though it is not installed`

setup.cfg
tjanez: Ok, then leave it at 119 and add a reference to Django's coding guidelines

setup.cfg
tjanez: I'm slightly in favour of D211 since this is the new official PEP 257's stance and it compacts the files a bit.

tjanez:
Great job @mstajdohar, this was a lot of hard work!

Expect for the comments that I wrote inline, this looks good to me.

PR #106, opened 2016-09-09
Rework handling of processes' tools

tjanez:
No longer treating `resolwe_bio/tools` as a Python package will give us a better estimate of the test coverage since (for the time being) we don't have a way to check coverage of Python scripts executed inside Resolwe Flow Executors.

As a bonus, improve file handling when generating expressions. @jkokosar and @nkran, please test if the generator commands work ok.


tjanez:
@dblenkus, I added a [test to ensure tools is not a package](https://github.com/genialis/resolwe-bio/pull/106/commits/d195cd924304322c7b5efec881982fa14ca62173#diff-51ae22269bac0d39f818dab2f92c9bd0R7).

resolwe_bio/management/commands/generate_etc.py
```diff
@@ -38,20 +37,24 @@ def create_etc(gene_ids, path, file_name):
...
-            csvwriter = csv.writer(f, delimiter=str('\t'))
-            times = [0, 4, 8, 12, 16, 20, 24]
-            csvwriter.writerow(['Gene'] + times)
-            with gzopen(gene_ids) as gene_ids:
+            # NOTE: Default line terminator is '\r\n'
+            # NOTE: Python2's csv module doesn't accept a unicode string for delimeter
+            csvwriter = csv.writer(f, delimiter=str('\t'), lineterminator='\n')
+            times = (0, 4, 8, 12, 16, 20, 24)
+            csvwriter.writerow(('Gene', ) + times)
+            with gzip.open(gene_ids, mode='rt') as gene_ids:
```
jkokosar: @tjanez This breaks when reading gzip files that have alternative line endings (the ouput of the django-admin command is not ok). Since `gzip.open`does not support universal new line endings, we introduced the `resolwe_bio.tools.utils.gzopen`function that is fast, handles different line endings and should be used here if there are no other alternatives.

tjanez: Hmm... since this is a generator function, where we control the input files, couldn't we just fix the input file to have the correct line ending?
Also, can you give me an example of a wrong output generated by this function?

BTW, on Python 3, [using `gzip.open` in *text* mode](https://docs.python.org/3/library/gzip.html#gzip.open) actually wraps the `GzipFile` object in a `io.TextIOWrapper` object which has universal newlines enabled by default.

tjanez: I've fixed the line endings of `resolwe_bio/tests/processes/files/dicty_genes.tab.gz` and now it also works on Python 2.

tjanez:
@jkokosar, I've added tools needed to test the contents of the generated samples and differential expressions and tests that use them.

Can you confirm the stored output files are correct?

jkokosar:
@tjanez Large files look OK. Nice work with the added tests!

resolwe_bio/tests/generators/test_management_commands.py
```diff
@@ -0,0 +1,82 @@
...
+        if six.PY2:
+            self.assertJSON(data, data.output['etc'], '', 'large/etc-py2.json.gz')
+            self.assertFile(data, 'etcfile', 'large/etc-py2.json.gz', compression='gzip')
+            # XXX: Shouldn't this file be referenced directly?
+            self._assert_file(data, 'etc.tab', 'large/etc-py2.tab')
+        else:
+            self.assertJSON(data, data.output['etc'], '', 'large/etc-py3.json.gz')
+            self.assertFile(data, 'etcfile', 'large/etc-py3.json.gz', compression='gzip')
+            # XXX: Shouldn't this file be referenced directly?
+            self._assert_file(data, 'etc.tab', 'large/etc-py3.tab')
```
dblenkus: Why you don't use `assertFile`?

dblenkus: Also, use `os.path.join`

tjanez: Done.

tjanez:
@dblenkus, I agree with added complexity so I decided to follow your pattern and reduce the complexity and just move the test files to `resolwe_bio/tests/files`.

Please review and merge when the tests pass.

jkokosar:
@dblenkus Please check if the new tests for generator functions successfully pass also on your local mac machine. I'm getting file hash mismatch errors.

dblenkus:
Generator works as expected on MacOS, the problem is that random generator in some cases generates different number of decimals. Diff looks like:

```
                            u'Zwilch': 27.9979968110566,
                            u'Zwint': 151.44935712786156,
                            u'Zxda': 79.01968310303782,
                            u'Zxdb': 346.5715655591474,
-                           u'Zxdc': 26.805096445873723,
?                                                     ^

+                           u'Zxdc': 26.805096445873726,
?                                                     ^

                            u'Zyg11a': 199.72169273800844,
                            u'Zyg11b': 167.94474805937296,
                            u'Zyx': 16.81807379428961,
                            u'Zzef1': 87.77117337567574,
                            u'Zzz3': 161.04567152211243,
```

dblenkus:
Solution would be to use [assertAlmostEqual](https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertAlmostEqual)-like function that handles dictionaries and lists. One implementation of such function is [here](https://github.com/larsbutler/oq-engine/blob/master/tests/utils/helpers.py#L214), but is under AGPL license, so we cannot copy it.

Same problem is with comparing files, but there is no simple way to fix this, but the these files are used only for development and are the same as JSON (which is already checked), so we can just omit that check.

dblenkus:
I've merged first three commits in separate PR (#108) and fixe some pylint issues, so @tjanez please rebase to master.

mstajdohar:
Please rebase to master and fix the PR description.

dblenkus:
Already accepted.

PR #107, opened 2016-09-13
Add knowledge base app

kostko:
* [x] Tests.
* [x] REST endpoint for `Feature` (admin-only access).
* [x] Query by list of identifiers.


dblenkus:
Do you need any haystack related setting in resolwe-bio's test project?

kostko:
Yes, I will add them when I start adding tests.

mstajdohar:
The coverage check fails because we do not test the `admin.py`. Should we test it? What do we do in this case, accept anyway?

kostko:
How can we test `admin.py`?

mstajdohar:
I don't know. This is why I asked if we should test it... :)

PR #111, opened 2016-09-20
Add Generate GO genesets processor

jkokosar:

PR #112, opened 2016-09-22
Set expression objects as inputs on import of DE data

jkokosar:


resolwe_bio/tools/parse_diffexp.py
```diff
@@ -34,8 +35,12 @@ def main():
     col_order = []
 
     if args.gene_id:
-        columns['gene_id'] = list(de_data[args.gene_id])
-        col_order.append('gene_id')
+        if args.gene_id == 'index':
```
nkran: Why is this particular condition needed here?

nkran:
LGTM. Also tested the import with new processor and test data and it works fine.

resolwe_bio/tools/parse_diffexp.py
jkokosar: This is to handle the way how GeneIDs get parsed from the DE results file. Some DE tools return tab-delimited files where column containing geneIDs is not labeled with the column name. When parsing such files, in some versions of the DE ouput files such column is returned as "Unnamed column 0" and sometimes as index column.

resolwe_bio/processes/import_data/differential_exp.yml
```diff
@@ -114,24 +113,166 @@
...
+        parse_diffexp.py ${NAME}.tab ${GENE_ID} ${LOGFC} ${FDR} ${LOGODDS} ${FWER} ${PVALUE} ${STAT} --output de_data.json
+        re-checkrc
+        re-progress 0.7
+      fi
+      re-save-file raw ${NAME}.tab.gz
+      re-save-file de_file de_file.tab.gz
+      re-save de_json de_data.json
+      re-save source "{{source}}"
+
+- slug: upload-diffexp-inputs
```
mstajdohar: Refactor the `upload-diffexp` process and remove the `upload-diffexp-inputs`. Add `case` and `control` fields as optional fields to the `upload-diffexp` process. This way you will not have to repeat the process logic and bioinformaticians will only have to remember one slug.

resolwe_bio/tools/geneset2storage.py
```diff
@@ -18,7 +18,7 @@ def main():
 
     with open(args.geneset_file) as infile:
         geneset = [gene.strip() for gene in infile]
-        data = {'geneset_json': geneset}
+        data = {'geneset_json': {'geneset_json': geneset}}
```
mstajdohar: Rename the inner `geneset_json` to `genes`.

resolwe_bio/tools/parse_diffexp.py
```diff
@@ -20,6 +20,7 @@ def parse_arguments():
     parser.add_argument('--logodds', help="Log Odds column name", type=str)
     parser.add_argument('--logfc', help="logfc column name", type=str)
     parser.add_argument('--stat', help="Statistics column name", type=str)
+    parser.add_argument('--output', help='Output JSON file')
```
mstajdohar: The parameter should have a default value, because the script crashes not if `--output` not given. I suggest the default is `de_data.json`.

jkokosar:
Close in favour of https://github.com/genialis/resolwe-bio/pull/111

PR #113, opened 2016-09-22
Add upload processors for common tab/txt/image files

jkokosar:


nkran:
Everything seems ok to me 👍

mstajdohar:
Refactor to support three data types: `data:file`, `data:file:table`, `data:file:image`. Use `mv` instead of `re-import` for `data:file`.

mstajdohar:
Tests fail.

resolwe_bio/processes/import_data/file_generic.yml
```diff
@@ -0,0 +1,27 @@
...
+      label: Input file
+      type: basic:file
+  output:
+    - name: file
+      label: File
+      type: basic:file
+  run:
+    runtime: polyglot
+    bash: |
+      mv "{{ src.file_temp|default:src.file }}" .
```
mstajdohar: You probably have to move and rename the file, so:
```
mv "{{ src.file_temp|default:src.file }}" "{{src.file}}"
```
Write a test for this.

resolwe_bio/processes/import_data/image.yml
```diff
@@ -0,0 +1,43 @@
...
+        Image file.
+      validate_regex: '\.(tiff|tif|jpg|jpeg|png|TIFF|TIF|JPG|JPEG|PNG)$'
+  output:
+    - name: file
+      label: Image file
+      type: basic:file
+  run:
+    runtime: polyglot
+    bash: |
+      if [[ '.{{ src.file }}' =~ \.(gif|GIF)$ ]]; then
```
mstajdohar: I would not use re-import here. Just move the file in place as is. re-import only makes sense when you wish to unify the extensions or handle compression. None of this is necessary here.

resolwe_bio/processes/import_data/tabular.yml
```diff
@@ -0,0 +1,50 @@
...
+        Tabular file. Supported extensions: .csv, .csv.gz (preferred) or .csv.*
+      validate_regex: '(\.(csv|tab|tsv|txt|xls|xlsx)(|\.gz|\.bz2|\.tgz|\.tar\.gz|\.tar\.bz2|\.zip|\.rar|\.7z))|(\.bz2)$'
+  output:
+    - name: file
+      label: Tabular file
+      type: basic:file
+  run:
+    runtime: polyglot
+    bash: |
+      if [[ '.{{ src.file }}' =~ \.(xls)$ ]]; then
```
mstajdohar: Re-import not needed unless we wish the file to be zipped. Do we? Just move the original file and then call `parse_tabular_file.py`. Save both, the original file and the parsed file.

resolwe_bio/tools/parse_tabular_file.py
```diff
@@ -0,0 +1,53 @@
...
+def main():
+    """Invoked when run directly as a program."""
+    args = parse_arguments()
+
+    ext = os.path.splitext(args.input_file)[-1].lower()
+    with gzip.open(args.output_file, mode='wt') as outfile:
+        csvwriter = csv.writer(outfile, delimiter=str('\t'), lineterminator='\n')
+
+        try:
+            if ext == '.tab' or ext == 'txt' or ext == 'tsv':
```
mstajdohar: The dot is missing for txt and tsv. Also, more concise and a general solution is:
```
if ext in ('.tab', '.txt', '.tsv'):
```

resolwe_bio/tools/parse_tabular_file.py
```diff
@@ -0,0 +1,53 @@
...
+                with open(args.input_file) as infile:
+                    for line in infile:
+                        outline = line.strip().split('\t')
+                        csvwriter.writerow(outline)
+            elif ext == '.csv':
+                with open(args.input_file) as infile:
+                    for line in infile:
+                        outline = line.strip().split(',')
+                        csvwriter.writerow(outline)
+            elif ext == '.xls' or ext == '.xlsx':
```
mstajdohar:
```
elif ext in ('.xls', '.xlsx'):
```

resolwe_bio/processes/import_data/tabular.yml
```diff
@@ -0,0 +1,55 @@
...
+      label: Tabular file
+      type: basic:file
+    - name: src_file
+      label: Source file
+      type: basic:file
+      required: false
+  run:
+    runtime: polyglot
+    bash: |
+      if [[ '.{{ src.file }}' =~ \.(xls)$ ]]; then
```
mstajdohar: Cant't you just do this instead of all the code here in `bash`?
```
mv "{{ src.file_temp|default:src.file }}" "{{ src.file }}"
FILENAME=$(basename "{{ src.file }}")  # get file name
NAME="${FILENAME%.*}"  # remove extension
parse_tabular_file.py "{{ src.file }}" "${NAME}.tab.gz"
```

PR #117, opened 2016-09-29
Workaround for Haystack limitation

kostko:


kostko:
One more thing I noticed while doing this workaround - if there is an exception during the indexing process, Haystack will hang the whole application, which will require a `SIGKILL` to stop. (Probably related to some buggy interaction with whoosh.)

PR #118, opened 2016-09-30
Set presample to False on Sample endpoint

dblenkus:


mstajdohar:
The `presample` field is not returned for Sample. Should we support that?

dblenkus:
I think not, because:
- it's `True` for all objects and you cannot change it
- it doesn't mean anything - once object becomes sample, you don't care about presamples

PR #119, opened 2016-10-05
Improve generator tests and work around platform differences

tjanez:
Introduce `GeneratorBioProcessTestCase` class with two methods:
- `assertEqualIgnoreNumbers` and
- `assertDictEqualIgnoreNumbers` which uses the former.
Convert `GenerateSamplesTest` and `GenerateEtcTest` to use `assertDictEqualIgnoreNumbers` and replace their test files.
Round expression and expression time course values to 2 digits after the decimal point which substantially reduces the size of the generated JSON and tab files.
Drop `exp_json` inner dict when generating expressions.

### Comment

In the future it would be great if we implemented proper generator tests which would run both, the real processor and the corresponding generator and compare the *structure* of the generated data objects' fields and files.


resolwe_bio/tests/generators/test_management_commands.py
```diff
@@ -10,7 +11,49 @@
 from resolwe_bio.models import Sample
 
 
-class GenerateSamplesTest(BioProcessTestCase):
+class GeneratorBioProcessTestCase(BioProcessTestCase):
+
+    def assertEqualIgnoreNumbers(self, first, second, msg=None):  # pylint: disable=invalid-name
```
mstajdohar: :)

resolwe_bio/tests/generators/test_management_commands.py
```diff
@@ -20,37 +63,46 @@ def test_generate_samples(self):
...
-            if six.PY2:
-                self.assertJSON(expr, expr.output['exp_json'], '', 'expressions-py2.json.gz')
-                self.assertFile(expr, 'exp', 'expressions-py2.tab.gz', compression='gzip')
-            else:
-                self.assertJSON(expr, expr.output['exp_json'], '', 'expressions-py3.json.gz')
-                self.assertFile(expr, 'exp', 'expressions-py3.tab.gz', compression='gzip')
+            # NOTE: The random.gammavariate() that is used for generating expression data produces
+            # slightly differently rounded numbers on different platforms (e.g. Linux and MacOS) so
+            # the check ignores the generated numbers
+            with mock.patch.object(self, 'assertDictEqual', self.assertDictEqualIgnoreNumbers):
```
mstajdohar: What is the point of the check then?

tjanez: It checks everything else in the generated JSON, the dict structure, lists, strings, ...

The problem is that the check is "static", i.e. it compares the structure to the saved JSON file. What we want in the future is a "dynamic" check that compares the generator's output to the output of a real processor it is simulating.

PR #120, opened 2016-10-06
Manually specify Django and Django REST Framework dependencies

tjanez:
This works around problems when running `python setup.py install --force` and incorrect versions of Django and Django REST Framework dependencies are installed.

This should fix the Read the Docs builds.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,6 +47,8 @@ setup(
         "git+https://github.com/genialis/resolwe.git@master#egg=resolwe-1.3.99",
     ),
     install_requires=(
+        'Django~=1.9.11',
+        'djangorestframework>=3.4.0',
         'resolwe>=1.3.1',
         'django-haystack==2.5.0',
         'drf-haystack==1.5.6',
@@ -54,7 +56,7 @@ setup(
     ),
     extras_require = {
         'docs':  [
-            'sphinx>=1.3.2',
+            'Sphinx>=1.3.2',
             'sphinx_rtd_theme',
         ],
         'package': [


setup.py
```diff
@@ -47,6 +47,14 @@
...
     ),
     install_requires=(
+        # XXX: Need to manually specify these, otherwise
+        # 'python setup.py install --force' (used by Read the Docs) fails with:
+        # error: Django 1.10.1 is installed but Django<1.10 is required by {'django-haystack'}
+        'django>=1.9,<1.10a1',
+        # XXX: Need to manually specify these, otherwise
+        # 'python setup.py install --force' (used by Read the Docs) fails with:
+        # djangorestframework 3.4.7 is installed but djangorestframework==3.3.2 is required by {'resolwe'}
+        'djangorestframework==3.3.2',
```
dblenkus: Can we update DRF to the latest version in resolwe?

mstajdohar: As soon as possible so that we have time to test it intensively by the next deploy at Rochester.

dblenkus: There are no major changes. If test passes it should work. 

Only notable change is that time is returned microsecond precision (6 digits), rather than millisecond precision (3 digits), so we there may be few tests to update.

tjanez: Ok, great. @dblenkus, please prepare a pull request for Resolwe and link it here, so I can remove the DRF dependency pinning after it is merged.

dblenkus: PR at resolwe is ready: https://github.com/genialis/resolwe/pull/133

tjanez:
@dblenkus, I've updated the PR. Please, take another look.

PR #124, opened 2016-10-21
Fix kb serializer

dblenkus:
After upgrading `django-rest-framework`, tests are failing with:

    AssertionError: (u"Creating a ModelSerializer without either the 'fields' attribute or the 'exclude' attribute has been deprecated since 3.3.0, and is now disallowed. Add an explicit fields = '__all__' to the FeatureSearchSerializer serializer.",)

### DIFF
--- a/resolwe_bio/kb/serializers.py
+++ b/resolwe_bio/kb/serializers.py
@@ -19,3 +19,4 @@ class FeatureSerializer(serializers.ModelSerializer):
         """Serializer configuration."""
 
         model = Feature
+        fields = '__all__'


resolwe_bio/kb/serializers.py
```diff
@@ -19,3 +19,4 @@ class Meta:
         """Serializer configuration."""
 
         model = Feature
+        fields = '__all__'
```
kostko: Why not? Search endpoints are all read-only and the REST endpoints are admin-only.

dblenkus: This doesn't change any behaviour. It just needs to be specified explicitly in new version of DRF.

kostko:
LGTM

PR #125, opened 2016-10-24
Refactor for updated Resolwe

kostko:
This requires genialis/resolwe#138.

### DIFF
--- a/resolwe_bio/expression_filters/__init__.py
+++ b/resolwe_bio/expression_filters/__init__.py
@@ -1,0 +1,1 @@
+"""Additional expression filters for Resolwe expression engines."""
--- a/resolwe_bio/expression_filters/sample.py
+++ b/resolwe_bio/expression_filters/sample.py
@@ -7,12 +7,7 @@ Sample Template Tags
 """
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-from django import template
-
 from resolwe_bio.models import Sample
-
-
-register = template.Library()  # pylint: disable=invalid-name
 
 
 def get_sample_attr(data, attr):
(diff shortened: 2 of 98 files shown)


dblenkus:
@mstajdohar, @tjanez, @jkokosar: Is there a reason to use `|default(src.file)` in first argument of `re-import` function? I.e.: 

    re-import "{{ src2.file_temp|default(src2.file) }}" "{{ src2.file }}" "bai" "bai" 0.3 extract

If not, we can remove it now when all processes are being refactored

dblenkus:
@kostko version of all changed processes must be increased

kostko:
Which version atom should be changed (eg. major/minor version) or is this arbitrary as long as the version is increased?

mstajdohar:
Arbitrary as long as the version is increased.

mstajdohar:
We can remove the `|default(src.file)` when moving uploaded files. All uploaded files have the `.file_temp` field. Should we make the `. file_temp` field required on inputs?

kostko:
For future reference, to bump process versions in a single line of bash:
```bash
find -name "*.yml" -exec sed -i -r 's/(\s+)version: ([0-9]+)\.([0-9]+)\.([0-9]+)/echo "\1version: \2.\3.$((\4 + 1))"/ge' {} \;
```

mstajdohar:
> For future reference, to bump process versions in a single line of bash...

That is dangerous so if you use it in the future, do it with thought and caution.

kostko:
> This is dangerous so if you use it in the future, do it with thought and caution.

Yes, you just have to take care to only match processors that were actually updated (and not also descriptor schemas, etc.) and review the changes before committing them.

kostko:
Rebased on latest master, please review.

PR #126, opened 2016-10-24
Add filtering to deseq2 process

JenkoB:


resolwe_bio/processes/differential_expression/deseq.yml
```diff
@@ -23,6 +23,12 @@
       type: list:data:expression
       description: >
         Control samples (replicates)
+    - name: filter
+      label: Filtering of raw counts
+      type: basic:integer
+      default: 10
+      description: >
+        The cut of value for filtering of raw counts
```
mstajdohar: cutoff -> https://en.wikipedia.org/wiki/Cutoff

JenkoB: I changed to: The threshold value for raw counts filtering.

resolwe_bio/processes/differential_expression/deseq.yml
```diff
@@ -23,6 +23,12 @@
       type: list:data:expression
       description: >
         Control samples (replicates)
+    - name: filter
+      label: Filtering of raw counts
```
mstajdohar: Write what is the parameter, not what is does. Think of this as along parameter name. For example, "Raw counts threshold" or "Raw counts filtering threshold".

resolwe_bio/processes/differential_expression/deseq.yml
```diff
@@ -23,6 +23,12 @@
...
+      type: basic:integer
+      default: 10
+      description: >
+        The threshold value for raw counts filtering.
```
mstajdohar: Write what it does. For example, "Filter genes in the expression matrix input. Remove genes where the number of counts in all samples is below the threshold.".

mstajdohar:
Also, rebase on to master.

dblenkus:
Fixed, rebased and (again) bumped process version.

PR #134, opened 2016-11-04
Fix index name in Subread aligner

jkokosar:


kostko:
Is there a test that catches such errors? If not, perhaps it should be added?

jkokosar:
We test the processors, but in this case the file name we use in the test matched perfectly the "hard-coded" file name for the genome index in the processor code, so the test didn't catch the error.

JenkoB:
Thanks Janez. I see the error, I will be more careful on this.

kostko:
I still think that we should add a test that will catch such errors.

dblenkus:
Do you have any idea how to test this?

It is not good idea to add additional tests, as we would double testing time. Maybe we can generate random names for all files and update db entries after process is done? This would definitely catch all such errors, but is a bit hackish and would (probably) make debugging harder.

mstajdohar:
Test with two genomes with two different names. This is also an opportunity to test strange characters in the genome name. The code should work if the name includes several dots, space...

jkokosar:
Added test as suggested. However, this still doesn't test for the same type of semantic error in other aligner processors.

resolwe_bio/tests/processes/test_alignment.py
```diff
@@ -200,3 +200,10 @@ def test_subread(self):
         inputs = {'genome': genome.id, 'reads': reads_paired.id}
         aligned_reads = self.run_process('alignment-subread', inputs)
         self.assertFile(aligned_reads, 'stats', 'subread_paired_reads_report.txt')
+
+        inputs = {'src': 'my.strange.genome name$.fasta.gz'}
```
dblenkus: You can make the name even stranger, like `Ḡ℮η◎ღℯ` :)

PR #135, opened 2016-11-06
Update pycodestyle config

dblenkus:
Tests are failing with:

    ************* Module resolwe_bio.kb.models
    C: 63, 4: Old-style class defined. (old-style-class)
    ************* Module resolwe_bio.kb.serializers
    C: 18, 4: Old-style class defined. (old-style-class)

@tjanez, @kostko do you have any idea why? Exact same definition for Meta class is passing on Resolwe. Is it possible that linters are run on python 2.7 here and on 3.5 on Resolwe (classes with no parent specified are old-style objects in py2 and new-style objects in py3)?

### DIFF
--- a/resolwe_bio/tools/summarize_expression_ncRNA.py
+++ b/resolwe_bio/tools/summarize_expression_ncRNA.py
@@ -31,8 +31,8 @@ args = parser.parse_args()
 
 gene_id_re = re.compile(r'ID=([\w\-\.]*)')
 gene_name_re = re.compile(r'oId=([\w\-\.]*)')
-class_code_re = re.compile(r'class_code=([\w\-\.]*)')
 parent_id_re = re.compile(r'Parent=([\w\-\.]*)')
+class_code_re = re.compile(r'class_code=([\w\-\.]*)')
 
 
 def _search(regex, string):
--- a/resolwe_bio/expression_filters/sample.py
+++ b/resolwe_bio/expression_filters/sample.py
@@ -38,6 +38,7 @@ def sample_name(data):
     """Return `name` of `Sample` that given `Data` object belongs to."""
     return get_sample_attr(data, 'name')
 
+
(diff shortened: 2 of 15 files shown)


resolwe_bio/tools/summarize_expression_ncRNA.py
```diff
@@ -31,8 +31,8 @@
 
 gene_id_re = re.compile(r'ID=([\w\-\.]*)')
 gene_name_re = re.compile(r'oId=([\w\-\.]*)')
-class_code_re = re.compile(r'class_code=([\w\-\.]*)')
 parent_id_re = re.compile(r'Parent=([\w\-\.]*)')
+class_code_re = re.compile(r'class_code=([\w\-\.]*)')
```
dblenkus: This one is strange - it passes with new order of regex and fail with the old one:

    resolwe_bio/tools/summarize_expression_ncRNA.py:35:1: E305 expected 2 blank lines after class or function definition, found 0

mstajdohar: Looks like a bug in the validator. The variable starts with `class_`...

tjanez: @mstajdohar, I think you are right since [this check](https://github.com/PyCQA/pycodestyle/blob/d12e2ea39745bd3bb9e819d7342e4c1d73dbb008/pycodestyle.py#L293-L294) only looks if the previous (logical) line starts with `def` or `class`.

@dblenkus, will you report an issue [@PyCGA/pycodestyle](https://github.com/PyCQA/pycodestyle/issues)?

dblenkus: There is already pull request for this problem: https://github.com/PyCQA/pycodestyle/pull/593

PR #140, opened 2016-11-11
Unify handling of Samples

dblenkus:
Requires: https://github.com/genialis/resolwe/pull/150

### DIFF
--- a/resolwe_bio/signals.py
+++ b/resolwe_bio/signals.py
@@ -7,7 +7,7 @@ Signal Handlers
 """
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-from django.db.models.signals import post_save
+from django.db.models.signals import pre_delete, post_save
 from django.dispatch import receiver
 
 from guardian import shortcuts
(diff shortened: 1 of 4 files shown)


dblenkus:
I've fixed linter errors. Tests will pass when all requirements will be accepted.

resolwe_bio/signals.py
```diff
@@ -56,4 +56,21 @@ def add_post_save_handler(sender, instance, **kwargs):
...
+            #      object is added to collections.
+            # for collection in Collection.objects.filter(data=instance.pk):
+            #     sample.collections.add(collection)
+
         sample.data.add(instance)
+
+
+@receiver(pre_delete, sender=Data)
+def add_pre_delete_handler(sender, instance, **kwargs):
+    """Delete Sample when last Data object is deleted."""
```
mstajdohar: Did you think about deleting a Sample when the last Data object is removed?

dblenkus: Good point. I've added this.

mstajdohar: Well, I am still thinking if we want this or not. But in any case, we have to either delete the Sample in both cases or not delete it at all.

PR #141, opened 2016-11-14
Fix upload gtf annotation

JenkoB:


mstajdohar:
@jkokosar please check this won't cause unexpected behavior in downstream analysis.

jkokosar:
@mstajdohar The block of code that was removed from this GTF upload processor produced the GFF file that was not saved in the outputs and was thus purged anyways. If the goal is to save GFF file, users should use the GFF upload processor.

PR #142, opened 2016-11-14
Split SRA process into two processes

dblenkus:


mstajdohar:
> fastq-dump used in the processor relies on the external services (NCBI), with the smallest SRA files being >400 MB. We cannot write such a test without mocking NCBI services.

@dblenkus, you thoughts on this?

dblenkus:
I agree that at the moment we can't write tests for these processes. If we want to cover such cases, we should discus how to do it (i.e. we can mock http response, inject our `fatsq-dump` command into docker which won't use internet,...). But all solutions are far from trivial.

mstajdohar:
Rebase on to master.

dblenkus:
Done.

PR #143, opened 2016-11-15
Fix edger processor that user can see DE for all genes

JenkoB:


docs/CHANGELOG.rst
```diff
@@ -53,6 +53,7 @@ Fixed
 * Set ``presample`` to ``False`` for Samples created on Sample endpoint
 * Fix FastQC report paths in processors
 * Fix ``htseq_count`` and ``featureCounts`` for large files
+* Fix ``edge_R`` that user can see DE for all genes
```
jkokosar: Not required. This is still "unreleased" change covered by "Add process for differential expression edgeR"

mstajdohar:
Fixed the change log entry and rebased on to master.

PR #144, opened 2016-11-15
Alter KB model fields

jkokosar:


resolwe_bio/kb/migrations/0002_alter_fields.py
```diff
@@ -0,0 +1,31 @@
...
+            field=django.contrib.postgres.fields.ArrayField(base_field=models.CharField(max_length=256), blank=True, default=[], size=None),
+        ),
+        migrations.AlterField(
+            model_name='feature',
+            name='name',
+            field=models.CharField(max_length=1024),
+        ),
+        migrations.AlterField(
+            model_name='feature',
+            name='sub_type',
```
dblenkus: This field hasn't changed. Why is it in the migrations? @kostko is this the change that was missing from the migrations?

resolwe_bio/kb/migrations/0002_alter_fields.py
```diff
@@ -0,0 +1,31 @@
+# -*- coding: utf-8 -*-
```
dblenkus: Change filename to something more meaningful that "alter fields".

mstajdohar:
Issues were addressed. @dblenkus, please do another review pass.

PR #145, opened 2016-11-17
Add HEAT-Seq pipeline related tools

jkokosar:
* Add HSQutils tools (trimm, dedup)
* Make samtools VC processor compatible with the HEAT-Seq pipeline
* Fix trimmomatic parameters


resolwe_bio/processes/reads_processing/hsqutils.yml
```diff
@@ -0,0 +1,86 @@
...
+      label: Download FastQC archive (Downstream)
+      type: list:basic:file
+  run:
+    runtime: polyglot
+    language: bash
+    program: |
+      NAME1=`basename '{{ reads.fastq.0.file }}' .fastq.gz`
+      NAME2=`basename '{{ reads.fastq2.0.file }}' .fastq.gz`
+
+      {% for r in reads.fastq %}
```
dblenkus: Use this instead of for loop:

    FW_READS="{{ reads.fastq|join(' ') }}"

Fix this in whole PR.

resolwe_bio/processes/variant_calling/samtools.yml
```diff
@@ -30,38 +45,42 @@
...
+      re-progress 0.1
 
-      echo "creating VCF file"
-      samtools mpileup -u -f "${GENOME_NAME}.fasta.gz" "${BAM_FILE}_nodupli.bam" > "${BAM_FILE}_raw.bcf";
-      echo "{\"proc.progress\":0.5,\"proc.rc\":$?}"
+      {% if options.rmdup %}
+        echo "removing duplicates"
+        samtools rmdup "{{ mapping.bam.file }}" "${BAM_FILE}_dedup.bam"
+        re-checkrc "The process of removing duplicates failed."
+        samtools mpileup -Bugf "${GENOME_NAME}.fasta" "${BAM_FILE}_dedup.bam" {% if bed %} -l "{{ bed.BED.file }}" {% endif %} > "${BAM_FILE}_raw.bcf"
```
dblenkus: Following three lines are almost the same as those in else. Put them out of if and in if just define which bam file to use.

PR #147, opened 2016-11-23
Fix gene_id field type for DE storage objects

jkokosar:


resolwe_bio/tests/processes/test_upload.py
```diff
@@ -135,6 +135,21 @@ def test_upload_de(self):
         self.assertFile(diff_exp, 'raw', 'deseq2_output.tab.gz')
         self.assertJSON(diff_exp, diff_exp.output['de_json'], '', 'deseq2_volcano_plot.json.gz')
 
+    def test_upload_de_check_field_type(self):
```
jkokosar: It fails as intended.

PR #149, opened 2016-11-25
Fix samtools variant caller

jkokosar:
* Fix samtools VC process so it correctly filters raw variants
* Add additional option to BWA aligner to improve Picard compatibility


resolwe_bio/processes/alignment/bwa.yml
```diff
@@ -141,9 +136,9 @@
 
       echo "ALIGNING with BWA-MEM:"
       {% if reads|type|subtype('data:reads:fastq:single:') %}
-        bwa mem {% if reporting.report_all %}-a{% endif %} -r {{re_seeding}} -k {{ seed_l }} -w {{ band_w }} -A {{scoring.match}} -B {{scoring.missmatch}} -O {{scoring.gap_o}} -E {{scoring.gap_e}} -L {{scoring.clipping}} -T {{ reporting.report_tr }} ${INDEX} fw_reads.fastq 1>"${NAME}_align_unsorted.sam"
+        bwa mem {% if reporting.report_all %}-a{% endif %} -r {{re_seeding}} -k {{ seed_l }} -w {{ band_w }} -A {{scoring.match}} -B {{scoring.missmatch}} -O {{scoring.gap_o}} -E {{scoring.gap_e}} -L {{scoring.clipping}} -T {{ reporting.report_tr }} {% if m %} -M {% endif %} ${INDEX} fw_reads.fastq 1>"${NAME}_align_unsorted.sam"
```
mstajdohar: Check if you can specify the output file name.

mstajdohar:
Would be really, really nice if you would write tests that would detect this bug.

mstajdohar:
Some tests have failed.

resolwe_bio/processes/alignment/bwa.yml
jkokosar: We use redirections `>` to name the output "sam" files, that are latter converted to bam. Test was expanded to check if the previously missing file exists.

PR #151, opened 2016-11-30
Add mapping endpoint

kostko:
This is to support mapping queries requested by @jkokosar in IN-80.


resolwe_bio/kb/views.py
```diff
@@ -108,3 +110,36 @@ def create(self, request):
...
             return super(FeatureViewSet, self).create(request)  # pylint: disable=no-member
+
+
+class MappingViewSet(mixins.ListModelMixin,
+                     mixins.RetrieveModelMixin,
+                     viewsets.GenericViewSet):
+    """API view for :class:`Mapping` objects."""
+
+    serializer_class = MappingSerializer
+    filter_backends = [filters.DjangoFilterBackend]
```
dblenkus: Does this have to be set?

resolwe_bio/kb/views.py
```diff
@@ -108,3 +110,36 @@ def create(self, request):
...
+    def as_view(cls, actions=None, **initkwargs):
+        """Support POST for searching against a list of feature ids."""
+        if actions.get('get', None) == 'list':
+            actions['post'] = 'list_with_post'
+
+        return super(cls, MappingViewSet).as_view(actions, **initkwargs)
+
+    def list_with_post(self, request):
+        """Support search via a POST request in addition to GET."""
+        queryset = self.queryset
```
dblenkus: Use `self.get_queryset()`

resolwe_bio/kb/views.py
kostko: Yes, otherwise the permission filter will crash (knowledge base is permissionless). Or is there another way of disabling per-object permissions for a specific endpoint?

kostko:
@mstajdohar you merged too soon as there were still minor comments to address.

mstajdohar:
Sorry, I followed the approve message. Please provide a fix.

resolwe_bio/kb/views.py
dblenkus: Ok, there is probably no better way.

PR #159, opened 2016-12-07
Extend input options for gene set upload

jkokosar:
Needed for https://github.com/genialis/genialis-apps/issues/320


mstajdohar:
I thought the file would be parsed in TypeScript. I will accept this solution if you give good arguments against it.

UnspokenMallard:
This was meant for cases when you want to save a gene set based only on the array of gene ids.

mstajdohar:
Sorry. I see now that the file upload was supported first and that you've added the support for a list of genes later. I would suggest you write another (yet similar) process for that and call it `create-geneset`. So we would have two processes to create gene sets. `upload-geneset` would accept a file and `create-geneset` would accept a list of genes.

resolwe_bio/processes/import_data/geneset.yml
```diff
@@ -50,8 +55,22 @@
...
     program: |
-      re-import "{{ src.file_temp }}" "{{ src.file }}" "tab|txt" "tab" 0.9
-      re-save-file geneset ${NAME}.tab.gz
+      {% if src and features %}
+        re-error "Provide eiter a gene set file or enter a list of features."
+      {% elif not src and not features %}
+        re-error "A gene set file or a list of features must be provided."
+      {% endif %}
+
+      {% if src %}
```
lukaw3d: If this was javascript, I would prefer a bit less duplication:
```sh
# Get the ${NAME}.tab and ${NAME}.tab.gz files
{% if src %}
  re-import "{{ src.file_temp }}" "{{ src.file }}" "tab|txt" "tab" 0.9
{% elif features %}
  NAME="geneset"
  echo "{{ features|join('\n') }}" >> ${NAME}.tab
  gzip ${NAME}.tab
{% endif %}

# Use the files
re-save-file geneset ${NAME}.tab.gz
re-save source {{source}}
geneset2storage.py "${NAME}.tab" geneset.json
re-save geneset_json geneset.json
```

PR #160, opened 2016-12-08
Add create-geneset-venn processor

jkokosar:


mstajdohar:
@lukaw3d, is the file on input OK?

lukaw3d:
I'd prefer if it was a string input, but I haven't checked if we support `basic:json:` as input. That would be a lot easier to send

mstajdohar:
We do not support `basic:json:` and adding it would not be trivial.

lukaw3d:
How about an un-validated basic:string input, that gets outputted to validated basic:json? Much like current input file

mstajdohar:
A lot of room for errors there. I can imagine problems with serialization of weird characters...

lukaw3d:
Well, isn't it the same within the contents of a json file?

mstajdohar:
Not really. Strings are serialized first by DRF and then they are "pasted" into bash.

resolwe_bio/processes/import_data/geneset.yml
```diff
@@ -108,3 +108,67 @@
...
+        - label: NCBI
+          value: NCBI
+        - label: UCSC
+          value: UCSC
+    - name: venn
+      label: Venn diagram
+      type: basic:file
+      description: >
+        JSON file. Supported extensions: .json.gz
+      validate_regex: '(\.json\.gz)$'
```
lukaw3d: This should allow non-zipped json files too

PR #166, opened 2016-12-12
Fix feature autocomplete result order

kostko:
This PR contains a workaround for buggy Haystack mapping for edgengram-based fields.
 
Requires index rebuild.


mstajdohar:
This works much better as before. Would you also add tests?

dblenkus:
I think that we can accept this without tests, as we will switch to ElasticSearch in near feature.

@kostko, @mstajdohar do jou agree?

PR #167, opened 2016-12-14
Add source query parameter to kb autocomplete

hadalin:


resolwe_bio/kb/views.py
```diff
@@ -89,11 +89,17 @@ class FeatureAutocompleteViewSet(HaystackViewSet):
 
     def filter_queryset(self, queryset):
         """Construct a correct filter query."""
-        query = self.request.query_params.get('query', '')
```
hadalin: @kostko Any reason why empty string instead of `None`?

kostko: No, probably a leftover from where the query was different.

mstajdohar:
Missing a test for the empty query.

docs/contributing.rst
```diff
@@ -86,6 +86,10 @@ Change directory to the Django test project::
 
     cd tests
 
+Run docker::
```
dblenkus: This is not part of this commit

kostko:
BTW, there is something wrong with commit authorship.

PR #168, opened 2016-12-15
Refactor knowledge base mapping endpoints

kostko:
Make the endpoints more similar to how features are handled. Also adds support for adding new mappings using the API.

Requires index rebuild.

### DIFF
--- a/resolwe_bio/kb/views.py
+++ b/resolwe_bio/kb/views.py
@@ -8,28 +8,15 @@ Views
 from django.utils.decorators import classonlymethod
 
 from rest_framework import viewsets, mixins, permissions, filters
-from rest_framework.response import Response
 
 from haystack.query import SQ
 
-from drf_haystack.serializers import HaystackSerializerMixin
 from drf_haystack.viewsets import HaystackViewSet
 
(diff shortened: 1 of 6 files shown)


mstajdohar:
Fix lint errors.

resolwe_bio/kb/views.py
```diff
@@ -123,34 +99,77 @@ def create(self, request):
...
+        """A custom create, which updates existing mappings instead of failing."""
+        try:
+            mapping = Mapping.objects.get(
+                source_db=request.data['source_db'],
+                source_id=request.data['source_id'],
+                target_db=request.data['target_db'],
+                target_id=request.data['target_id']
+            )
+            self.kwargs[self.lookup_field] = mapping.pk
+            return super(MappingViewSet, self).update(request)  # pylint: disable=no-member
```
dblenkus: Is this ok to update the object in create method?

kostko: Yes, we also have this for features to avoid a needless round-trip. Otherwise the API would first need to check if the object  already exists on the server and then create or update (and this would also introduce a potential race condition).

PR #170, opened 2016-12-22
Document when '--process-dependency-links' has to be used

tjanez:
When installing from source or preparing a contributor development environment, use `--process-dependency-links` pip argument to be able to install the latest `master` version of Resolwe from git.
Specify an exact version of Resolwe (during development phase of Resolwe Bioinformatics) and change it to `master` so that it will be always installed from git.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -91,7 +91,7 @@ From source
 
 .. code::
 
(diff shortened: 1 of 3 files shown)


mstajdohar:
Add to change log.

tjanez:
@mstajdohar, since this is only a change for developers/contributors, I didn't put it in the CHANGELOG.

PR #175, opened 2017-01-06
Use new ES API for feature search

kostko:
Requires genialis/resolwe#185.


mstajdohar:
Still some linter errors to fix.

PR #179, opened 2017-01-08
Fix save to storage for large GOEA results

jkokosar:


resolwe_bio/tools/goea.py
```diff
@@ -66,7 +66,9 @@ def main():
                         stderr=DEVNULL
                         )
         out, err = process.communicate()
-        print('{{"terms":{}}}'.format(out.decode("UTF-8")))
+
+        with open('terms.txt', 'w') as f:
```
mstajdohar: It would be nicer to name the file `terms.json`.

mstajdohar:
Also, rebase onto master.

PR #180, opened 2017-01-09
Use bulk push in insert_features command

dblenkus:


mstajdohar:
Fix tests for Py3 and it's ready to merge.

PR #183, opened 2017-01-11
Add cuffquant workflow

jkokosar:


resolwe_bio/tests/workflows/test_rna_seq.py
```diff
@@ -0,0 +1,19 @@
...
+        inputs = {'src': 'annotation.gff.gz', 'source': 'DICTYBASE'}
+        annotation = self.run_process('upload-gff3', inputs)
+
+        self.run_process(
+            'workflow-rnaseq-cuffquant', {
+                'reads': reads.id,
+                'genome': genome.id,
+                'annotation': annotation.id,
+            }
+        )
```
mstajdohar: Should you check if the status is OK?

dblenkus: This check is already performed in `run_process`.

PR #184, opened 2017-01-12
Fix data names in macs 14

JenkoB:


JenkoB:
@jkokosar please review.

PR #185, opened 2017-01-12
Fix sample hierarhical clustering

JenkoB:


resolwe_bio/processes/clustering/hierarchical_clustering.yml
```diff
@@ -118,6 +118,7 @@
     runtime: polyglot
     language: bash
     program: |
-      samplehcluster.py {% for e in exps %}"{{e.exp.file}}" {% endfor %} --sampleids {% for e in exps %}{{e|sample_id}} {% endfor %}{% if genes %}--genes {{genes|join(" ")}}{% endif %}--linkage {{linkage}} --dstfunc {{distance}} {% if filter %}--filter{% endif %}
+      samplehcluster.py {% for e in exps %}"{{e.exp.file}}" {% endfor %} --sampleids {% for e in exps %}{{e|sample_id}} {% endfor %}{% if genes %} --genes {{genes|join(" ")}}{% endif %}--linkage {{linkage}} --dstfunc {{distance}} {% if filter %}--filter{% endif %}
```
mstajdohar: It would be better to write genes to a file and then give the file to the `samplehcluster.py` script. This may fail for many genes. @jkokosar can give you some pointers on how to do that.

PR #188, opened 2017-01-18
Faster feature import

### LINKED ISSUE #186: Use `get_or_create` in `insert_features`management command
It is not recommended to use `try`/`except` in scenarios where failure in expected (handling errors takes much more time than other checks), so it would be much faster to use [`get_or_create`](https://docs.djangoproject.com/en/1.10/ref/models/querysets/#get-or-create) query and checks second parameter of the returned tuple, which tells if object was created or not.

mstajdohar:
Closes #186.

### DIFF
--- a/resolwe_bio/kb/management/commands/insert_features.py
+++ b/resolwe_bio/kb/management/commands/insert_features.py
@@ -113,9 +113,12 @@ class Command(BaseCommand):
                     'aliases': aliases,
                 }
 
-                try:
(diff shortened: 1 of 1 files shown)


mstajdohar:
Closes #186.

mstajdohar:
The [test_feature_search](https://travis-ci.org/genialis/resolwe-bio/jobs/192986168) fails in a couple of tests. @tjanez, @dblenkus, can you check why?

PR #193, opened 2017-01-20
Replace Haystack with Resolwe ES API

kostko:
The only missing thing is search in the admin interface which used the `SearchModelAdmin` from Haystack.

Will require `urls.py` update in base.


kostko:
@mstajdohar can you test if this breaks result ordering requirements?

resolwe_bio/kb/elastic_indexes.py
```diff
@@ -54,3 +81,35 @@ def process_object(self, obj, push=True):
...
+    target_id = dsl.String(index='not_analyzed')
+    relation_type = dsl.String(index='not_analyzed')
+
+    class Meta:
+        """Meta class for mapping search document."""
+
+        index = 'mapping_search'
+
+
+class MappingSearchIndex(BaseIndex):
```
dblenkus: Use bulk push:
* override `process_object` with `push=False`
* manually push documents in management command and tests
(see `FeatureSearchIndex` and `insert_features` command for example)

kostko: Done.

kostko:
Requires fixes from genialis/resolwe#198.

mstajdohar:
The gene autocomplete works great.

kostko:
Tests currently fail due to genialis/resolwe#199, should be fixed as soon as it is merged.

PR #196, opened 2017-01-25
Support Django 1.10

mstajdohar:
Depends on https://github.com/genialis/resolwe/pull/200


resolwe_bio/migrations/0007_update_checksum.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
```
dblenkus: You don't need to update checksums in resolwe-bio

PR #189, opened 2017-01-18
Hierarhical clustering

JenkoB:
Add feature hierarhical gene clustering
Fix hierarhical sample clustering

PR #199, opened 2017-01-25
Fix clustering

JenkoB:


JenkoB:
@hadalin assigned

resolwe_bio/tools/genehcluster.py
```diff
@@ -115,7 +115,7 @@
         'gene_symbols': calculated_genes_dict,
         'order': dend['leaves'],
         'zero_gene_symbols': list_zero_genes,
-        'zero_sample_ids': [args.sampleids[index] for index in sample_zero]
+        'zero_sample_ids': [int(args.sampleids[index]) for index in sample_zero]
```
hadalin: I'm not sure, but I think these are already integers. It's the [test](https://github.com/genialis/resolwe-bio/blob/master/resolwe_bio/tests/files/sample_cluster_data.json.gz) [files](https://github.com/genialis/resolwe-bio/blob/master/resolwe_bio/tests/files/gene_cluster_data.json.gz) that have strings.

hadalin: https://github.com/genialis/resolwe-bio/pull/189/files#diff-2dfb667aaa381ef6c7c5a97d8e79e6c4

JenkoB: I added also new test files, where this is changed.

hadalin:
Can you update commit message to _Fix clustering_ (without _1_ at the end).

PR #200, opened 2017-01-26
Prepare release 1.3.0

tjanez:
@jkokosar, please check if the Change Log includes all notable changes from release 1.2.1.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -84,14 +84,14 @@ From PyPI_
 
 .. code::
 
-    pip install resolwe-bio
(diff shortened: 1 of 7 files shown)


dblenkus:
Do we still need `--no-binary` argument?

tjanez:
> Do we still need --no-binary argument?

No, since we [decided to only release sdist version of Resolwe](https://github.com/genialis/resolwe/pull/205#issuecomment-275635710), that is no longer necessary. I've updated the PR.

PR #203, opened 2017-01-30
Fix parse_diffexp.py to handle adjusted p-vlaue and p-value when 0

JenkoB:
This needs to be handled on frontend when is -10log transformation on volcano plot. So this PR is not ok.


jkokosar:
@JenkoB Please submit an issue in the genialis-apps repository.

PR #202, opened 2017-01-28
Support resolwe test framework

mstajdohar:
Depends on https://github.com/genialis/resolwe/pull/208

PR #205, opened 2017-01-31
Use new import paths in resolwe

dblenkus:


mstajdohar:
I have canceled the tests. Will not pass without https://github.com/genialis/resolwe-bio/pull/202

mstajdohar:
But the https://github.com/genialis/resolwe-bio/pull/202 would not pass without this PR. I have merged the two into https://github.com/genialis/resolwe-bio/pull/202.

PR #215, opened 2017-02-14
Fix generate_samples command

dblenkus:


mstajdohar:
Rebase on to master. Check if we have to fix coverage parameters. Your code if OK, but the coverage results seam odd.

PR #216, opened 2017-02-15
Make tests use the live Resolwe API host instead of external server

tjanez:
Requires genialis/resolwe#220.

@jkokosar, could you take a look why the `test_go_enrichment` test returns different values for `pval` and `score` which cause the test to fail?

Here is the diff:
```
$ diff obtained-result.json result-from-file.json 
75,76c75,76
<                                                 "pval": 0.5333333333333332,
<                                                 "score": 0.75,
---
>                                                 "pval": 0.6666666666666671,
>                                                 "score": 1.5,
86,87c86,87
<                                         "pval": 0.5333333333333332,
<                                         "score": 0.75,
---
>                                         "pval": 0.6666666666666671,
>                                         "score": 1.5,
97,98c97,98
<                                 "pval": 0.5333333333333332,
<                                 "score": 0.75,
---
>                                 "pval": 0.6666666666666671,
>                                 "score": 1.5,
108,109c108,109
<                         "pval": 0.5333333333333332,
<                         "score": 0.75,
---
>                         "pval": 0.6666666666666671,
>                         "score": 1.5,
124,125c124,125
<                                         "pval": 0.5333333333333332,
<                                         "score": 0.75,
---
>                                         "pval": 0.6666666666666671,
>                                         "score": 1.5,
135,136c135,136
<                                 "pval": 0.39999999999999963,
<                                 "score": 0.5,
---
>                                 "pval": 1.0,
>                                 "score": 1.0,
146,147c146,147
<                         "pval": 0.39999999999999963,
<                         "score": 0.5,
---
>                         "pval": 1.0,
>                         "score": 1.0,
157,158c157,158
<                 "pval": 0.39999999999999963,
<                 "score": 0.5,
---
>                 "pval": 1.0,
>                 "score": 1.0,
171,172c171,172
<                         "pval": 0.5333333333333332,
<                         "score": 0.75,
---
>                         "pval": 0.6666666666666671,
>                         "score": 1.5,
182,183c182,183
<                 "pval": 0.5333333333333332,
<                 "score": 0.75,
---
>                 "pval": 0.6666666666666671,
>                 "score": 1.5,
```

### DIFF
--- a/resolwe_bio/tests/processes/test_enrichment.py
+++ b/resolwe_bio/tests/processes/test_enrichment.py
@@ -1,10 +1,11 @@
 # pylint: disable=missing-docstring
 from resolwe.test import tag_process
-from resolwe_bio.utils.test import KBBioProcessTestCase
+from resolwe_bio.utils.test import with_resolwe_host, KBBioProcessTestCase
 
 
 class EnrichmentProcessorTestCase(KBBioProcessTestCase):
 
+    @with_resolwe_host
     @tag_process('goenrichment')
     def test_go_enrichment_dicty(self):
         with self.preparation_stage():
@@ -25,6 +26,7 @@ class EnrichmentProcessorTestCase(KBBioProcessTestCase):
         self.assertEqual(len(enrichment.process_warning), 0)
         self.assertJSON(enrichment, enrichment.output['terms'], '', 'go_enriched_terms_dicty.json.gz')
 
+    @with_resolwe_host
     @tag_process('goenrichment')
     def test_go_enrichment(self):
         with self.preparation_stage():
--- a/resolwe_bio/utils/test.py
+++ b/resolwe_bio/utils/test.py
@@ -7,6 +7,7 @@ import wrapt
 
 from django.conf import settings
 from django.core.management import call_command
+from django.test import LiveServerTestCase
 
 from resolwe.test import with_resolwe_host as resolwe_with_resolwe_host, ProcessTestCase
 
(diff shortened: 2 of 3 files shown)


tjanez:
> @jkokosar, could you take a look why the test_go_enrichment test returns different values for pval and score which cause the test to fail?

The mystery has been solved (thanks @JenkoB and @jkokosar!). Apparently, I had some leftovers in my ElasticSearch or PostgreSQL container which caused the GO Enrichment process to obtain mappings for all genes instead of just one. This in turn cased different Enrichment results shown above.

I've added extra checks to both Enrichment processes' tests to make them more robust.

resolwe_bio/utils/test.py
```diff
@@ -45,12 +46,17 @@ def skipUnlessLargeFiles(*files):  # pylint: disable=invalid-name
     return lambda func: func
 
 
-class BioProcessTestCase(ProcessTestCase):
-    """Test class for bioinformatics processes."""
+class TransactionBioProcessTestCase(TransactionProcessTestCase):
+    """Base class for writing bioinformatics process tests not enclosed in a transaction.
```
dblenkus: Add a brief description when to use this class - i.e. for tests that uses API

resolwe_bio/utils/test.py
```diff
@@ -93,6 +99,19 @@ def prepare_expression(self, f_rc='exp_1_rc.tab.gz',
         return self.run_process('upload-expression', inputs)
 
 
+class BioProcessTestCase(TransactionBioProcessTestCase, DjangoTestCase):
```
dblenkus: Do we even need transaction and non-transaction tests cases? What are advantages of having two instead of just one (non-transaction)?

resolwe_bio/tests/processes/test_enrichment.py
```diff
@@ -21,6 +23,7 @@ def test_go_enrichment_dicty(self):
         enrichment = self.run_process('goenrichment', inputs)
         self.assertJSON(enrichment, enrichment.output['terms'], '', 'go_enriched_terms_dicty.json.gz')
 
+    @with_resolwe_api_host
```
dblenkus: Why don't you add this decorator to the `setUp` method, so all tests can use it?

tests/settings.py
```diff
@@ -10,9 +10,6 @@
 
 SECRET_KEY = 'secret'
 
-# TODO: Support local Resolwe server in tests
-RESOLWE_API_HOST = 'https://qa.genialis.com'
```
dblenkus: Change this to `RESOLWE_HOST_URL` and leave it in settings for other tests.

mstajdohar:
@tjanez Address Domen's comments.

mstajdohar:
Rebase this on to the master branch and address @dblenkus' comments.

dblenkus:
@tjanez please remove duplicated commits and rebase

resolwe_bio/utils/test.py
tjanez: Done.

resolwe_bio/utils/test.py
tjanez: I guess we want as many tests to be *vanilla*, i.e. do everything in a transaction that is then reverted.

But I'm open to arguments for using just the non-transaction one.

resolwe_bio/tests/processes/test_enrichment.py
tjanez: I think even better would be to extend the `with_resolwe_api_host` decorator to also work with the whole Class.

tests/settings.py
tjanez: I rather put it out and changed the `GenomeBrowserProcessorTestCase` to explicitly set it with `@override_settings(RESOLWE_HOST_URL='https://dummy.host.com')`

tjanez:
@dblenkus, please re-review.

PR #220, opened 2017-02-20
Add Vaccinesurvey sample schema

JureZmrzlikar:


resolwe_bio/descriptors/sample_vaccinesurvey.yml
```diff
@@ -0,0 +1,134 @@
...
+      group:
+        - name: annotator
+          label: Annotator/Scientist
+          type: basic:string
+          description: >
+            The name of the person that prepared the data.
+        - name: study_code
+          label: Study code identifying patient
+          type: basic:string
+        - name: sex
```
jkokosar: Could be "gender"

JureZmrzlikar: I just tried to be politically correct, since there are so many types of gender these days :-)

PR #221, opened 2017-02-20
Use Genialis' Jenkins instead of Travis CI and several other fixes

tjanez:


.travis.yml
```diff
@@ -1,64 +0,0 @@
-sudo: required
```
dblenkus: Remove `.travis.yml` from `.gitignore`.

docs/contributing.rst
```diff
@@ -92,7 +92,13 @@ Run docker::
...
 To run the tests, use::
 
-    ./manage.py test resolwe_bio --parallel=2
+    ./manage.py test resolwe_bio --parallel 2
+
+.. note::
+
+    If you don't specify the number of parallel test processes (i.e. you just
+    use ``--parallel``), Django will run one test process per each core
+    available on the machine.
```
dblenkus: It is not a good idea to run tests with `--parallel` on macOS (and probably Windows), where docker runs in virtual machine and becomes totally unresponsive even with 6 parallel tests (docker's vm has 6 cores and 10gb of ram, so bottle neck is probably network).

.travis.yml
tjanez: Fixed.

docs/contributing.rst
tjanez: Thanks for the tip. I'll warn users/developers on MacOS and Windows about this.

PR #222, opened 2017-02-21
Fix-pca

JenkoB:
fix pca to write the error if some of selected genes are not in expression files.


resolwe_bio/tools/pca.py
```diff
@@ -57,8 +57,11 @@ def isgzipped(f):
                     (l.split('\t') for l in f) if len(gene_exp) == 2 and isfloat(gene_exp[1])})
 
         allgenes.update(exp[-1].keys())
-
 if args.genes:
+    zero_genes = []
```
mstajdohar: How about this:
```
genes = set(args.genes)
zero_genes = genes.difference(allgenes)
```

mstajdohar:
Fix linters.

resolwe_bio/tools/pca.py
lukaw3d: Fixed

PR #223, opened 2017-02-21
Create IGV session for bam

JenkoB:


JenkoB:
@jkokosar @dblenkus please review

resolwe_bio/tests/processes/test_genome_browser.py
```diff
@@ -0,0 +1,26 @@
...
+            'genomeid': 'hg19',
+            'bam': [bam.pk, bam1.pk],
+            'locus': 'chr7:79439229-79481604'
+        }
+
+        igv_session = self.run_process('igv', inputs)
+
+        # remove changing lines from the output
+        def filter_created(line):
+            if b'<Resource' in line:
```
tjanez: This is too broad, since it will also match line ` <Resources>` in `resolwe_bio/tests/files/igv_session_bam.xml`. Try to be as specific as possible when you are filtering out lines to compare.

I would use something like: `if b'<Resource path=' in line:`.
Also, change filter's name to match what it is actually doing. I would suggest something like `filter_resource`.

resolwe_bio/tools/make_igv_session.py
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env python2
```
tjanez: Is there any reason to use Python 2 for new tools?

resolwe_bio/tools/make_igv_session.py
```diff
@@ -0,0 +1,35 @@
...
+# pylint: disable=missing-docstring,invalid-name,import-error
+"""Create IGV session."""
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import argparse
+from lxml import etree
+
+parser = argparse.ArgumentParser(description='Create igv session')
+
+parser.add_argument('-bam', '--bam_path', nargs='+', default=[], help='list of paths to bam file')
```
tjanez: Should this argument be positional and required?

Also make help text consistent and start with a capital letter.

resolwe_bio/tools/make_igv_session.py
```diff
@@ -0,0 +1,35 @@
...
+parser.add_argument('-genomeid', '--genomeid', help='genome')
```
tjanez: Hmm... the short flag is the same as the long one?

Also make help text more descriptive and start it with a capital letter.

resolwe_bio/tools/make_igv_session.py
```diff
@@ -0,0 +1,35 @@
...
+parser.add_argument('-locus', '--locus', help='Locus where to open IGV')
```
tjanez: Again, try to come up with a short flag that is not the same as the full argument name.

resolwe_bio/tools/make_igv_session.py
dblenkus: `lxml` is not installed on python3 in docker image.

dblenkus:
@tjanez Please review changes and merge.

resolwe_bio/tools/make_igv_session.py
```diff
@@ -0,0 +1,35 @@
...
+# pylint: disable=missing-docstring,invalid-name,import-error
+"""Create IGV session."""
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import argparse
+from lxml import etree
+
+parser = argparse.ArgumentParser(description='Create igv session')
+
+parser.add_argument('-bam', '--bam_path', nargs='+', default=[], help='List of paths to bam file')
```
tjanez: If I understand the script correctly, bam files are required?

So I suggest to make them a positional argument (which is always required).
Another option would be to use the `required` keyword argument when calling `add_argument()`.

Are other arguments optional or also always required?

Another thing: I wouldn't be afraid to use really short option flags, e.g. `-g` for genome id and `-l` for locus.

Also, we typically follow the pattern to use double quotes for strings meant for human consumption and use single quotes for everything else. Since arguments and help messages are things that users of the tool will see, put them in double quotes.

To sum up, you could replace these 3 lines with the following:
```python
parser.add_argument("bam_files", nargs='+', metavar="BAM_FILE", help="List of paths to bam files")
parser.add_argument("-g", "--genomeid", help="Genome id")
parser.add_argument("-l", "--bam_locus", help="Locus where to open IGV")
```

if only bam files are required arguments which are treated as positonal arguments.

The tool's help would now look like:
```
usage: make_igv_session.py [-h] [-g GENOMEID] [-l BAM_LOCUS]
                           BAM_FILE [BAM_FILE ...]

Create igv session

positional arguments:
  BAM_FILE              List of paths to bam files

optional arguments:
  -h, --help            show this help message and exit
  -g GENOMEID, --genomeid GENOMEID
                        Genome id
  -l BAM_LOCUS, --bam_locus BAM_LOCUS
                        Locus where to open IGV
```

resolwe_bio/tools/make_igv_session.py
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env python2
+# pylint: disable=missing-docstring,invalid-name,import-error
+"""Create IGV session."""
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import argparse
+from lxml import etree
+
+parser = argparse.ArgumentParser(description='Create igv session')
```
tjanez: As described below, use double quotes for description string.

resolwe_bio/tools/make_igv_session.py
```diff
@@ -0,0 +1,35 @@
...
+
+parser.add_argument('-bam', '--bam_path', nargs='+', default=[], help='List of paths to bam file')
+parser.add_argument('-genome', '--genomeid', help='Genome id')
+parser.add_argument('-locus', '--bam_locus', help='Locus where to open IGV')
+
+args = parser.parse_args()
+
+Global = etree.Element(
```
tjanez: As described above, use single quotes for everything in the lines below.

resolwe_bio/tools/make_igv_session.py
JenkoB: Thanks, for detail review. I have changed quotes type.

resolwe_bio/tools/make_igv_session.py
JenkoB: In the script all arguments are required. Genome id is required, to upload reference genome in IGV session together with bam sessions, and locus is by default set to All in yml. file. f I add metavar in arguments the xml file is not created in right way, that is why I set the required argument to True.

mstajdohar:
@tjanez Your comments have been addressed. Please review and merge.

PR #224, opened 2017-02-22
Bump upload-fasta and trimmomatic-paired process versions

jkokosar:


tjanez:
@jkokosar, the commit message `Bump upload-fasta and trimmomatic-paired process versions` only has 57 characters which is OK. I've changed it to that.

PR #225, opened 2017-02-27
Refactor amplicon workflow

jkokosar:


docs/CHANGELOG.rst
```diff
@@ -13,6 +13,7 @@ Unreleased
 Added
 -----
 * Amplicon workflow
+* Add amplicon descriptor schemas
```
tjanez: Make this consistent with the line above and just use `Amplicon descriptor schemas`

PR #230, opened 2017-03-03
Added transformation from Phred 64 to Phred 33

JenkoB:


JenkoB:
@jkokosar please review

resolwe_bio/tests/processes/test_upload.py
```diff
@@ -80,6 +80,7 @@ def test_upload_cuffquant_expr(self):
         self.run_process('upload-expression-cuffnorm', inputs)
 
     def test_upload_paired_end_reads(self):
+        self.keep_data()
```
mstajdohar: This should be removed.

resolwe_bio/processes/import_data/seq_reads.yml
```diff
@@ -8,7 +8,7 @@
   requirements:
     expression-engine: jinja
   data_name: '{{ src.0.file|default("?") }}'
-  version: 1.0.22
+  version: 1.0.23
```
mstajdohar: Increase the middle version number.

resolwe_bio/processes/import_data/seq_reads.yml
```diff
@@ -85,7 +91,7 @@
   requirements:
     expression-engine: jinja
   data_name: '{{ src1.0.file|default("?") }}'
-  version: 1.0.22
+  version: 1.0.23
```
mstajdohar: Increase the middle version number.

resolwe_bio/processes/import_data/seq_reads.yml
```diff
@@ -160,14 +161,53 @@
...
-        fi
+      NAMES_FW=($FW_FASTQC)
+
+      for name in ${NAMES_FW[@]}; do
+        READS_NAME_FW=`basename $name _fastqc.zip`
+
+        for report_dir in ./fastqc/"${READS_NAME_FW}_fastqc"/; do
+          ENCODING=`parse_encoding_type.py ${report_dir}fastqc_data.txt`
+
+          if [ "${ENCODING}" == "Illumina 1.5" ] ||  [ "${ENCODING}" == "Illumina 1.3" ]; then
```
mstajdohar: This IF statement is the same as for the single ended reads. We should write a bash script that checks the encoding version and does the transformation. The we should call this script from both processes.

Issue #228, opened 2017-03-02
Improve and enable test_amplicon_report test


tjanez:
Currently, `resolwe_bio.tests.processes.test_generate_report.ReportProcessorTestCase.test_amplicon_report` test is disabled due to requiring a custom Docker image.

I think we need to improve the following:

1. The test is quite slow. It takes **~35s** on my machine. I profiled the test line-by-line and obtained the following result:
    ```
    Timer unit: 1e-06 s
    
    Total time: 35.5463 s
    File: /home/tadej/Genialis/resolwe-bio/resolwe_bio/tests/processes/test_generate_report.py
    Function: test_amplicon_report at line 8
    
    Line #      Hits         Time  Per Hit   % Time  Line Contents
    ==============================================================
         8                                               @skipDockerFailure("Processor requires a custom Docker image.")
         9                                               @profile
        10                                               def test_amplicon_report(self):
        11         1       933503 933503.0      2.6          template = self.run_process('upload-file', {'src': 'report_template.tex'})
        12         1       898266 898266.0      2.5          logo = self.run_process('upload-file', {'src': 'genialis_logo.pdf'})
        13                                           
        14         1       936879 936879.0      2.6          bam = self.run_process('upload-bam', {'src': '56GSID_10k_trimmed.bam'})
        15         1       924495 924495.0      2.6          bed = self.run_process('upload-bed', {'src': '56g_targets_small.bed'})
        16                                           
        17         1      2076697 2076697.0      5.8          coverage = self.run_process('coveragebed', {'alignment': bam.id, 'bed': bed.id})
        18                                           
        19         1      2835931 2835931.0      8.0          genome = self.run_process('upload-genome', {'src': 'hs_b37_chr22_frag.fasta.gz'})
        20         1       890908 890908.0      2.5          bed_picard = self.run_process('upload-bed', {'src': '56g_targets_picard_small.bed'})
        21                                           
        22         1            4      4.0      0.0          inputs = {'src': 'Mills_and_1000G_gold_standard.indels.b37.chr22_small.vcf.gz'}
        23         1       905177 905177.0      2.5          indels = self.run_process('upload-variants-vcf', inputs)
        24                                           
        25         1       869405 869405.0      2.4          dbsnp = self.run_process('upload-variants-vcf', {'src': 'dbsnp_138.b37.chr22_small.vcf.gz'})
        26                                           
        27                                                   inputs = {
        28         1            5      5.0      0.0              'alignment': bam.id,
[...]

PR #234, opened 2017-03-08
Improve amplicon report

jkokosar:
Resolves https://github.com/genialis/resolwe-bio/issues/228


resolwe_bio/tools/make_report.py
```diff
@@ -66,6 +128,33 @@ def parse_targetPCRmetrics(metrics_report):
...
+def get_mean_coverage(covd):
+    """Calculate mean coverage from coverageBed (.covd) report file."""
+    with open(covd) as pbc:
+        cov = 0
+        lines = 0
+        for line in pbc:
+            cov = cov + int(line.strip().split('\t')[6])
+            lines = lines + 1
+
+    return float(cov) / lines
```
dblenkus: Use `len(pbc)` instead of `lines`.

resolwe_bio/tools/make_report.py
```diff
@@ -66,6 +128,33 @@ def parse_targetPCRmetrics(metrics_report):
         return dict(zip(labels, values))
 
 
+def get_mean_coverage(covd):
+    """Calculate mean coverage from coverageBed (.covd) report file."""
+    with open(covd) as pbc:
+        cov = 0
```
dblenkus: `cov = sum(int(line.strip().split('\t')[6]) for line in pbc)`

resolwe_bio/tools/make_report.py
dblenkus: You don't need to cast ints to floats, if you add `from __future__ import division`, as this is default behaviour in python 3. Transition to python 3 will also be easier, if you use `future` features.

resolwe_bio/tools/make_report.py
```diff
@@ -117,33 +193,64 @@ def make_scatterplot(x, y, hline=None, savefile=None, dpi=300):
...
-        cov_list, _ = parse_tsv_to_list(args.cov)
-        amplicons = [line[4] for line in cov_list]
-        x = list(range(len(amplicons)))
-
-        # Barplot:
-        cov1 = [float(line[8]) for line in cov_list]
-        make_barplot(x, cov1, savefile='barplot.png')
-        template = template.replace('{#IMAGE1#}', 'barplot.png')
+        cov_list, _ = _tsv_to_list(args.cov)
+        x = list(range(len(cov_list)))
```
dblenkus: Use more descriptive name

resolwe_bio/processes/exome_coverage/coveragebed.yml
```diff
@@ -0,0 +1,50 @@
...
+# Bedtools coverage (coveragebed)
+# ===============================
+---
+
+- slug: coveragebed
+  name: coverageBed
+  requirements:
+    expression-engine: jinja
+  data_name: "Coverage ({{ alignment|sample_name|default('?') }})"
+  version: 2.0.7
```
jkokosar: Also, process description is fixed!

PR #238, opened 2017-03-13
Add filtering option to star aligner

JenkoB:
@jkokosar  please review


jkokosar:
Set the default for the added parameter and add changelog.

JenkoB:
Resolved. No default option.

PR #242, opened 2017-03-15
Support resolwe toolkit

mstajdohar:
Depends on https://github.com/genialis/resolwe/pull/238


mstajdohar:
Tests pass locally. It seams Jenkins uses an old version of Resolwe. @tjanez, @dblenkus, please inspect.

PR #243, opened 2017-03-17
Fix sample desctiptor to show strain option for any organism

JenkoB:
@jkokosar please review


mstajdohar:
@jkokosar Please review.

PR #244, opened 2017-03-21
Wrap REF and ALT column (amplicon report)

JureZmrzlikar:


resolwe_bio/tools/make_report.py
```diff
@@ -3,6 +3,7 @@
 from __future__ import division
 
 import os
+import re
```
jkokosar: Unused import?

PR #247, opened 2017-03-28
Make PCA consistent with Hierarchical clustering

lukaw3d:
Previously, PCA returned `expression data ids` in a field `samples`,
while Hierarchical Clustering returned `sample ids` in a field `sample_ids`.


resolwe_bio/processes/clustering/pca.yml
```diff
@@ -45,18 +45,18 @@
...
       re-save proc.rc $?
       re-save proc.progress 0.1
 
-      obj_id=${obj_id%?}
-      exp=${exp%?}
+      sample_ids=${sample_ids%?}
+      exp_files=${exp_files%?}
 
       re-save proc.rc $?
       re-save proc.progress 0.2
```
JenkoB: For clustering it we added checking if is the source of selected genes the same as the source of expression files, dose it make sense to add this here?

mstajdohar:
Rebase to master

PR #248, opened 2017-03-30
Amplicon workflow v2

jkokosar:


resolwe_bio/processes/amplicon/bwa_trim.yml
```diff
@@ -0,0 +1,180 @@
+# ==========================================
+# Align reads and trimm resulting SAM output
+# ==========================================
+---
+
+- slug: align-bwa-trim
```
JenkoB: Why this can't be added to current bwa aligner? Is result SAM or BAM file?

jkokosar: It trims primers after aligment directly from SAM file, only than the file is converted to BAM.

mstajdohar: Could this be an option in the current bwa aligner? For instance, trim primers if master file given.

resolwe_bio/processes/amplicon/bwa_trim.yml
```diff
@@ -0,0 +1,180 @@
...
+  version: 0.0.1
+  type: data:alignment:bam:bwatrim
+  category: analyses
+  flow_collection: sample
+  persistence: CACHED
+  description: >
+    Align with BWA mem and trim the sam output.
+  input:
+    - name: master_file
+      label: Master file
```
mstajdohar: Provide a better description. Not obvious what this is for.

resolwe_bio/tools/plotcoverage.py
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/env python2
```
mstajdohar: We should start using Python 3.

mstajdohar:
Fix tests

PR #257, opened 2017-04-19
Change iCount-demultiplex tool persistence to TEMP

jkokosar:


jkokosar:
To have this process run on a separate, high priority worker. This process triggers a new process via resdk - new processes must finished for the parent process to finish also. If only one worker is available, this gets you in the deadlock situation.

mstajdohar:
Fix tests.

PR #262, opened 2017-04-24
Fix filters

hadalin:
The question is are these patches even necessary? Why not use `entity` in the first place or is `sample` used for backwards compatibility?


jkokosar:
@hadalin @dblenkus What is the status of this PR? If it looks ok, please rebase and merge.

PR #264, opened 2017-04-25
Add IGV to archive samples process

JenkoB:
@jkokosar @mstajdohar please review


resolwe_bio/tools/make_igv_session_archive.py
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env python3
+# pylint: disable=missing-docstring,invalid-name,import-error
+"""Create IGV session for archive."""
+from __future__ import absolute_import, division, print_function, unicode_literals
```
jkokosar: Since you are using Python3, this import is not required.

resolwe_bio/tools/make_igv_session_archive.py
```diff
@@ -0,0 +1,38 @@
...
+
+parser = argparse.ArgumentParser(description='Create igv session')
+
+parser.add_argument("-f", "--files", nargs='+', required=True, help="List of files")
+
+args = parser.parse_args()
+
+Global = etree.Element(
+    'Global',
+    genome='hg19',
```
jkokosar: Expose this as a function parameter, so we can prepare different archiver processes that use different genomes.

JenkoB:
Fixed.

mstajdohar:
Rebase please

PR #265, opened 2017-04-25
Change rna-seq schema (field starnded)

JenkoB:
@jkokosar please review

### DIFF
--- a/resolwe_bio/descriptors/bbduk-star-htseq.yml
+++ b/resolwe_bio/descriptors/bbduk-star-htseq.yml
@@ -15,7 +15,7 @@
 
 - slug: rna-seq-quantseq
   name: RNA-Seq
-  version: 0.0.1
+  version: 0.0.2
   description: RNA-Seq template
   schema:
     - name: trimming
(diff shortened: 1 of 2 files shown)


resolwe_bio/descriptors/bbduk-star-htseq.yml
```diff
@@ -38,9 +38,9 @@
       type: basic:string
       default: 'yes'
       choices:
-        - label: 'Yes'
+        - label: 'Forward'
```
jkokosar: Shouldn't this be just "Stranded"? Also, consider adding description to this field. From the htseq-count manual:

-s <yes/no/reverse>, --stranded=<yes/no/reverse>
whether the data is from a strand-specific assay (default: yes)

For stranded=no, a read is considered overlapping with a feature regardless of whether it is mapped to the same or the opposite strand as the feature. For stranded=yes and single-end reads, the read has to be mapped to the same strand as the feature. For paired-end reads, the first read has to be on the same strand and the second read on the opposite strand. For stranded=reverse, these rules are reversed.

PR #268, opened 2017-05-10
Add Dicty primary analysis schema

jkokosar:


resolwe_bio/descriptors/dicty.yml
```diff
@@ -0,0 +1,12 @@
+- slug: dicty-align-reads
+  name: Align reads (dicty)
+  version: 0.0.1
+  description:  Dicty primary analysis schema.
```
mstajdohar: This description will be displayed on the first recipe card. I suggest not to use the word schema, but rather explain the pipeline.

PR #277, opened 2017-05-22
Derive output filenames in hisat2 from reads filename

mzganec:


mstajdohar:
Write tests (examples that have failed before), mention the fix in the change log and review the commit message—it's quite long.

dblenkus:
Commit message should be limited to 70 (as advised by @tjanez), so length is not a problem. But try to be more specific, i.e. which process have you fixed,... And I also don't see nothing related to the "avoid **copying reads** before alignment" part.

mzganec:
@mstajdohar The test now provides an example that fails using the original hisat2 process.
@dblenkus You are correct, copying reads before alignment was related to an intermediate, but now obsolete version.

mzganec:
Please merge into master.

PR #280, opened 2017-05-30
Fix protocol labels in rna-seq-quantseq descriptor schema

mzganec:
https://github.com/genialis/bug-reports/issues/15
Fix protocol labels following the suggestion by @rafecooks.


mzganec:
Please review and merge with master.

PR #281, opened 2017-06-01
Fix quotation marks in rnaseq workflow

mzganec:
When selecting rnaseq workflow in Bioinformatics Dashboard / Tool Catalog, the unquoted default values _yes_ and _no_ make the corresponding input fields display incorrectly on the front-end. The tests run correctly in both versions of the workflow.

### DIFF
--- a/resolwe_bio/processes/workflows/rnaseq.yml
+++ b/resolwe_bio/processes/workflows/rnaseq.yml
@@ -4,7 +4,7 @@
   data_name: "Pipeline ({{ reads|sample_name|default('?') }})"
   requirements:
     expression-engine: jinja
-  version: 0.0.1
+  version: 0.0.2
   type: data:workflow:rnaseq:htseq
   input:
     - name: reads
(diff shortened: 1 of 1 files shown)


mstajdohar:
@hadalin Please investigate why this happens in UI and not in tests.

dblenkus:
If there are no quotes, Python tries to guess right type, thus interpreting `no` as `False`. Quotes must be added to explicitly define value as string.

    In [1]: import yaml

    In [2]: yaml.load('test: no')
    Out[2]: {'test': False}

    In [3]: yaml.load('test: "no"')
    Out[3]: {'test': 'no'}

mstajdohar:
@dblenkus Why does it work in tests?

mstajdohar:
I guess the UI control expects string input for a list of possible values? Tests don't care.

dblenkus:
It depends on weather specific tool interprets `"no"` and `False` in the same way.
But in either case it is much better to actually have the type that you expect :)

mzganec:
Please merge into master.

PR #283, opened 2017-06-01
Add amplicon table and jbrowse tracks

jkokosar:


resolwe_bio/processes/jbrowse/variants.yml
```diff
@@ -0,0 +1,39 @@
+# =========================
+# JBrowse tracks - Variants
+# =========================
+#
+---
+
+- slug: jbrowse-vcf-track
```
mstajdohar: Why can't we display the `data:variants:vcf` directly in JBrowse? If the JBrowse can display variants directly, wouldn't it be better to support it without a special track? The same way we support BAM data objects directly.

resolwe_bio/tools/make_amplicon_table.py
```diff
@@ -0,0 +1,172 @@
+#!/usr/bin/env python2
```
mstajdohar: I would like we start using Python 3 as soon as possible and add the required packages to Python 3. The longer we use Python 2, the harder will be to switch.

resolwe_bio/tools/make_amplicon_table.py
```diff
@@ -0,0 +1,172 @@
...
+
+import argparse
+import json
+
+import HTSeq
+
+
+def parse_arguments():
+    """Parse command line arguments."""
+    parser = argparse.ArgumentParser(description="Parse Diff Exp output files")
```
mstajdohar: Suggested description text: Parse the output files of differential expression

resolwe_bio/tools/make_amplicon_table.py
```diff
@@ -0,0 +1,172 @@
...
+                int(amplicon[1]),
+                int(amplicon[2])
+            )
+
+    with open(args.amplicon_coverage) as ac:
+        for line in ac:
+            cov = line.strip().split('\t')
+            amp_cov[cov[0]] = float(cov[1])
+
+    for annot_var in args.variants:
```
mstajdohar: How many variants do we usually have?

resolwe_bio/processes/jbrowse/variants.yml
lukaw3d: I don't like the way we support BAMs, but it does save disk space.
Having a special case for VCF too would mean making another additional request to the server (or a new custom endpoint), instead of being clear `query(type_startswith = data:jbrowse:)`

mstajdohar: I understand your concerns, but a general solution could visualize all variant files on the server, whereas the optimized solution requires a special pipeline. @jkokosar Do you see an advantage in generality?

jkokosar: The vast majority of VCFs will be produced within pipelines, so there is no major downside to having a special track, since this can be automated in workflows. The issue is if one would upload VCF directly and would need to remember to make this additional track before being able to visualize. For only this cases there is a decision of practicality  vs. efficiency to be made.

resolwe_bio/tools/make_amplicon_table.py
jkokosar: < 500 variants

jkokosar: This argument passes one or multiple _variant files_ (2 in this use case). There is no issue with the character limit.

resolwe_bio/tools/make_amplicon_table.py
```diff
@@ -0,0 +1,172 @@
...
+                    break
+
+            if 'DP4' in variant:
+                process = 'LoFreq'
+                other_fields = 'DP4={};SB={}'.format(variant['DP4'], variant['SB'])
+            else:
+                process = 'GATK'
+                other_fields = 'AD:{};FS:{}'.format(variant['GEN[0].AD'], variant['FS'])
+
+            if '' in variant['ID'].split(';'):
```
JureZmrzlikar: In case `variant['ID']` is equal to `'ABC;'`, you can loose some data. It is probably better to check:
`if variant['ID'].strip(';') == '':`

mstajdohar:
Rebase please

resolwe_bio/processes/amplicon/amplicon_table.yml
```diff
@@ -0,0 +1,43 @@
...
+    - name: variant_table
+      label: Variant table
+      type: basic:json
+  run:
+    runtime: polyglot
+    language: bash
+    program: |
+      NAME='{{ coverage|sample_slug }}'
+
+      make_amplicon_table.py {{master_file.master_file.file}} {{coverage.mean_cov.file}} \
```
mzganec: Filenames should be wrapped in quotes to correctly process spaces.

resolwe_bio/processes/amplicon/amplicon_table.yml
```diff
@@ -0,0 +1,43 @@
...
+      -v {% for vars in annot_vars %}"{{vars.annotation.file}}" {% endfor %}
+      re-checkrc
+
+      mv amplicon_table.json ${NAME}_amplicon_table.json
+      re-save variant_table ${NAME}_amplicon_table.json
```
mzganec: Replace `${NAME}_amplicon_table.json` with `"${NAME}_amplicon_table.json"` in both lines.

jkokosar: Not required here, since ${NAME} is derived from {{Data|sample_slug}} which can never contain whitespace.

jkokosar: This would be an issue if one would use {{Data|sample_name}}

PR #284, opened 2017-06-02
Miscellaneous testing improvements

tjanez:
@jkokosar, it appears that `resolwe_bio/tests/files/56GSID_10k_trimmed.bam` is not used anymore so I deleted it. Is that ok?


mstajdohar:
Rebase

PR #287, opened 2017-06-06
WIP: Add test for RNA-seq pipeline

mzganec:
The newly introduced `workflow_meta_rnaseq` has the same inputs as `workflow-rnaseq-paired` which is already present in the master branch. In the case where single reads are provided as an input to the meta-process, `palindrome_clip_threshold` is disregarded in `rnaseq.py` and can be set to hidden in the YAML file of the meta-process when this feature is available (see [https://github.com/genialis/genialis-apps/issues/589](https://github.com/genialis/genialis-apps/issues/589)). The included descriptor schema can be used with `workflow-rnaseq-single`, `workflow-rnaseq-paired`, and `workflow_meta_rnaseq`.

### DIFF
--- a/resolwe_bio/processes/workflows/rnaseq.yml
+++ b/resolwe_bio/processes/workflows/rnaseq.yml
@@ -229,3 +229,76 @@
           mode: union
           stranded: '{{input.stranded}}'
           id_attribute: '{{input.id_attribute}}'
+
+- slug: dss-rnaseq
+  name: RNA-Seq for biologists
+  requirements:
+    expression-engine: jinja
+    resources:
+      network: true
+  data_name: "Pipeline ({{ reads|sample_name|default('?') }})"
+  version: 0.0.1
+  type: data:dss:rnaseq
+  category: dss
+  persistence: RAW
+  description: >
+    Run RNA-Seq workflow
+  input:
+    - name: reads
+      label: Input reads
+      type: data:reads:fastq
+    - name: genome_and_annotation
+      label: Organism
+      type: basic:string
+      default: hs
+      choices:
+        - label: Homo sapiens
+          value: hs
+        - label: Mus musculus
+          value: mm
+    - name: adapters
+      label: Adapters
+      type: basic:string
+      default: 'yes'
+      choices:
+        - label: Remove
+          value: 'yes'
+        - label: Do not remove
+          value: 'no'
+    - name: minlen
+      label: Minimum read length
+      type: basic:integer
+      default: 10
+      description: >
+        Trimmed reads shorter than the minimum read length will be discarded.
+    - name: trailing
+      label: Trailing quality
+      type: basic:integer
+      default: 28
+      description: >
+        The minimum quality required to keep a base. Bases with lower quality
+        will be removed from the end of reads.
+    - name: stranded
+      label: Assay type
+      type: basic:string
+      default: 'no'
+      choices:
+        - label: Strand non-specific
+          value: 'no'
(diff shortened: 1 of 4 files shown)


mzganec:
Please review and merge into master.

resolwe_bio/processes/workflows/rnaseq.yml
```diff
@@ -229,3 +229,103 @@
           mode: union
           stranded: '{{input.stranded}}'
           id_attribute: '{{input.id_attribute}}'
+
+- slug: workflow-meta-rnaseq
```
mstajdohar: Call this dss-rnaseq. You can read about decision support systems here: https://en.wikipedia.org/wiki/Decision_support_system. Instead of a `dss` prefix, maybe we could use `dsc` (decision support component) or `dsp` (decision support process).

resolwe_bio/processes/workflows/rnaseq.yml
```diff
@@ -229,3 +229,103 @@
...
+  name: RNA-Seq workflow (meta)
```
mstajdohar: RNA-Seq for Biologists

resolwe_bio/processes/workflows/rnaseq.yml
```diff
@@ -229,3 +229,103 @@
...
+  requirements:
+    expression-engine: jinja
+  data_name: "Pipeline ({{ reads|sample_name|default('?') }})"
+  version: 0.0.1
+  type: data:workflow:rnaseq:htseq
```
mstajdohar: I suggest `data:dss:rnaseq` (of `data:dsp:rnaseq`). Depends on what prefix we choose.

resolwe_bio/processes/workflows/rnaseq.yml
```diff
@@ -229,3 +229,103 @@
...
+  category: workflow:primary
```
mstajdohar: `dss`

resolwe_bio/processes/workflows/rnaseq.yml
```diff
@@ -229,3 +229,103 @@
...
+  persistence: RAW
+  description: >
+    Run RNA-Seq workflow
+  input:
```
mstajdohar: This guys should have simple inputs, actually the same as the descriptor schema. The inputs should make sense to biologists.

resolwe_bio/processes/workflows/rnaseq.yml
```diff
@@ -229,3 +229,103 @@
...
+    - name: reads
+      label: Input reads
+      type: data:reads:fastq
+    - name: genome
```
mstajdohar: So a simple input instead of this one would be:
```
    - name: genome_and_annotation
       label: Organism
       type: basic:string
       default: hs
       choices:
         - label: Homo sapiens
           value: hs
         - label: Mus musculus
           value: mm
```
Then you should fetch the correct genome in the process itself (using resdk). The logic which genome to use should be part of the process. We want to remove the decision logic from the configuration.

resolwe_bio/tools/rnaseq.py
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env python
```
mstajdohar: User Python 3 wherever possible or we'll have a lot of problems in the near future.

mzganec:
Depends on [https://github.com/genialis/resolwe-bio/pull/323](https://github.com/genialis/resolwe-bio/pull/323).

PR #291, opened 2017-06-12
Set process scheduling class and network requirements

jkokosar:
"Gene info" processes would require a network access. Since this processes are obsolete (replaced by the KB functionality) I have removed them.


kostko:
Looks good, just be sure that all interactive processes take less than 30 seconds to run.

jkokosar:
@kostko Is a process killed after 30s?

kostko:
> @kostko Is a process killed after 30s?

Yes, an interactive process should be killed if it exceeds 30s of CPU time. If the process takes longer, then it should be considered a batch process and moved to that scheduling class.

kostko:
BTW, can you also make sure that previous hacks using `persistence` for getting higher priority are removed? Because this behavior will be removed in the future (e.g., `persistence` will have no effect on process execution priority).

resolwe_bio/processes/support_processors/archive.yml
```diff
@@ -6,10 +6,11 @@
     executor:
       docker:
         image: "resolwebio/archive-samples"
-  version: 0.0.2
+  version: 0.0.3
   type: "data:archive:samples"
   category: other
   persistence: TEMP
+  scheduling_class: interactive
```
mstajdohar: @kostko We need this for the [archiver process](https://github.com/genialis/resolwe/blob/master/resolwe/toolkit/processes/archiver.yml) in resolwe too.

kostko: Is the archiver really an interactive process? What if there is a lot of stuff to archive?

mstajdohar: Users are waiting for results... If it's not interactive, it's useless.

jkokosar: This process should start instantaneously, but can still take more than 30s, since it is possible to select/bundle larger sets of data into tar.gz How we should go about that?

kostko: > Users are waiting with an open dialog for the achieve

Exactly my point. So is it ok if they wait 30 seconds? 60 seconds? How long is acceptable for an interactive process? Currently, 30 seconds is the limit and anything longer than that is killed. We need to set a hard limit on CPU time for interactive processes, otherwise this will just be abused to get priority. cc @tjanez @dblenkus

tjanez: I think it would be naive to treat the archiver-like processes as interactive. Of course, if users "behave" and only download/create small archives, this will work, but in general, archives could be arbitrarily large.

The current implementation is actually vulnerable to filling-up the high priority Celery queue with archiver processes' tasks and all other (*really*) interactive processes will have to wait. Do we want that? I guess not.

Once we agree that archiver processes should not be treated as interactive, we can start discussing how to properly solve the archiver processes' use case.

mstajdohar: Let's make this arciver interactive now for the sake of user experience. Let's not make the resolwe archiver interactive. We can implement a better solution later. Wenshould text how long does it take on average to make the archieve in our user workflow. If it's well below 30s, this solutuion may stick for a while.

PR #295, opened 2017-06-20
Add new inputs to import SRA process

mzganec:
The new inputs enable fine-tuning the import process. The immediate use case is to specify _Maximum spot ID_ to limit the number of downloaded reads and thus speed up the process.

In the future a parallel fastq-dump (see [here](https://github.com/rvalieris/parallel-fastq-dump)) can be used to reduce the download time (not implemented yet).


dblenkus:
Are there common use-cases for added arguments?

I think that we should think which parameters we want to add. As the idea of the platform is to be easy to use and intuitive, there should be as little choices as possible.

mzganec:
Following @jkokosar's suggestion I'll add an advanced checkbox which will reveal additional inputs.

jkokosar:
@mzganec please rebase

PR #297, opened 2017-06-22
Modify chip-seq processes

jvrakor:
@dblenkus  @jkokosar

### DIFF
--- a/resolwe_bio/processes/alignment/bowtie.yml
+++ b/resolwe_bio/processes/alignment/bowtie.yml
@@ -231,8 +231,10 @@
   name: Bowtie
   requirements:
     expression-engine: jinja
+    resources:
+      cores: 1
   data_name: "Alignment ({{ reads.fastq.0.file|basename|default('?') }})"
-  version: 1.0.22
+  version: 1.0.23
   type: data:alignment:bam:bowtie2
   category: analyses:alignment
   flow_collection: sample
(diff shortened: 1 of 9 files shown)


resolwe_bio/processes/alignment/bowtie.yml
```diff
@@ -358,6 +358,12 @@
           description: >
             searches for at most X distinct, valid alignments for each read. The search terminates when it can't find more distinct valid alignments, or when it finds X, whichever happens first.          default: 5
           default: 5
+    - name: threads
```
jkokosar: Set this up in the process requirements section. Access the value set for a process from a `{{ requirements.resources.cores }}` variable

mstajdohar: @kostko Will this work?

kostko: Yes, this should work if the given requirement is set.

Note that I remember a discussion about moving this `requirements` into a different namespace due to including `contributor` as well. In that case, this will have to be updated as well.

resolwe_bio/processes/alignment/bowtie.yml
```diff
@@ -231,8 +231,10 @@
   name: Bowtie 2.2.3
```
jkokosar: Please drop the Bowtie version from the process name. Bowtie might get updated in the docker image, but the process name will stay the same.

resolwe_bio/processes/alignment/bowtie.yml
```diff
@@ -49,7 +49,7 @@
       description: >
```
jkokosar: Bump the process version for both bowtie processes

PR #299, opened 2017-06-26
Set log(fc) default to 1 in diffexp descriptor schema

jvrakor:
@jkokosar @mstajdohar


mstajdohar:
Merge the commits. The change log entry should be part of the change.

dblenkus:
And update the process version.

PR #300, opened 2017-06-29
Add memory as a resource in star and jbrowse annotation(gff3, gtf)  processes

jvrakor:
@jkokosar @mstajdohar


resolwe_bio/processes/alignment/star.yml
```diff
@@ -7,8 +7,10 @@
   name: STAR
   requirements:
     expression-engine: jinja
+    resources:
+      memory: 32768
```
mstajdohar: @dblenkus At some point we should support defining this with the unit, e.g. 32GB, and update the parser.

PR #301, opened 2017-06-29
Add S. tuberosum organism choice to RNA-seq workflow

mzganec:


mstajdohar:
Included in https://github.com/genialis/resolwe-bio/pull/305

PR #302, opened 2017-06-29
Fix typo in trimmomatic and bbduk processes

jvrakor:
@mstajdohar @jkokosar


mstajdohar:
Update the process versions and note the change in the change log.

jvrakor:
Included in https://github.com/genialis/resolwe-bio/pull/297

PR #303, opened 2017-06-29
Remove coverage report from tests

dblenkus:
Coverage is no longer used and is even not configured, so it is ok to
remove it.


mstajdohar:
Merged.

PR #308, opened 2017-07-04
Update processes to support new input sanitization

kostko:
Requires genialis/resolwe#306

TODO
* [x] Fix remaining 19 tests.
* [x] Rebase as some processes have been moved to other repositories.

### DIFF
--- a/resolwe_bio/processes/WGBS/bsmap.yml
+++ b/resolwe_bio/processes/WGBS/bsmap.yml
@@ -8,7 +8,7 @@
   requirements:
     expression-engine: jinja
   data_name: "bsmap ({{ reads|sample_name|default('?') }})"
-  version: 1.0.1
+  version: 1.0.2
   type: data:alignment:bam:bsmap
   category: analyses:wgbs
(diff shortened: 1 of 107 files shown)


kostko:
All tests fixed (with Resolwe from the linked PR), please review. I noticed several issues in general with Resolwe-bio, so this includes some other fixes as well (refactoring processes to not have overly long lines or doing strange things with bash). I didn't fix these issues in all the processes, I just went through those processes that needed changes due to new input sanitization code.

I hope that we have tests for all processes, as otherwise this may break badly or produce incorrect results. So all process authors, please review.

kostko:
Yes, I've mostly refactored things where tests failed, but didn't look so much into processes where tests passed after the initial quote replacement. So there are more cases where at least the following should be done:
* Use arrays instead of string concatenation.
* Break long argument lists into multiple lines to improve readability.

kostko:
Can I go ahead and merge this?

mstajdohar:
Yes

PR #311, opened 2017-07-13
Add basic knn leaf ordering for gene hierarchical clustering

mzganec:


mstajdohar:
This knn is quite an elegant solution. I am amazed it works so well.

PR #317, opened 2017-07-26
Fix NaN in explained variance in PCA

mzganec:


resolwe_bio/tools/pca.py
```diff
@@ -114,7 +114,7 @@ def save_pca(coordinates, explained_variance_ratios=[0, 0], components=[[], []],
     )
     sys.exit(0)
 
-pca = PCA(n_components=0.99, whiten=True)
+pca = PCA(n_components=1.0 - sys.float_info.epsilon, whiten=True)
```
mstajdohar: Write a comment

mstajdohar: cc @dblenkus, @jberci You'll love this fix. ;)

PR #318, opened 2017-07-27
Improve variant table name in amplicon report

jkokosar:


resolwe_bio/tools/make_report.py
```diff
@@ -129,6 +129,16 @@ def parse_target_pcr_metrics(metrics_report):
         return dict(zip(labels, values))
 
 
+def vcf_table_name(vcf):
```
mstajdohar: I suggest the make the input more verbose, e.g. vcf_name, vcf_file_name, file_name.

resolwe_bio/tools/make_report.py
```diff
@@ -129,6 +129,16 @@ def parse_target_pcr_metrics(metrics_report):
...
+    """Create caption for VCF table."""
```
mstajdohar: Maybe change to Format VCF table caption

PR #321, opened 2017-07-28
Enable network access for all import processes

jkokosar:


resolwe_bio/processes/import_data/etc.yml
```diff
@@ -7,8 +7,10 @@
   name: Expression time course
   requirements:
     expression-engine: jinja
+    resources:
+      network: true
```
mstajdohar: Check if the network resource is really needed as you are not using re-import here. Either support re-import or remove the network requirement.

resolwe_bio/processes/import_data/geneset.yml
```diff
@@ -61,8 +63,10 @@
   name: Create gene set
   requirements:
     expression-engine: jinja
+    resources:
+      network: true
```
mstajdohar: Check if the network resource is really needed as you are not using re-import in this process.

resolwe_bio/processes/import_data/hmmer_database.yml
```diff
@@ -7,8 +7,10 @@
   name: Upload HMMER database
   requirements:
     expression-engine: jinja
+    resources:
+      network: true
```
mstajdohar: Check if the network resource is really needed as you are not using re-import in this process. Either support re-import or remove the network requirement.

PR #328, opened 2017-08-10
Fix import-sra to work with input sanitization fix and add test

mzganec:
Caching option in `fastq-dump` was changed from disabled to enabled which speeds up download of reads.


resolwe_bio/processes/import_data/sra_file.yml
```diff
@@ -156,26 +156,33 @@
     runtime: polyglot
     language: bash
     program: |
-
+      CACHE=$PWD/fastq_cache
```
kostko: This is incorrect as `$PWD` may contain spaces. Please quote the string in double quotes.

resolwe_bio/processes/import_data/sra_file.yml
```diff
@@ -156,26 +156,33 @@
...
+      mkdir $CACHE
```
kostko: Quote the argument in double quotes.

resolwe_bio/processes/import_data/sra_file.yml
```diff
@@ -156,26 +156,33 @@
...
       mkdir $HOME/.ncbi/
```
kostko: Quote the argument in double quotes.

resolwe_bio/processes/import_data/sra_file.yml
```diff
@@ -287,15 +294,25 @@
     runtime: polyglot
     language: bash
     program: |
-
+      CACHE=$PWD/fastq_cache
```
kostko: Same as above.

mstajdohar:
Tests fail

mzganec:
@mstajdohar Tests fail because of an error on Jenkins that needs to be fixed by someone with sufficient credentials. Tests fail early, before SRA import test even starts (see console output).

mstajdohar:
cc @tjanez

tjanez:
@mzganec, @mstajdohar, fixed.

PR #333, opened 2017-08-14
Add Odocoileus virginianus texanus organism to sample descriptor

jvrakor:
@jkokosar @mstajdohar


resolwe_bio/descriptors/sample_geo.yml
```diff
@@ -31,6 +31,8 @@
               value: Rattus norvegicus
             - label: Solanum tuberosum
               value: Solanum tuberosum
+            - label: Odocoileus virginianus texanus
```
jkokosar: Please sort this addition alphabetically in the list of choices. In this way it will be correctly sorted also in the drop-down selection box.

PR #334, opened 2017-08-16
Fix multiple adapter sequence input in cutadapt

jvrakor:
@jkokosar @mzganec @mstajdohar


jvrakor:
No, because it is a nucleic acid sequence.

mzganec:
What is the purpose of these two lines [https://github.com/jvrakor/resolwe-bio/blob/49ca4e6ff6240b70764632c453f6661e5d74c6bc/resolwe_bio/processes/reads_processing/cutadatp.yml#L409](https://github.com/jvrakor/resolwe-bio/blob/49ca4e6ff6240b70764632c453f6661e5d74c6bc/resolwe_bio/processes/reads_processing/cutadatp.yml#L409) (`zcat` followed by `gzip`)? Similarly for reverse reads. The `gzip` line fails in tests with `gzip: in_fw_reads.fastq.gz already exists;	not overwritten`.

mzganec:
While you are at it, you can also fix the filename (`cutadatp.yml` -> `cutadapt.yml`).

resolwe_bio/processes/reads_processing/cutadapt.yml
```diff
@@ -405,12 +405,10 @@
       # Run Cutadapt on both reads
       {% if minlen %}
         if [ -f ${NAME1}_trimmed.fastq.gz ]; then
-         zcat ${NAME1}_trimmed.fastq.gz > in_fw_reads.fastq
-         gzip in_fw_reads.fastq
+         cat ${NAME1}_trimmed.fastq.gz > in_fw_reads.fastq.gz
```
mzganec: Reads are often large, so it would make sense to just move the file `mv ${NAME1}_trimmed.fastq.gz in_fw_reads.fastq.gz` instead of copying. The same goes for the reverse reads.

resolwe_bio/processes/reads_processing/cutadapt.yml
```diff
@@ -10,7 +10,7 @@
   requirements:
```
mzganec: In other processes we use a hyphen (single-end, paired-end).

resolwe_bio/processes/reads_processing/cutadapt.yml
jvrakor: Good point! Thanks

mstajdohar:
The test `ReadsFilteringProcessorTestCase` fails. Please review.

mstajdohar:
Do not forget to rebase afterwards.

PR #335, opened 2017-08-16
Add resolwebio/{base,legacy} docker images

kostko:
Requires genialis/resolwe#350.

* [x] Add R packages.
* [x] Remove all packages related to BAM Liquidator (which is no longer present in this image).
* [x] Update processes to use the new image.
* [x] Fix remaining tests.


kostko:
I've now added a base `resolwebio/base` Docker image, which just contains some `ONBUILD` instructions that build the image based on some configuration files. In this way, we do not need to repeat the same build and cleanup instructions in every image, but instead can just provide some configuration files, which are used to build the image.

For building manual packages, all images can also reuse some helper functions for automatic downloading, SHA256 verification, unpacking and configuring additional binary paths.

resolwe_bio/docker_images/base/Dockerfile.ubuntu-14.04
```diff
@@ -0,0 +1,77 @@
...
+    # Install Python packages.
+    if [ -f /var/cache/build/packages-python2.txt ]; then \
+        echo "Installing Python 2 packages..." && \
+        pip install --require-hashes -r /var/cache/build/packages-python2.txt \
+        ; \
+    fi && \
+
+    if [ -f /var/cache/build/packages-python2-stage2.txt ]; then \
+        echo "Installing stage 2 Python 2 packages..." && \
+        pip install --require-hashes -r /var/cache/build/packages-python2-stage2.txt \
```
tjanez: Hmm... this seems a bit awkward. Is `pip` really incapable of installing all Python packages with one call?

resolwe_bio/docker_images/base/Dockerfile.ubuntu-14.04
```diff
@@ -0,0 +1,77 @@
...
+        echo "Installing stage 2 Python 3 packages..." && \
+        pip3 install --require-hashes -r /var/cache/build/packages-python3-stage2.txt \
+        ; \
+    fi && \
+
+    # Build manual packages.
+    if [ -f /var/cache/build/packages-manual.txt ]; then \
+        echo "Building manual packages..." && \
+        sed -e 's/^#.*$//g' -e '/^$/d' /var/cache/build/packages-manual.txt | \
+            xargs -n 1 -I '{}' bash '/var/cache/build/packages-manual/{}.sh' \
```
tjanez: Here it would be even more useful if you used the long option names.

resolwe_bio/docker_images/base/Dockerfile.ubuntu-14.04
```diff
@@ -0,0 +1,77 @@
...
+        ; \
+    fi && \
+
+    # Cleanup packages.
+    echo "Cleaning up packages..." && \
+
+    if [ -f /var/cache/build/packages-debian-cleanup.txt ]; then \
+        sed -e 's/^#.*$//g' -e '/^$/d' /var/cache/build/packages-debian-cleanup.txt | \
```
tjanez: Hmm... this means we have to list build requirements twice, once in `packages-debian.txt` and then in `packages-debian-cleanup.txt`?

What about having two files, `packages-debian.txt` and `packages-debian-buildreq.txt`, with the first one only containing the packages that we want to have installed and the second one containing all extra build requirements that are then purged at the end?

resolwe_bio/docker_images/base/Dockerfile.ubuntu-16.04
```diff
@@ -0,0 +1,77 @@
+FROM docker.io/resolwe/base:ubuntu-16.04
```
tjanez: Obviously, same comments apply as for the Ubuntu 14.04-based image.

resolwe_bio/docker_images/base/common.sh
```diff
@@ -0,0 +1,85 @@
+#
+# Download package and verify SHA256 digest.
+#
+download_and_verify() {
+    local vendor=$1
+    local name=$2
+    local version=$3
+    local sha256sum=$4
+    local url=$(eval "echo \"$5\"")
```
tjanez: Add a comment why `$(eval "echo \"$<argument-number>\"")` is needed.

resolwe_bio/docker_images/base/common.sh
```diff
@@ -0,0 +1,85 @@
...
+    # Unpack.
+    echo "Unpacking..."
+    case "$extension" in
+        *.tar*|tar*|*.tgz|tgz)
+            tar -xf "${name}.${extension}" --directory ${name}
+            ;;
+        *.zip|zip)
+            unzip -q "${name}.${extension}" -d ${name}
+            ;;
+        *.elf|elf)
```
tjanez: What are `.elf` files?

resolwe_bio/docker_images/base/common.sh
```diff
@@ -0,0 +1,85 @@
...
+            ;;
+        *)
+            >&2 echo "ERROR: Unknown archive format '${extension}'."
+            exit 1
+    esac
+
+    rm -f "${name}.${extension}"
+    cd "${name}"
+
+    if [ -n "${subdir}" ]; then
```
tjanez: Since this is run with Bash, you can use `[[` and omit `"`s.

resolwe_bio/docker_images/legacy/packages-debian.txt
```diff
@@ -0,0 +1,56 @@
+bedtools
+samtools
+picard-tools
+bowtie
+bowtie2
+bwa
+tophat
+abyss
+# required by abyss scripts
```
tjanez: How many of the packages below are required by `abyss`?

Please, use new lines to structure the contents of the file. Also add comments on why something is needed where is appropriate.

resolwe_bio/docker_images/legacy/packages-debian.txt
```diff
@@ -0,0 +1,56 @@
...
+libxml2-dev
+libxslt1-dev
+# required for compiling R package arrayQualityMetrics
+libxt-dev
+libcairo2-dev
+xml2
+libcurl3
+libcurl4-openssl-dev
+trimmomatic
+texlive-latex-base
```
tjanez: Auch, do we do really need LaTeX?

resolwe_bio/docker_images/legacy/packages-manual/primerclip.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash -e
+
+. /var/cache/build/packages-manual/common.sh
+
+download_and_verify \
+    genialis \
+    primerclip \
+    2 \
+    5b74d96bab9bff0b7880ca134c4d67c67230e7d36f161269282d5994254da3c4 \
+    'https://data.genialis.com/repo/3f3f14d1-5100-46d5-af7b-abca7ff4a48f/292c647949ed795d6c73d8088cb839586e2287c6/?file_name=primerclip-v${version}&op=download&t=ca7225aa30&p=/tools/primerclip-v${version}#.elf'
```
tjanez: Replace the links to data.genialis.com with links to data2.genialis.com. Ask @abukosek where are the files on the new server and what are the links to them.

resolwe_bio/docker_images/legacy/packages-manual/snpeff.sh
```diff
@@ -0,0 +1,25 @@
...
+    4_3k \
+    396fadbefb4994838713c9a3e96393d16e7ec833254a962834e7cbd699b9f35e \
+    https://sourceforge.net/projects/snpeff/files/snpEff_v\${version}_core.zip \
+    snpEff
+
+rm -rf clinEff examples
+java -jar snpEff.jar download GRCh37.75 -v
+
+mkdir bin
+echo 'exec java -Xmx16g -jar /opt/snpeff/snpeff/snpEff.jar "$@"' > bin/snpEff
```
tjanez: Would these files also need a shebang?

resolwe_bio/docker_images/legacy/packages-manual/r-packages.sh
```diff
@@ -0,0 +1,68 @@
...
+# XXX: Use notary (https://github.com/ropenscilabs/notary) when ready.
+sed -e 's/^#.*$//g' -e '/^$/d' /var/cache/build/packages-r.txt | \
+    Rscript --slave --no-save --no-restore-history \
+        -e "install.packages(readLines('stdin'), repos='https://cloud.r-project.org')"
+
+# XXX: This is unverifiable and in CLEAR TEXT and thus may compromise the whole image.
+# XXX: Use notary (https://github.com/ropenscilabs/notary) when ready.
+Rscript --slave --no-save --no-restore-history -e " \
+  package_list = c( \
+    'arrayQualityMetrics', \
```
tjanez: Should we make a list of R packages installed from bioconductor similar to `packages-r.txt`?

resolwe_bio/processes/WGBS/bsmap.yml
```diff
@@ -7,6 +7,9 @@
   name: BSMAP
   requirements:
     expression-engine: jinja
+    executor:
+      docker:
+        image: "resolwebio/legacy:latest"
```
tjanez: I think there is no need to quote the name of the image (applies everywhere).

resolwe_bio/utils/test.py
```diff
@@ -97,6 +97,16 @@ def prepare_expression(self, f_rc='exp_1_rc.tab.gz',
         inputs = {'rc': f_rc, 'exp': f_exp, 'exp_type': f_type, 'exp_name': name, 'source': source}
         return self.run_process('upload-expression', inputs)
 
+    def assertAlmostEqualList(self, actual, expected):
```
tjanez: I think this would be also useful in Resolwe. What do you think?

resolwe_bio/docker_images/base/Dockerfile.ubuntu-14.04
```diff
@@ -0,0 +1,65 @@
+FROM docker.io/resolwe/base:ubuntu-14.04
+
+MAINTAINER Genialis <dev-team@genialis.com>
+
+# Add common build helpers.
+ADD common.sh /var/cache/build/packages-manual/common.sh
```
kostko: Because this is something only needed during build and could be removed later (although we currently don't remove it). Should we put this somewhere else? I can create something like `/build`.

resolwe_bio/docker_images/base/Dockerfile.ubuntu-14.04
kostko: Yes. The problem is with specific packages, which require numpy during `setup.py` evaluation as these packages will crash while pip evaluates them.

resolwe_bio/docker_images/base/Dockerfile.ubuntu-14.04
kostko: Good idea, I'll change this.

resolwe_bio/docker_images/base/common.sh
kostko: Raw binaries, I needed to specify some extension, so I added `.elf`.

resolwe_bio/docker_images/legacy/packages-debian.txt
kostko: Not sure, I just copied this over from the old image? We probably need it for generating some reports.

resolwe_bio/docker_images/base/Dockerfile.ubuntu-14.04
kostko: There is no long option for `-I` (there is `--replace`, but the man page says it's deprecated). I'll just change `-n` to `--max-args`.

kostko:
All tests fixed. @jkokosar mentioned that we should also run some skipped tests.

kostko:
In order to correctly test this PR, after merging the Resolwe PR, we need to add a Docker Hub repository for the new images before merging this PR. They need to be (temporarily) set up to build from this branch instead of master. cc @tjanez

resolwe_bio/docker_images/base/Dockerfile.ubuntu-14.04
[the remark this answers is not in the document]
tjanez: Ok, I see.

If you remove it at the end, then I'm fine with `/build`. Otherwise, I would prefer something like `/opt/genialis/docker-build-helpers`.

resolwe_bio/docker_images/base/Dockerfile.ubuntu-14.04
```diff
@@ -0,0 +1,65 @@
...
+    apt-get --no-install-recommends -y install wget && \
+    apt-get autoremove --purge -y && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+# Package manifests and custom build scripts.
+ONBUILD ADD packages-*.txt /var/cache/build/
+ONBUILD ADD packages-manual /var/cache/build/packages-manual/
+
+ONBUILD RUN export DEBIAN_FRONTEND=noninteractive && \
```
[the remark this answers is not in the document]
tjanez: Ok, agreed.

resolwe_bio/docker_images/base/Dockerfile.ubuntu-14.04
tjanez: Auch... another of many pip's shortcomings...

resolwe_bio/docker_images/base/Dockerfile.ubuntu-14.04
tjanez: 👍

resolwe_bio/docker_images/base/common.sh
tjanez: Ok, thanks for the explanation.

resolwe_bio/docker_images/legacy/packages-debian.txt
tjanez: Ok, leave it in then. We should be careful which (sub)parts of TeX Live we add when we create *new* images.

tjanez:
> In order to correctly test this PR, after merging the Resolwe PR, we need to add a Docker Hub repository for the new images before merging this PR. They need to be (temporarily) set up to build from this branch instead of master. cc @tjanez

I'll set up automated builds and we can push manually built images the first time.

tjanez:
@kostko, please also add a README to `resolwe_bio/docker_images/base/`, similar to [README for Resolwe toolkit's base Docker image](https://github.com/genialis/resolwe/blob/f14f3453f60766969fb8b408e4af92961f5138ae/resolwe/toolkit/docker_images/base/README.md).

tjanez:
> @kostko, please also add a README to `resolwe_bio/docker_images/base/`, similar to [README for Resolwe toolkit's base Docker image](https://github.com/genialis/resolwe/blob/f14f3453f60766969fb8b408e4af92961f5138ae/resolwe/toolkit/docker_images/base/README.md).

Please, do a similar thing for the `legacy` image and with a bold note that this image is deprecated and should not be used for new processes.

resolwe_bio/utils/test.py
kostko: Moved to genialis/resolwe#350.

kostko:
@tjanez I also need permission to push to `genialis/iclip` as this is needed to complete tests.

tjanez:
> @tjanez I also need permission to push to `genialis/iclip` as this is needed to complete tests.

I was curious and I checked out which processes actually need `genialis/iclip` Docker image.
If I'm not mistaken, only one process, [`upload-fasta-nucl`](https://github.com/genialis/resolwe-bio/blob/f5dec1c5c93edf48db61e4b1e0bee3ebf8239ddf/resolwe_bio/processes/import_data/nucleotide_seq.yml#L6-L55) needs this image ([in current `master`, it requires the previous image, namely `resolwe/bio-linux8-resolwe-preview:icount`](https://github.com/genialis/resolwe-bio/blob/f5dec1c5c93edf48db61e4b1e0bee3ebf8239ddf/resolwe_bio/processes/import_data/nucleotide_seq.yml#L14)).

Looking at process' `program`, it seems it only needs `samtools` and a few basic utilities to run. Would it make sense to just use the `resolwebio/legacy` image for it?
It seems odds to require some "third-party" Docker image for Resolwe Bioinformatics processes (and tests) to run.

/cc @jkokosar, @mstajdohar

kostko:
Good point, it is probably not needed. I'll change it.

jkokosar:
Legacy image probably still uses an old version of samtools (0.19...) that
will produce a sequence index incompatible with iclip processes

tjanez:
> Legacy image probably still uses an old version of samtools (0.19...) that will produce a sequence index incompatible with iclip processes

You probably meant that the version of samtools is 0.1.19. @jkokosar, should we bump samtools to 1.3.1 directly in the legacy image?
For me, it is still odd to require a "third-party" Docker image, `genialis/iclip`, for a newer version of samtools.

I see that [Ubuntu 17.04 comes with samtools 1.3.1 packaged in the official repository](https://packages.ubuntu.com/zesty/samtools).

To sum up, we have 2 options:
- We make a Ubuntu 17.04 base image and install samtools from there.
- We install samtools 1.3.1 as a manual package.

kostko:
I am against using 17.04. If we need to bump, we will install it manually as it is very simple to do (and I think we already have a script for that in the new iclip docker file in the PR).

jkokosar:
We didn't bump the samtools version before, because some other processes
depend on older version of samtools. Now that the chipseq processes use
their own image, I'm not sure if this is still true.

kostko:
I'll try bumping the version and running all tests.

resolwe_bio/utils/test.py
tjanez: Perfect, thanks!

kostko:
Version of samtools and bcftools bumped, which required another round of process and test fixes. I've pushed the new images and also bumped Resolwe Bio to 3.0.0a1 due to breaking changes.

kostko:
The equality of old and new `.bai` binary file used in tests has been verified using a [deserializer from this repository](https://github.com/LabAdvComp/python-bam).

tjanez:
I've taken another look and it looks very good!

One thing that would make the git history a little nicer would be to slightly modify and reorder the commits:
1. Add resolwebio/{base,legacy} docker images (leave as is)
2. Fix processes to use the new Docker images (leave as is)
3. Bump version to 3.0.0a1 due to breaking changes:

    - edit setup.py to require Resolwe 3.0.0a1 and add the appropriate CHANGELOG entry
    - Change the commit message to something like: Bump version to 3.0.0a1 due to requiring Resolwe 3.0.x
4. Remove clustering-hierarchical-genes-etc process (modify CHANGELOG entry to include BACKWARD INCOMPATIBLE notice)
5. Fix tests (without setup.py changes)

kostko:
For some reason the following test fails on Jenkins with a different output, but works for me locally (with the same Docker image). Can someone check locally if it fails for them too?

Failing test:
* `resolwe_bio.tests.processes.test_diff_expression.DiffExpProcessorTestCase.test_deseq2`

tjanez:
> For some reason the following test fails on Jenkins with a different output, but works for me locally (with the same Docker image, ID `5a8c22428a93`). Can someone check locally if it fails for them too?

> Failing test:
* `resolwe_bio.tests.processes.test_diff_expression.DiffExpProcessorTestCase.test_deseq2`

> It could also be something related to parallel execution, but I don't see how exactly as the tests should be isolated (both database and data directories).

@kostko, I've reproduced the problem on my workstation. There was no need to run the tests in parallel, it even appeared when running just this test.

When [`self.assertFile(diff_exp, 'raw', 'diffexp_deseq2.tab.gz', compression='gzip')`](https://github.com/genialis/resolwe-bio/blob/master/resolwe_bio/tests/processes/test_diff_expression.py#L63) fails, it fails due to exactly the same hash mismatch as [on Jenkins](https://ci.genialis.com/job/genialis-github/job/resolwe-bio/view/change-requests/job/PR-335/36/console):

```
======================================================================
FAIL: test_deseq2 (resolwe_bio.tests.processes.test_diff_expression.DiffExpProcessorTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/tadej/Genialis/resolwe-bio/resolwe_bio/tests/processes/test_diff_expression.py", line 64, in test_deseq2
    self.assertFile(diff_exp, 'raw', 'diffexp_deseq2.tab.gz', compression='gzip')
  File "/home/tadej/.virtualenvs/resolwe-bio/lib/python3.5/site-packages/resolwe/test/testcases/process.py", line 487, in assertFile
    self._assert_file(obj, field['file'], fn, **kwargs)
  File "/home/tadej/.virtualenvs/resolwe-bio/lib/python3.5/site-packages/resolwe/test/testcases/process.py", line 456, in _assert_file
    correct_hash, output_hash) + self._debug_info(obj))
AssertionError: '3d473a985cc0ad1c5b474e75d028507bd3c02009696c4afbe3b4763acbd1b311' != '5603096843dd89f1408a62012b33591f7f97280cc52450084f71ef45689b8d88'
- 3d473a985cc0ad1c5b474e75d028507bd3c02009696c4afbe3b4763acbd1b311
+ 5603096843dd89f1408a62012b33591f7f97280cc52450084f71ef45689b8d88
 : File contents hash mismatch: 3d473a985cc0ad1c5b474e75d028507bd3c02009696c4afbe3b4763acbd1b311 != 5603096843dd89f1408a62012b33591f7f97280cc52450084f71ef45689b8d88
```

I've used the same image as you did:
```bash
$ docker images | grep resolwebio/legacy
docker.io/resolwebio/legacy                    latest                5a8c22428a93        2 days ago          3.89 GB
docker.io/resolwebio/legacy                    <none>                37566a789704        6 days ago          3.88 GB
```

Digging deeper into this, it appears that there is a problem with rounding errors for some floating number in the `diffexp_deseq2.tab.gz` file:

```bash
$ zdiff diffexp_deseq2-*.gz
3c3
< DPU_G0067102	1794.00090077659	-0.49179888884998	0.472973484123619	-1.03980224126357	0.298431787634463	0.306151892124043
---
> DPU_G0067102	1794.00090077659	-0.491798888849984	0.472973484123619	-1.03980224126358	0.298431787634459	0.306151892124043
```

kostko:
Tests on Jenkins seem really unstable, I am constantly getting:
```

PR #336, opened 2017-08-17
Use build-essential package in chipseq DockerFile

jkokosar:


resolwe_bio/docker_images/chipseq/Dockerfile
```diff
@@ -75,6 +76,7 @@ RUN apt-get update && \
 
     # Remove unnecessary packages after bamliquidator compilation
     apt-get purge -y \
```
mstajdohar: @kostko @abukosek Will this remove the dependancies?

kostko: No, you need to also call `autoremove --purge`.

mstajdohar: @jkokosar Also rebase to master while you're at it, please.

PR #337, opened 2017-08-17
LaTeX docker image

JureZmrzlikar:


resolwe_bio/docker_images/latex/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM resolwe/base:ubuntu-16.04
+
+RUN apt-get update && apt-get upgrade -y && apt-get install -y \
+    wget \
+    texlive-latex-base \
+    texlive-latex-recommended \
+    texlive-latex-extra
+
+# Cleanup.
+RUN apt-get clean && rm -rf /var/lib/apt/lists/*
```
dblenkus: Combine two `RUN` commands.

Each command creates a layer of docker image. So if you clean the image in separate command, all unnecessary things will be presented in the intermediate layer and you won't save anything on the space.

resolwe_bio/docker_images/latex/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM resolwe/base:ubuntu-16.04
+
+RUN apt-get update && apt-get upgrade -y && apt-get install -y \
```
kostko: Any reason for running `upgrade` here? I think all upgrades should only be handled in base images.

JureZmrzlikar: Agh, no reason at all. Thanks for noticing.

mstajdohar:
Deprecated by #338.

PR #340, opened 2017-08-18
Fix the cleanup step in chipseq Docker image

mstajdohar:


resolwe_bio/docker_images/chipseq/Dockerfile
```diff
@@ -84,9 +84,11 @@ RUN apt-get update && \
       libboost-timer-dev \
       libgoogle-perftools-dev \
       libtbb-dev && \
-
-    # Cleanup
+    apt-get autoremove --purge -y
     apt-get clean && \
-    rm -rf /var/lib/apt/lists/*
+    rm -rf /var/lib/apt/lists/* && \
+    rm -rf /var/lib/apt/lists/* && \
```
kostko: Remove one of the lines above.

mstajdohar: Fixed, thanks.

kostko:
Why are there some unrelated commits now? They weren't here before.

mstajdohar:
We need to accept six PRs by Monday, preferably today. All have been reviewed, accepted and tested before (except for this one that has not been tested yet but does not influence any tests). Running tests again separately on each would be a waste of time and resources (and not feasible in one day). As a workaround I have rebased the six PRs in a chain.

tjanez:
> Why are there some unrelated commits now? They weren't here before.

> We need to accept six PRs by Monday, preferably today. All have been reviewed, accepted and tested before (except for this one that has not been tested yet but does not influence any tests). Running tests again separately on each would be a waste of time and resources (and not feasible in one day). As a workaround I have rebased the six PRs in a chain and canceled tests of the intermediated PRs.

Ohh, this is really ugly... @mstajdohar, we need to dedicate some resources to fix the tests taking too much time. Otherwise, we'll always have to make ugly work-around like this one...

mstajdohar:
The tests for resolwe-bio are really huge pain.

PR #341, opened 2017-08-21
Add custom Cutadapt process

jvrakor:
@mzganec @jkokosar @mstajdohar


resolwe_bio/tests/processes/test_reads_filtering.py
```diff
@@ -232,6 +232,32 @@ def test_cutadapt_paired(self):
         self.assertFiles(cutadapt_paired, 'fastq2', ['cutadapt_paired_reverse_trimmed.fastq.gz'],
                          compression='gzip')
 
+    def test_cutadapt_diagenode_single(self):
+        self.keep_data()
```
mzganec: Throw out `self.keep_data()` (in two places). You can also merge the tests for single- and paired-end reads into a single test to save some time during tests.

PR #343, opened 2017-08-21
Add custom Cutadapt process

jvrakor:
@mzganec @jkokosar @mstajdohar

PR #346, opened 2017-08-23
Remove keeping data from etc process test

jvrakor:
@jkokosar @mstajdohar


jvrakor:
Moved to https://github.com/genialis/resolwe-bio/pull/343

PR #348, opened 2017-08-23
Add RNA-seq dockerfile

jvrakor:

### DIFF
--- a/resolwe_bio/docker_images/rnaseq/Dockerfile
+++ b/resolwe_bio/docker_images/rnaseq/Dockerfile
@@ -1,0 +1,98 @@
+FROM docker.io/resolwe/base:ubuntu-16.04
+
+LABEL MAINTAINER Resolwe Bioinformatics authors https://github.com/genialis/resolwe-bio
+
+RUN apt-get update && \
+
+    apt-get -y install --no-install-recommends \
+      python \
+      python-pip \
+      r-base-core \
+      samtools \
+      unzip \
+      wget \
+      # Required for fastqc.
+      default-jre \
+      perl \
+      #Required for R packages.
+      build-essential \
+      libz-dev \
+      libcurl4-openssl-dev \
+      libxml2-dev \
+      && \
+    
+    # Required for HTSeq.
+    echo "Installing numpy..." && \
+    pip install numpy==1.13.1 && \
+
+    # Required for HTSeq.
+    echo "Installing pysam..." && \
+    pip install pysam==0.11.2.2 && \
+
+    echo "Installing HTSeq..." && \
+    pip install HTSeq==0.9.1 && \
+
+    echo "Installing cutadapt..." && \
+    pip install cutadapt==1.14 && \
+
+    echo "Installing FastQC..." && \
+    FastQC_VERSION=0.11.5 && \
+    FastQC_SHA256SUM=dd7a5ad80ceed2588cf6d6ffe35e0f161c0d9977ed08355f5e4d9473282cbd66 && \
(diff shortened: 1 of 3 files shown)


resolwe_bio/docker_images/rnaseq/Dockerfile
```diff
@@ -0,0 +1,83 @@
+FROM docker.io/resolwe/base:ubuntu-16.04
+
+LABEL MAINTAINER Resolwe Bioinformatics authors https://github.com/genialis/resolwe-bio
+
+RUN apt-get update && \
+    apt-get -y install --no-install-recommends software-properties-common && \
```
jkokosar: Check if you need 'software-properties-common'. I think we needed this to build / install bamliquidator in chipseq docker image. It might not be needed here

resolwe_bio/docker_images/rnaseq/Dockerfile
```diff
@@ -0,0 +1,83 @@
...
+    echo "$express_SHA256SUM */tmp/express.tgz" | sha256sum -c - && \
+    mkdir -p /opt/pachterlab/express-$express_VERSION && \
+    tar -xf /tmp/express.tgz --directory /opt/pachterlab/express-$express_VERSION --strip-components=1 && \
+    rm /tmp/express.tgz && \
+    rm -r /opt/pachterlab/express-$express_VERSION/sample_data && \
+    echo "PATH=\$PATH:/opt/pachterlab/express-$express_VERSION" >> /etc/profile.d/rnaseq.sh && \
+
+    # Cleanup
+    apt-get purge -y \
+      libhdf5-serial-dev \
```
jkokosar: Those packages were needed to compile the bamliquidator tool in chipseq docker image and were removed since they are not needed when running the tool. Those tools were not installed in this image so there is no need to remove them here.

jkokosar:
Rebase and add changelog. We should set the processes to use this image in a separate PR.

resolwe_bio/docker_images/rnaseq/Dockerfile
kostko: Indeed, this is only needed to add packages from PPAs. I don't see any PPAs being used here, so I think it can be removed.

resolwe_bio/docker_images/rnaseq/Dockerfile
```diff
@@ -0,0 +1,83 @@
...
+      libboost-timer-dev \
+      libgoogle-perftools-dev \
+      libtbb-dev && \
+    apt-get autoremove --purge -y && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/* && \
+    rm -rf /root/.cache/pip && \
+    rm -rf /tmp/*
+
+CMD ["/bin/bash"]
```
kostko: This can be removed.

resolwe_bio/docker_images/rnaseq/Dockerfile
```diff
@@ -0,0 +1,83 @@
...
+      unzip \
+      wget \
+      # Required for fastqc.
+      default-jre \
+      perl \
+      && \
+    
+    # Required for HTSeq.
+    echo "Installing numpy..." && \
+    pip install numpy==1.13.1 && \
```
jkokosar: Why not install python packages using pip3? Please check if it works as desired.

jkokosar:
To use this tool with the _htseq-count_ process, we would also need to add R in the image, since the process calls _expression_fpkm_tpm.R_ script.

resolwe_bio/docker_images/rnaseq/Dockerfile
jvrakor: All of them or just 'libhdf5-serial-dev'?

jkokosar: all of them

PR #352, opened 2017-08-29
Fix cuffnorm process group analysis

jvrakor:
@mstajdohar @jkokosar @dblenkus


resolwe_bio/tools/cuffnorm_output_table_headers.py
```diff
@@ -18,7 +18,7 @@
 samples = []
 # This sort must produce the same order as cufflinks_sample_groups.py
 for x in sorted(zip(replicates, sample_names), key=itemgetter(0)):
-    samples.append(x[1])
+    samples.append('_Group'.join([x[1], x[0]]))
```
dblenkus: '{}_Group{}'.format(x[1], x[0])

resolwe_bio/processes/expression/cuffnorm.yml
```diff
@@ -7,8 +7,10 @@
   name: Cuffnorm
   requirements:
     expression-engine: jinja
+    resources:
+      cores: 1    
   data_name: 'Cuffnorm ({{cuffquant|map("sample_name")|join(", ")|default("?")}})'
-  version: 1.1.1
+  version: 1.1.2
```
dblenkus: Change the version to `1.2.0` as output is not the same as in the previous version.

jvrakor: The version was changed due to core requirement change but I was hesitant of changing the version to 1.2.0 later since the process itself didn't change, only a tool.

jvrakor: @mstajdohar @kostko 
Should the version be changed if the .yaml file was not changed at all but only a tool it uses? If yes, that would mean that use of a newer version of a package in docker image would result in having to bump all processes (that use that package) versions.

mstajdohar: This is not required, but it is considered a good practice to also bump the process version.

mstajdohar:
@dblenkus Please do a second review. Your requested changes have been addressed.

PR #353, opened 2017-09-05
Escape user input for amplicon report latex

jkokosar:


resolwe_bio/tools/make_report.py
```diff
@@ -21,10 +21,18 @@
 parser.add_argument('--logo', help="Logo.")
 
 
-def _format_latex(string):
+def _escape_latex(string):
     """Format normal string to be LaTeX compliant."""
-    string = string.replace('_', '\_')
+    string = string.replace('\\', '\\\\')
```
JureZmrzlikar: Too much repetition. You could make a list of special latex charachters and iterate through it - to avoid repeating `string.raplace` this many times. something like:
`special_chars = ['\\', '&', '%', '$', '#', '_', '{', '}']`
`for special_char in special_chars:`
    `    string = string.replace(special_char, '\' + special_char)`
In this case, last two charachters still need to be done by hand.
Even better, you could define a constant, somthing like `LATEX_SPECIAL_CHARS`. It would be a dictionary, mapping special chars to their replacements.

mstajdohar: Good suggestion, but I suggest we accept the current implementation (given our time constraints) to avoid re-testing the PR.

JureZmrzlikar: Sure, agree.

PR #355, opened 2017-09-06
Fix validation error in Test sleep progress process

dblenkus:


mstajdohar:
We could move this process to resolwe.

PR #357, opened 2017-09-08
Add mm10_dm6 and hg19_dm6 organisms to descriptor schema

jvrakor:
@mstajdohar @jkokosar @rafecooks


rafecooks:
Discussed with Charles.

He requests the following:
Organism = primary (e.g. `mouse`, `human`)
Build = `hg19_dm6` or `mm10_dm6`

This would be similar to how we handle the fields for EERC spike-ins, where the organism is human or mouse and the build is "hg19_eerc" or similar.

Further, in terms of the processing and data management, he requests that the split BAMS both attach to the same sample, but one of them is primary and labeled BAM, the other is secondary and can be called some other descriptor to avoid confusing the system.

PR #363, opened 2017-09-18
Fix Chemut workflow and enable tests

jkokosar:


resolwe_bio/processes/variant_calling/vc_chemut.yml
```diff
@@ -280,7 +280,7 @@
         -glm {{ Varc_param.glm }} \
         -stand_call_conf {{ Varc_param.stand_call_conf }} \
         -stand_emit_conf {{ Varc_param.stand_emit_conf }} \
-        {% if Varc_param.intervals %}{% for l in Varc_param.intervals %} {{"-L" if l != ""}} {{l}} {% endfor %}{% endif %} \
+        {% if Varc_param.intervals %}{% for l in Varc_param.intervals %} '-L' {{l}} {% endfor %}{% endif %} \
```
kostko: Why the quotes around `-L`?

jkokosar: Not needed, thanks!

PR #364, opened 2017-09-18
Fix cuffnorm to correctly use sample names as labels in outputs

jvrakor:
@jkokosar @mstajdohar @dblenkus @kostko


resolwe_bio/processes/expression/cuffnorm.yml
```diff
@@ -196,8 +197,14 @@
       re-save-file boxplot cuffnorm_exprs_boxplot.pdf
       re-save-file fpkm_exp_raw cuffnorm_all_fpkm_exprs_raw.txt
       re-save-file replicate_correlations cuffnorm_replicate_correlations.pdf
-      re-save-file fpkm_means cuffnorm_all_fpkm_means.txt
-      re-save-file exp_fpkm_means cuffnorm_exprs_fpkm_means.txt
+      if [ -e cuffnorm_all_fpkm_means.txt ]
```
kostko: You can consolidate this into one line:
```bash
if [ -e cuffnorm_all_fpkm_means.txt ]; then
```

Also below.

resolwe_bio/processes/expression/cuffnorm.yml
```diff
@@ -147,13 +149,12 @@
...
 
       # Run normalizeRNASeq.
-      # TODO: Check what should be used for 'groupString' argument.
       Rscript --slave --no-save --no-restore-history -e " \
         library(RNASeqT); \
         normalizeRNASeq( \
           geneFPKMFile='genes.fpkm_table', \
           name='cuffnorm', \
-          groupString='${sample_names}', \
+          groupString='${labels}', \
```
kostko: Has something changed and this will now work? Please elaborate.

kostko: Is this something related to 7435626f3867d809d1e1f41a7399a5385c882ea3?

kostko: If so, why didn't the tests fail? Please make sure that this is covered in tests.

jvrakor: Yes, this was fixed in https://github.com/genialis/resolwe-bio/pull/352

kostko: So it means that it is not covered in tests? Because otherwise, the tests should fail with the wrong behavior?

jvrakor: The tool was changed but none of the Rscript outputs is covered in the tests.

kostko: Does it make sense to include it in tests in some way?

jvrakor: Maybe. I will expand the cuffnorm tests so they cover more

docs/CHANGELOG.rst
```diff
@@ -10,6 +10,10 @@ This project adheres to `Semantic Versioning <http://semver.org/>`_.
 Unreleased
 ==========
 
+Fix
+---
+* Make not always present ``cuffnorm`` outputs not required and fix
+  what was done in the previous ``cuffnorm`` commit
```
jkokosar: Maybe rephrase this to "...and revert previous ``cuffnorm`` commit (commit hash)"

mstajdohar:
Please rebase

jvrakor:
This needs to be reviewed again because it was changed a lot since @kostko reviewed it.

mstajdohar:
@kostko Would you do another pass?

mstajdohar:
Address @JenkoB's comment and rebase, please.

resolwe_bio/tests/processes/test_expression.py
```diff
@@ -66,13 +66,29 @@ def test_cuffnorm(self):
...
+            sample_4 = self.run_process("upload-cxb", inputs)
+
+            inputs = {'src': 'cuffquant_5.cxb', 'source': "UCSC"}
+            sample_5 = self.run_process("upload-cxb", inputs)
+
+            inputs = {'src': 'cuffquant_6.cxb', 'source': "UCSC"}
+            sample_6 = self.run_process("upload-cxb", inputs)
+
+            inputs = {'src': 'cuffquant_7.cxb', 'source': "UCSC"}
+            sample_7 = self.run_process("upload-cxb", inputs)
```
dblenkus: What is the benefit of using so many samples? Keep in mind that tests should be small and fast.

resolwe_bio/processes/expression/cuffnorm.yml
```diff
@@ -144,16 +144,15 @@
       re-progress 0.8
 
       # Replace all cuffnorm output tables headers with sample names.
-      cuffnorm_output_table_headers.py ${replicates} ${sample_names}
+      cuffnorm_output_table_headers.py ${replicates} ${sample_names} ${experiment}
```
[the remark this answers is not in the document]
jvrakor: File extensions are not usually part of the sample name but if a user wishes to keep them or have dots in a sample name we should be able to allow that. @mstajdohar ?

resolwe_bio/tests/processes/test_expression.py
jvrakor: First we wish to have at least 3 groups as before we had only 2 groups and the tests did not catch some things that way. We also wanted to have at least 2 samples in each of these groups so we can see if the sorting works as it should. This is also the reason why we have 3 samples in one of the groups. We may be able to drop one sample from the test if that is preferred.

PR #365, opened 2017-09-18
Add bam-split process with supporting processes

jvrakor:


resolwe_bio/processes/support_processors/bam_split.yml
```diff
@@ -0,0 +1,88 @@
...
+      BAM={{bam.bam.file}}
+      NAME=`basename {{ bam.bam.file }} .bam`
+      ORG1={{organism}}
+      ORG2={{organism2}}
+
+      samtools view -h -o "${NAME}.sam" "${BAM}"
+      re-checkrc
+      re-progress 0.1
+
+      samtools view -S -F 4 "${NAME}.sam" | awk -vOFS="	"  '{sindex=index($3,"_'${ORG2}'"); if(sindex >0) {$3=substr($3,1,sindex-1);  print $0 > "'${NAME}'_'${ORG2}'.sam";} else { print $0 > "'${NAME}'_'${ORG1}'.sam";}}'
```
kostko: This is ugly, please break command and awk script into multiple lines.

resolwe_bio/processes/support_processors/bam_split.yml
```diff
@@ -0,0 +1,88 @@
...
+
+      samtools sort "${NAME}_${ORG2}.bam" "${NAME}_${ORG2}.sorted"
+      re-checkrc
+      re-progress 0.6
+      
+      mv "${NAME}_${ORG1}.sorted.bam" "${NAME}_${ORG1}.bam" 
+      mv "${NAME}_${ORG2}.sorted.bam" "${NAME}_${ORG2}.bam"
+
+      re-export "${NAME}_${ORG1}.bam"
+      echo 'run {"process":"upload-bam-primary","input":{"bam":'{{bam | id}}',"src":"'"${NAME}_${ORG1}.bam"'"}}'
```
kostko: @dblenkus @tjanez We should really introduce (universal, e.g. not tied to specific Docker image) helpers for these operations. Probably when we refactor the process<->executor communication protocol to use named pipes instead of stdout.

resolwe_bio/processes/import_data/sam_header.yml
```diff
@@ -0,0 +1,39 @@
...
+# ===================
+---
+
+- slug: upload-header-sam
+  name: Upload Header SAM file
+  requirements:
+    expression-engine: jinja
+    executor:
+      docker:
+        image: resolwebio/legacy:latest
```
kostko: Any reason why this (very simple) process requires the (relatively big) legacy image, while the above use rnaseq? This could probably even work with base directly as it doesn't use any tools.

jvrakor: No specific reason. All the import processes seem to use this image however. I will remove docker requirements in this process and we might want to do the same for (some) other.

kostko: Removing requirements will break this process, because the default is the old biolinux-based image (which is no longer supported). You should use explicit requirements and use `resolwebio/base:latest` instead.

kostko: Sorry, those don't have `latest`, you should use the `ubuntu-16.04` tag.

resolwe_bio/processes/import_data/bam.yml
```diff
@@ -116,3 +116,103 @@
...
+  persistence: RAW
+  description: >
+    Upload a primary mapping file in BAM format.
+  input:
+    - name: bam
+      label: Hybrid bam
+      type: data:alignment:bam
+      required: false
+      description: >
+        Primary bam will be uploaded to the same collection and sample where hybrid bam is.
```
jkokosar: It will be appended to the same sample. The collection inclusion is just a consequence of the sample already being (or not) included in the collection(s).

resolwe_bio/processes/support_processors/bam_split.yml
```diff
@@ -0,0 +1,99 @@
...
+
+      cat {{header2.sam.file}} "${NAME}_${ORG2}.sam" | samtools view -bS - > "${NAME}_${ORG2}.bam"
+      re-checkrc
+      re-progress 0.4
+
+      samtools sort "${NAME}_${ORG1}.bam" "${NAME}_${ORG1}.sorted"
+      re-checkrc
+      re-progress 0.5
+
+      samtools sort "${NAME}_${ORG2}.bam" "${NAME}_${ORG2}.sorted"
```
jkokosar: Does samtools sort allow for inplace sorting? So you could avoid `mv` bellow?

resolwe_bio/tests/processes/test_support_processors.py
```diff
@@ -17,6 +17,30 @@ def test_reference_compatibility(self):
         compatibility_test = self.run_process('reference_compatibility', inputs)
         self.assertFile(compatibility_test, 'report_file', 'sp_test_compatibility_report.txt')
 
+    def test_bam_split(self):
+        self.keep_data()
```
jkokosar: remove this

PR #367, opened 2017-09-20
Add resolwebio/base image based on Ubuntu 17.04

kostko:


docs/CHANGELOG.rst
```diff
@@ -12,7 +12,8 @@ Unreleased
 
 Added
 -----
-* Add Mus musculus to all BCM workflows' schemas 
+* Add Mus musculus to all BCM workflows' schemas
```
tjanez: Hmm... why does GitHub show this line as changed?

kostko: Because I have my editor configured to remove all extra spaces at the end and the person who previously updated the changelog doesn't ;-) There is an extra space that got removed.

PR #369, opened 2017-09-21
Make dnaseq docker image

JureZmrzlikar:


JureZmrzlikar:
Rebased and added changelog entry.

JureZmrzlikar:
@tjanez , now I use `resolwebio/base:ubuntu-17.04` image like you suggested.

Issue #370, opened 2017-09-22
Remove obstolete Mongo escaping


kostko:
There is some left-over escaping for MongoDB syntax from the times that Resolwe used MongoDB. It should be removed as it may cause problems.

Function in question, which is used in some places and should be removed:
https://github.com/genialis/resolwe-bio/blob/9518106c09be420a78f7298b328071d3299e8751/resolwe_bio/tools/utils.py#L11-L13

mstajdohar:
@jkokosar We call `escape_mongokey` in tools' scripts.

mzganec:
Fixed in [https://github.com/genialis/resolwe-bio/pull/371](https://github.com/genialis/resolwe-bio/pull/371).

PR #371, opened 2017-09-22
Remove obsolete mongokey escape functionality

mzganec:


mzganec:
Without this fix, expressions containing periods (`.`) in gene names result in hieroglyphs in the JSON output field when uploaded using `upload-expression`.

kostko:
This seems related to #370. Can you remove all uses of this function and the function altogether?

mzganec:
The mongokey escape function and all its uses have been removed as per @kostko's suggestion.

PR #380, opened 2017-10-09
Speed-up insert management commands in kb

dblenkus:
**Important:** This requires PostgreSQL 9.5

Requires https://github.com/genialis/resolwe/pull/378

### DIFF
--- a/resolwe_bio/kb/management/commands/insert_features.py
+++ b/resolwe_bio/kb/management/commands/insert_features.py
@@ -7,15 +7,16 @@ Insert Knowledge Base Features
 """
 from __future__ import absolute_import, division, print_function, unicode_literals
 import csv
+import json
 import logging
-from tqdm import tqdm
 
+from django.core.exceptions import ValidationError
 from django.core.management.base import BaseCommand
+from django.db import connection
 
 from resolwe.elastic.builder import index_builder
 from resolwe.utils import BraceMessage as __
 
-from resolwe_bio.kb.elastic_indexes import FeatureSearchIndex
 from resolwe_bio.kb.models import Feature
 from .utils import decompress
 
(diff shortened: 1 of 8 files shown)


resolwe_bio/kb/management/commands/insert_features.py
```diff
@@ -87,53 +89,89 @@ def add_arguments(self, parser):
...
-
-        index_builder.push(index=FeatureSearchIndex)
-
-        count_total = count_inserted + count_updated + count_unchanged + count_failed
-        logger.info("Total features: %d. Inserted %d, updated %d, "  # pylint: disable=logging-not-lazy
-                    "unchanged %d, failed %d." %
-                    (count_total, count_inserted, count_updated, count_unchanged, count_failed))
+                # TODO: Raise error in features are not unique
+                if (row['Source'], row['ID']) not in unique_features:
+                    features.append([
```
kostko: Please add a comment that this is a list of lists and not list of dicts due to performance reasons. And also that any changes here may require the query to be updated.

resolwe_bio/kb/management/commands/insert_mappings.py
```diff
@@ -34,26 +36,57 @@ def add_arguments(self, parser):
...
+
+            mappings = []
+            unique_mappings = set()
+            for row in csv.DictReader(tab_file, delimiter=str('\t')):
+                if row['relation_type'] not in relation_type_choices:
+                    raise ValidationError(
+                        "Unknown relation type: {}".format(row['relation_type'])
+                    )
+
+                mapping = (
```
kostko: Add similar comment as mentioned above for features.

resolwe_bio/kb/management/commands/insert_mappings.py
```diff
@@ -34,26 +36,57 @@ def add_arguments(self, parser):
...
+            with connection.cursor() as cursor:
+                cursor.execute(
+                    """
+                    INSERT INTO {table_name} (
+                        relation_type, source_db, source_id, target_db, target_id
+                    )
+                    SELECT
+                        value->>0, value->>1, value->>2, value->>3, value->>4
+                    FROM json_array_elements(%s)
+                    ON CONFLICT DO NOTHING
```
kostko: No updates for mappings?

tests/docker-compose.yml
```diff
@@ -2,7 +2,7 @@
 # Development configuration of dependent services for Docker Compose.
 #
 postgresql:
-    image: postgres:9.4
+    image: postgres:9.6
```
kostko: Will also require a bump on Jenkins and in production. This is a breaking change, add it to changelog as such.

resolwe_bio/kb/management/commands/insert_mappings.py
dblenkus: We have `unique_together` constraint that includes all fields: https://github.com/genialis/resolwe-bio/blob/master/resolwe_bio/kb/models.py#L97
So conflict means that exact same mapping is already present and we don't need to update anything.

Maybe I can rewrite this to `ON CONFLICT (relation_type, source_db, source_id, target_db, target_id) DO NOTHING` to make it clearer, but I have to test how this affects performance.

kostko: I see, ok then just add a comment why this is so.

dblenkus:
@kostko I've addressed your comments. Now only the indexing part is missing, but it has to be supported in Resolwe first.

mstajdohar:
Rebase

mstajdohar:
@dblenkus, did you support indexing in Resolwe?

dblenkus:
This is waiting for @tjanez to support PostgreSQL 9.6, everything else is done.

PR #384, opened 2017-10-12
Run partial test suite for pull requests

kostko:
Requires genialis/resolwe#381.

### DIFF
--- a/tests/settings.py
+++ b/tests/settings.py
@@ -39,6 +39,7 @@ INSTALLED_APPS = (
     'resolwe.flow',
     'resolwe.elastic',
     'resolwe.toolkit',
+    'resolwe.test_helpers',
 
     'resolwe_bio',
     'resolwe_bio.kb',
@@ -158,3 +159,9 @@ if os.environ.get('RESOLWEBIO_TESTS_SKIP_DOCKER_FAILURES', '').lower() in ["no",
 
 ELASTICSEARCH_HOST = os.environ.get('RESOLWE_ES_HOST', 'localhost')
 ELASTICSEARCH_PORT = int(os.environ.get('RESOLWE_ES_PORT', '59201'))
+
+# Testing.
+
+TEST_RUNNER = 'resolwe.test_helpers.test_runner.ResolweRunner'
+TEST_PROCESS_REQUIRE_TAGS = True
+TEST_PROCESS_PROFILE = True
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,11 @@
 [tox]
-envlist = py{34,35},docs,linters,packaging,extra,migrations
+envlist =
+    py{34,35}{,-partial},
+    docs,
+    linters,
+    packaging,
+    extra,
+    migrations
 skip_missing_interpreters = True
 
 [tox:jenkins]
@@ -28,6 +34,10 @@ commands =
     )\'
 # run tests
     python tests/manage.py test {env:TEST_SUITE:resolwe_bio} --noinput \
+        # only test changes against a given Git commit reference for *-partial
+        # testing environments (default to "master" branch)
+        py{27,34,35}-partial: --only-changes-to {env:RESOLWE_TEST_ONLY_CHANGES_TO:master} \
+        py{27,34,35}-partial: --changes-file-types .resolwebio-filetypes.yml \
         --verbosity 2 --parallel
 whitelist_externals = bash
 setenv =
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ sftp-config.json
 !.env
 !.scripts/
 !.placeholder
+!.resolwebio-filetypes.yml
 
 # Ignore backup files
 *.orig
--- a/.resolwebio-filetypes.yml
+++ b/.resolwebio-filetypes.yml
@@ -1,0 +1,29 @@
+# Processes.
+- type: process
+  match: '^resolwe_bio/processes/.+\.ya?ml$'
+# Tests.
+- type: test
+  match: '^resolwe_bio/tests/.+\.py$'
+# Test data files.
+- type: force_run
+  name: test data files
+  match: '^resolwe_bio/tests/files/.+$'
+# Other code.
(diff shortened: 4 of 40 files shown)


kostko:
We currently detect the following kind of changed files:
* process definitions
* tests
* other code
* documentation
* package manifest
* git metadata
* test runner
* unknown (everything else)

Modification of any "unknown files" or "other code" will trigger a full test suite to be executed. Should we detect anything else and not trigger the full test suite in those cases?

dblenkus:
How is treated ["tools"](https://github.com/genialis/resolwe-bio/tree/master/resolwe_bio/tools) dir?

I think it would be (generally) good idea to list required tools under process requirements and only map them to the docker container. Then it would be much easier to determine which processes are affected by changed tool. But this requires quite some work, so it can be added later.

kostko:
> How is treated "tools" dir?

Since there are all Python scripts it is either treated as "other code" or "unknown", both of which will trigger the full test suite to be executed.

> I think it would be (generally) good idea to list required tools under process requirements and only map them to the docker container. Then it would be much easier to determine which processes are affected by changed tool. But this requires quite some work, so it can be added later.

Yes, without such requirements it would be just annoying to infer this automatically. You could match based on filenames that just happen to be in the process' bash program, but this doesn't seem right to me.

I agree that the requirements based approach would be best. Also, such tools should be distributed together with the process not in some strange directory. In this case, the process should be a directory and it should "include" additional files. Another way is to require specific Docker images which include these tools. The current approach with "magic tools directories, which include everything" is quite bad.

mstajdohar:
> I agree that the requirements based approach would be best ... The current approach with "magic tools directories, which include everything" is quite bad.

We could remove this functionality altogether and assume the process authors will add their scripts to the Docker image directly. @jkokosar?

kostko:
> We could remove this functionality altogether and assume the process authors will add their scripts to the Docker image directly. @jkokosar?

This would be the easiest solution.

jkokosar:
> We could remove this functionality altogether and assume the process authors will add their scripts to the Docker image directly. @jkokosar, what is the impact of this on convenience?

@mstajdohar This would require every change of the processes that utilise the tools scripts to be a at least three step affair: docker image change, image tagging, final PR where the the tool change is used in the processes itself. The development cycle would be extended. We would also lose the convenience of git code tracking within the same repo. I imagine that the tools folder would need to become its own repository, from where we would install this toolkit in the docker images.

I would suggest the removal of the tools folder when the Python syntax for processes is supported, since than only the R code would remain problematic.

mstajdohar:
What's the status of this PR? Can we merge it? Please rebase.

kostko:
This PR is waiting for review. I will rebase after review.

mstajdohar:
Ah, sorry. ping @tjanez

tox.ini
```diff
@@ -28,7 +28,9 @@ commands =
     )\'
 # run tests
     python tests/manage.py test {env:TEST_SUITE:resolwe_bio} --noinput \
-        --verbosity 2 --parallel
+        --verbosity 2 --parallel \
+        --only-changes-to origin/{env:CHANGE_TARGET} \
```
tjanez: This won't work if a user runs Tox locally where the `CHANGE_TARGET` environment variable is not defined or on Jenkins when testing a branch that is not a pull request (e.g. `master`) where this environment variable is not defined.

Since Tox doesn't support conditional statements I would use its ability to create generative testing environments and define `-partial` testing environments:

```ini
envlist =
    py{27,34,35}{,-partial},
    docs,
    linters,
    packaging,
    extra,
    migrations
```

and then use this here to only conditionally include the added arguments:

```ini
# run tests
    python tests/manage.py test {env:TEST_SUITE:resolwe_bio} --noinput \
        # only test changes against a given Git commit reference for *-partial
        # testing environments (default to "master" branch)
        py{27,34,35}-partial: --only-changes-to {env:RESOLWE_TEST_ONLY_CHANGES_TO:master} \
        py{27,34,35}-partial: --changes-file-types .resolwebio-filetypes.yml \
        --verbosity 2 --parallel
```

The `Jenkinsfile` would need to be updated to handle this:
```
if (env.CHANGE_TARGET) {
    // run partial test suite depending on detected changes to the target
    // branch if we are testing a pull request
    withEnv(["RESOLWE_TEST_ONLY_CHANGES_TO=origin/${env.CHANGE_TARGET}"]) {
        sh "tox -e py34-partial"
    }
} else {
    sh "tox -e py34"
}
```

It would be nice if you also updated the documentation on [Using Tox](http://resolwe-bio.readthedocs.io/en/stable/contributing.html#using-tox) and mention the `-partial` version of test environments and how a developer can control the Git commit against which changes are calculated with `RESOLWE_TEST_ONLY_CHANGES_TO` environment variable.

tjanez:
> We could remove this functionality altogether and assume the process authors will add their scripts to the Docker image directly. @jkokosar, what is the impact of this on convenience?

I would also like this approach.

> @mstajdohar This would require every change of the processes that utilise the tools scripts to be a at least three step affair: docker image change, image tagging, final PR where the the tool change is used in the processes itself. The development cycle would be extended.

@jkokosar, this is a known deficiency in the current (lacking) implementation of Docker images. We plan to address this by figuring out how to be able to make Docker image and code changes in one commit and the CI system being able to build temporary Docker images against which the tests are run.

> We would also lose the convenience of git code tracking within the same repo.

The tools would not need to live in a separate git repository, they could remain here, just in the directories where Docker image build files live.

> I imagine that the tools folder would need to become its own repository, from where we would install this toolkit in the docker images.

I think that would be unnecessary.

tests/settings.py
```diff
@@ -38,6 +38,7 @@
     'resolwe.flow',
     'resolwe.elastic',
     'resolwe.toolkit',
+    'resolwe.test_helpers',
```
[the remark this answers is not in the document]
kostko: Done.

tox.ini
kostko: Good idea, done.

PR #389, opened 2017-10-17
 Support new Resolwe's manager in test settings

dblenkus:


docs/conf.py
```diff
@@ -22,6 +22,8 @@
         'resolwe.flow',
         'resolwe_bio',
     ),
+    FLOW_MANAGER={'REDIS_PREFIX': 'xxx'},
+    FLOW_EXECUTOR={},
```
dblenkus: @jberci please make sure Resolwe will run even if these settings will be missing.

jberci: This should work now, I think, so you don't need these lines anymore.

tests/routing.py
```diff
@@ -0,0 +1,12 @@
...
+# from channels.routing import route
+
+from channels.routing import RouteClass
+
+from resolwe.flow.managers import manager
+
+
+manager.update_routing()
+channel_routing = [  # pylint: disable=invalid-name
+    RouteClass(type(manager)),
```
jberci: Same as in PR #308 in resolwe, use `route_class` here instead of RouteClass. It's a drop-in replacement.

tests/settings.py
```diff
@@ -45,6 +46,8 @@
 
 ROOT_URLCONF = 'tests.urls'
 
+TEST_RUNNER = 'resolwe.flow.utils.runner.RunnerWithManagerContext'
```
jberci: This should be `resolwe.test_helpers.test_runner.ResolweRunner` these days.

PR #390, opened 2017-10-18
Update resolwebio/latex docker image

JureZmrzlikar:


resolwe_bio/docker_images/latex/packages-manual/README.txt
```diff
@@ -0,0 +1 @@
+This file enables committing `packages-manual` folder.
```
kostko: Instead of this you could also create a `.placeholder` file.

JureZmrzlikar: This would require to remove `.*` line in `.gitignore` file. I'm unsure if want to do that...

kostko: No, you can just unignore `.placeholder` by adding `!.placeholder`.

PR #397, opened 2017-10-26
Add AF filter to amplicon report

Teja91:


resolwe_bio/tests/files/report_template.tex
```diff
@@ -86,6 +86,8 @@
 \noindent
 {\lightfont {#PANEL#} | \today}
 
+\medskip
+\noindent Results have been filtered to exclude any putative variant detected below AF = {{#AF_THRESHOLD#}}\%. Variants below this threshold may be inspected manually in the unfiltered VCF files.
```
jkokosar: Is the wording correct? Variants _above_ this threshold ... ?

Teja91: It is. The report excludes variants below the threshold. If the user wants to see those that were excluded (so below threshold), they should check the VCF files.

jkokosar:
Please add a CHANGELOG entry and rebase to the latest master

JureZmrzlikar:
Rebased and added CHANGELOG.

PR #399, opened 2017-10-27
Add species to feature primary key

kostko:


resolwe_bio/kb/migrations/0005_species.py
```diff
@@ -0,0 +1,54 @@
...
+
+    dependencies = [
+        ('resolwe_bio_kb', '0004_add_unique_together'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='mapping',
+            name='source_species',
+            field=models.CharField(default='unknown', max_length=50),
```
kostko: We could infer this based on existing features in the database instead of using unknown. But when this change will be merged, we will probably repopulate all mappings anyway, so this should not be a problem?

resolwe_bio/kb/management/commands/insert_features.py
```diff
@@ -115,6 +115,7 @@ def handle(self, *args, **options):
 
                 feature, created = Feature.objects.get_or_create(source=row['Source'],
                                                                  feature_id=row['ID'],
+                                                                 species=row['Species'],
```
dblenkus: Remove `species` from `values`.

resolwe_bio/kb/migrations/0005_species.py
dblenkus: Aren't there some features that doesn't have unique `(db, id)` combination? I think that it is safer (and easier) to repopulate mapings.

kostko: Yes, this is correct, but the old database could not hold those anyway, because previously `(db, id)` were the primary key. I agree that it is safer to repopulate mappings, so we should leave it as `unknown` in this migration.

PR #400, opened 2017-10-27
Adjust bam-split process so it can be included in workflows

jvrakor:
Included in https://github.com/genialis/resolwe-bio/pull/420

### DIFF
--- a/resolwe_bio/processes/support_processors/bam_split.yml
+++ b/resolwe_bio/processes/support_processors/bam_split.yml
@@ -10,9 +10,10 @@
     executor:
       docker:
         image: resolwebio/rnaseq:1.1.0
-  data_name: "Split ({{ bam|sample_name|default('?') }})"
-  version: 0.0.2
-  type: data:split
+  data_name: "{{ bam.bam.file.split('.')[0] }}_{{ bam.build.split('/')[0] }}.bam"
+  version: 0.1.0
+  type: data:alignment:bam:primary
+  flow_collection: sample
   category: other
   persistence: CACHED
   description: >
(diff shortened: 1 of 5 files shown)


resolwe_bio/tools/build_species_mapper.py
```diff
@@ -0,0 +1,15 @@
...
+
+parser = argparse.ArgumentParser(description="Map species to genome build")
+parser.add_argument('build', help="Genome build")
+args = parser.parse_args()
+
+mapper = {
+    'dm6': 'Drosophila melanogaster',
+}
+
+print(mapper[args.build])
```
JureZmrzlikar: When this will get `build` value that is not present in the `mapper` it will fail with `KeyError`. It would be nice if you could raise a more descriptive error (preferably) or provide a default value (do we want that?).

resolwe_bio/processes/support_processors/bam_split.yml
```diff
@@ -56,18 +55,23 @@
...
-      BAM={{bam.bam.file}}
+      BAM={{ bam.bam.file }}
       NAME=`basename {{ bam.bam.file }} .bam`
-      ORG1={{organism}}
-      ORG2={{organism2}}
+      BUILD={{ bam.build }}
+      BUILD1=${BUILD%/*}
+      BUILD2=${BUILD##*/}
+      SPECIES1={{ bam.species }}
+      SPECIES2=`build_species_mapper.py ${BUILD2}`  # Update when new hybrid organism
```
mzganec: This can be done directly in the process by using a [jinja dict](http://jinja.pocoo.org/docs/2.10/templates/#dict) making the `build_species_mapper.py` script redundant.

PR #405, opened 2017-11-07
Fix merging multiple expressions in DESeq process

mzganec:


mstajdohar:
", change tests" is not needed in a commit message.

mstajdohar:
Test fails. Did you update the results?

PR #409, opened 2017-11-09
Add primerclip, samtools and picard to dnaseq docker

jkokosar:


resolwe_bio/docker_images/dnaseq/packages-debian.txt
```diff
@@ -1,3 +1,5 @@
+bedtools
+bwa
```
mzganec: Add `BWA` to change log and commit message.

PR #411, opened 2017-11-09
Add antibody information to reads descriptor schema

jvrakor:
@jkokosar @mstajdohar


jkokosar:
Bump the descriptor schema version

jvrakor:
It doesn't have a version now. So I should just set it to 0.0.1?

jkokosar:
Yes

PR #412, opened 2017-11-10
Update ES indices to work with Resolwe 4.0

dblenkus:


dblenkus:
Resolwe used to push each document separately, so we had to use this to push all documents at once and thus speed up indexing. But [now this is the default behaviour](https://github.com/genialis/resolwe/commit/5de4980db51310b5eeb8474d7b91f9ef4bdba6db) in Resolwe, so it is no longer needed.

PR #413, opened 2017-11-13
Chip seq docker

jvrakor:
@jkokosar @mstajdohar


kostko:
I think it would be better if you refactor the Docker image to be based on `resolwebio/base`. I see that it still uses Ubuntu 14.04, so perhaps you should try 16.04 or later first.

jkokosar:
@kostko @jvrakor @dblenkus This image installs the bamliquidator tool, which cannot be built on 16.04. It might still make sense to build it using the new framework, but do not attempt building it on 16.04 since we couldn't get it to compile in the past.

kostko:
> @kostko @jvrakor @dblenkus This image installs the bamliquidator tool, which cannot be built on 16.04. It might still make sense to build it using the new framework, but do not attempt building it on 16.04 since we couldn't get it to compile in the past.

Does it by any chance use Boost libraries and the compilation issue was related to Boost? If so, then perhaps it will build on 17.04 or later, have you tried that?

jkokosar:
@jvrakor Please try what Kostko suggested. I do not remember what was causing the issues with 16.04.

resolwe_bio/docker_images/chipseq/README.md
```diff
@@ -1,13 +1,14 @@
...
 
 Included bioinformatics tools
 -----------------------------
 * ROSE2 (v1.0.5)
 * BAMPLOT (v0.9.0)
-* MACS (v1.4.2)
+* MACS (v1.4.3)
 * MACS2 (v2.1.1.20160309)
 * BAMLIQUIDATOR (v1.3)
 * SAMTOOLS (v0.1.19)
```
jkokosar: samtools version is probably not correct anymore

jvrakor: samtools version is the same

jkokosar:
@kostko Plase review

PR #415, opened 2017-11-16
Update RNASeq docker image

mzganec:


mstajdohar:
Include a more comprehensive description in README.

This PR substitutes Python 2 with Python 3. Legacy processes will keep using the old Docker with Python 2. If/when we update a legacy process, we should also update the Python scripts that are called in the process and use the latest Docker image.

PR #417, opened 2017-11-22
Add wigtobigwig.sh to chipseq docker image folder

jvrakor:
@jkokosar @mstajdohar


kostko:
Wasn't this already there? At least in my commit it was, so it got removed?

jvrakor:
It was there in my original commit and in your commit. Then I refactored the commits because I had to change some files and this became untracked and I didn't notice that.

PR #420, opened 2017-11-24
Refactor macs14 process

jvrakor:

Includes https://github.com/genialis/resolwe-bio/pull/400

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -13,6 +13,8 @@ Unreleased
 Added
 -----
 * Add Ubuntu 17.10 base Docker image
+* Add ``species`` and ``genome build`` output fields to ``macs14``
+  process
 
 Changed
 -------
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -58,6 +58,8 @@ Changed
   and ``feature_type`` in ``upload-expression-star`` process
 * The description text has been updated to better describe the difference
   between gene/transcript-type analyses in `cutadapt-star-htseq` descriptor schema
+* Remove ``gsize`` input from ``macs14`` process and automate genome
+  size selection
 
 Fixed
 -----
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -60,6 +60,7 @@ Changed
   between gene/transcript-type analyses in `cutadapt-star-htseq` descriptor schema
 * Remove ``gsize`` input from ``macs14`` process and automate genome
   size selection
+* Adjust bam-split process so it can be included in workflows
 
 Fixed
 -----
--- a/resolwe_bio/processes/chip_seq/macs14.yml
+++ b/resolwe_bio/processes/chip_seq/macs14.yml
@@ -11,7 +11,7 @@
       docker:
         image: resolwebio/chipseq:1.1.0
   data_name: "MACS results ({{ treatment|sample_name|default('?') }})"
-  version: 2.1.0
+  version: 2.2.0
   type: data:chipseq:macs14
   category: analyses
   persistence: CACHED
@@ -25,23 +25,6 @@
       label: BAM Background File
       type: data:alignment:bam
       required: false
-    - name: gsize
-      label: Effective genome size
-      type: basic:string
-      default: '2.7e9'
-      choices:
-        - label: Homo sapiens
-          value: '2.7e9'
-        - label: Mus musculus
-          value: '1.87e9'
-        - label: Caenorhabditis elegans
-          value: '9e7'
-        - label: Drosophila melanogaster
-          value: '1.2e8'
-        - label: Rattus norvegicus
-          value: '2e9'
-        - label: Dictyostelium discoideum
-          value: '3.4e7'
     - name: pvalue
       label: P-value
       type: basic:string
@@ -92,8 +75,31 @@
         re-error "BAM {{ treatment|name }} has {{ treatment.species }} while {{ control|name }} has {{ control.species }} species information.\n"
       {% endif %}
 
-      CONTROL='{% if control %}--control {{ control.bam.file }} {% endif %}'
-      macs --treatment {{ treatment.bam.file }} ${CONTROL} --format BAM --gsize {{ gsize }} --name $NAME --pvalue {{ pvalue }} --wig --single-profile --space=50
+      {% set species_to_gsize = {
+        'Homo sapiens': '2.7e9',
+        'Mus musculus': '1.87e9',
+        'Dictyostelium discoideum': '3.4e7',
+        'Drosophila melanogaster': '1.2e8',
+        'Caenorhabditis elegans': '9e7',
+        'Rattus norvegicus': '2e9',
+        }
+      %}
+
+      {% set gsize = species_to_gsize[treatment.species] %}
+      {% if not gsize %}
+        re-error "{{ treatment.species }} is not a valid hybrid species. Accepted species are: {{ species_to_gsize.keys() | join(', ') }}"
+      {% endif %}
+
+      macs \
+        --treatment {{ treatment.bam.file }} \
+        {% if control %}--control {{ control.bam.file }}{% endif %} \
+        --format BAM \
+        --gsize {{ gsize }} \
+        --name $NAME \
+        --pvalue {{ pvalue }} \
+        --wig \
+        --single-profile \
+        --space=50
       re-checkrc
 
       zip -r ${NAME}_MACS_wiggle.zip ${NAME}_MACS_wiggle
--- a/resolwe_bio/processes/chip_seq/macs14.yml
+++ b/resolwe_bio/processes/chip_seq/macs14.yml
@@ -11,7 +11,7 @@
       docker:
         image: resolwebio/chipseq:1.1.0
   data_name: "MACS results ({{ treatment|sample_name|default('?') }})"
-  version: 2.0.5
+  version: 2.1.0
   type: data:chipseq:macs14
   category: analyses
   persistence: CACHED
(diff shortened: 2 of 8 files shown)


resolwe_bio/tools/species_to_gsize.py
```diff
@@ -0,0 +1,27 @@
...
+        'mus musculus': '1.87e9',
+        'dictyostelium discoideum': '3.4e7',
+        'drosophila melanogaster': '1.2e8',
+        'caenorhabditis elegans': '9e7',
+        'rattus norvegicus': '2e9',
+    }
+    return mapping[name.lower()]
+
+
+print(gsize_organism(args.species))
```
JureZmrzlikar: When this will get `species` value that is not present in the `mapping` it will fail with KeyError. It would be nice if you could raise a more descriptive error.

resolwe_bio/tools/species_to_gsize.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env python2
```
jkokosar: Refactor to Python 3 script.

resolwe_bio/processes/chip_seq/macs14.yml
```diff
@@ -72,14 +55,37 @@
...
+      label: Build
+      type: basic:string
   run:
     runtime: polyglot
     language: bash
     program: |
       NAME=`basename {{ treatment.bam.file }} .bam`
+      
+      SPECIES={{ treatment.species }}
+      {% set dict = {
```
JureZmrzlikar: Choose a more descriptiove name, for example `gsize` or `species_to_gsize`.

resolwe_bio/processes/chip_seq/macs14.yml
```diff
@@ -72,14 +55,37 @@
...
+      GSIZE={{ dict[treatment.species] }}
+      SPECIES_LIST="{% for species in dict.keys() %}{{species}}, {% endfor %}"
+      SPECIES_LIST=${SPECIES_LIST%??} 
+      if [ -z $GSIZE ]; then
+        re-error "${SPECIES} is not a valid hybrid species. Accepted species are: ${SPECIES_LIST}."
+      fi
 
       CONTROL='{% if control %}--control {{ control.bam.file }} {% endif %}'
-      macs --treatment {{ treatment.bam.file }} ${CONTROL} --format BAM --gsize {{ gsize }} --name $NAME --pvalue {{ pvalue }} --wig --single-profile --space=50
+      macs --treatment {{ treatment.bam.file }} ${CONTROL} --format BAM --gsize ${GSIZE} --name $NAME --pvalue {{ pvalue }} --wig --single-profile --space=50
```
JureZmrzlikar: Revrite into more readable form, like [here](https://github.com/genialis/resolwe-bio/blob/master/resolwe_bio/processes/alignment/star.yml#L208).

resolwe_bio/processes/chip_seq/macs14.yml
```diff
@@ -72,14 +55,37 @@
...
+    - name: build
+      label: Build
+      type: basic:string
   run:
     runtime: polyglot
     language: bash
     program: |
       NAME=`basename {{ treatment.bam.file }} .bam`
+      
+      SPECIES={{ treatment.species }}
```
JureZmrzlikar: Unused variable.

resolwe_bio/processes/chip_seq/macs14.yml
```diff
@@ -72,14 +55,37 @@
...
+        'Dictyostelium discoideum': '3.4e7',
+        'Drosophila melanogaster': '1.2e8',
+        'Caenorhabditis elegans': '9e7',
+        'Rattus norvegicus': '2e9',
+        }
+      %}
+      GSIZE={{ dict[treatment.species] }}
+      SPECIES_LIST="{% for species in dict.keys() %}{{species}}, {% endfor %}"
+      SPECIES_LIST=${SPECIES_LIST%??} 
+      if [ -z $GSIZE ]; then
```
JureZmrzlikar: Use jinja syntax for this. Define `SPECIES_LIST` only if you need it (in case there is no matching species). Something like this:
{% if dict.treatment.species %}
  SPECIES_LIST=...
  re-error "{{treatment.species}} is not a valid hybrid species. Accepted species are: ${SPECIES_LIST}."
{% endif %}

resolwe_bio/processes/chip_seq/macs14.yml
JureZmrzlikar: Also note that code from #400 is duplicated here (regarding `upload-bam` and `upload-bam-indexed` process)

jkokosar:
@JureZmrzlikar Please re-review

docs/CHANGELOG.rst
```diff
@@ -56,6 +56,7 @@ Added
 -----
 
 * Add Ubuntu 17.10 base Docker image
+* Add ``species`` and ``organism`` output fields to ``macs14`` process
```
mstajdohar: Not organism, a genome build output was added. Please change the log entry and also update the commit text.

resolwe_bio/processes/chip_seq/macs14.yml
```diff
@@ -99,3 +105,6 @@
       then
         re-save-file neg_peaks ${NAME}_negative_peaks.xls
       fi
+
+      re-save species {{ treatment.species }}
```
mstajdohar: I suggest that you check in the beginning of the process that the treatment and control inputs have the same species and build. If not, raise an informative error.

resolwe_bio/processes/chip_seq/macs14.yml
mstajdohar: I agree with Jure on this, please use the Jinja synatax. In addition to Jure's suggestions, this may be useful too:
```
{% set gsize = species_to_gsize[treatment.species]  %}
```

resolwe_bio/processes/chip_seq/macs14.yml
```diff
@@ -83,9 +66,34 @@
...
+        re-error "${SPECIES} is not a valid hybrid species. Accepted species are: ${SPECIES_LIST}."
+      fi
 
-      CONTROL='{% if control %}--control {{ control.bam.file }} {% endif %}'
-      macs --treatment {{ treatment.bam.file }} ${CONTROL} --format BAM --gsize {{ gsize }} --name $NAME --pvalue {{ pvalue }} --wig --single-profile --space=50
+      macs \
+        --treatment {{ treatment.bam.file }} \
+        {% if control %}--control {{ control.bam.file }}{% endif %} \
+        --format BAM \
+        --gsize ${GSIZE} \
```
mstajdohar: If you define the gsize in Jinja you can simply use `{{ gsize }}` here.

resolwe_bio/processes/import_data/bam.yml
```diff
@@ -180,56 +180,6 @@
           re-error "BAI should have the same name as BAM with .bai extension"
         fi
 
-- slug: upload-bam-primary
```
mstajdohar: Why is this not needed anymore?

resolwe_bio/processes/import_data/bam.yml
```diff
@@ -240,7 +190,7 @@
     resources:
       network: true
   data_name: '{{ src.file|default("?") }}'
-  version: 0.0.1
+  version: 0.1.0
   type: data:alignment:bam:secondary
```
mstajdohar: Why do we need this, why can't we just use a regular bam upload?

resolwe_bio/processes/import_data/bam.yml
```diff
@@ -260,13 +210,31 @@
       description: >
         A mapping file in BAM format. The file will be indexed on upload, so additional BAI files are not required.
       validate_regex: '\.(bam)$'
+    - name: species
+      label: Species
+      type: basic:string
+      description: >
+        Species latin name.
```
mstajdohar: How about a more formal description, e.g. Binomial nomenclature. @jkokosar can comment what biologists use. I have found that term in the [Wikipedia](https://en.wikipedia.org/wiki/Binomial_nomenclature).

resolwe_bio/processes/import_data/bam.yml
jvrakor: because `bam-split` process data object is now of a `data:alignment:bam:primary` type so it can be directly used in workflows (process spawned processes can't be used).

resolwe_bio/processes/import_data/bam.yml
jvrakor: Because we decided that all 3 data objects (original bam and 2 resulting bam-split objects) will live on a sample. We want to distinguish between them so they can be used in future analyses.

resolwe_bio/processes/import_data/bam.yml
```diff
@@ -260,13 +210,31 @@
...
+      allow_custom_choice: true
```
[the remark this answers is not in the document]
jvrakor: for now no, but I expected there would be in the future

resolwe_bio/processes/chip_seq/macs14.yml
```diff
@@ -72,14 +55,45 @@
...
+        'Homo sapiens': '2.7e9',
+        'Mus musculus': '1.87e9',
+        'Dictyostelium discoideum': '3.4e7',
+        'Drosophila melanogaster': '1.2e8',
+        'Caenorhabditis elegans': '9e7',
+        'Rattus norvegicus': '2e9',
+        }
+      %}
+      GSIZE={{ species_to_gsize[treatment.species] }}
+      if [ -z $GSIZE ]; then
```
JureZmrzlikar: IMO, this should be rewritten into Jinja syntax. What do you think, @mstajdohar, @jkokosar ?

PR #422, opened 2017-11-27
Fix BAM file input in rsem process

jkokosar:


jkokosar:
@mstajdohar please re-review

PR #425, opened 2017-11-29
Add upload-metabolic-pathway process

matyasfodor:


matyasfodor:
retest this please

mstajdohar:
@JureZmrzlikar Would you do the first review pass?

resolwe_bio/processes/import_data/pathway_map.yml
```diff
@@ -0,0 +1,34 @@
...
+  output:
+    - name: pathway
+      label: Pathway json
+      type: basic:json
+  run:
+    runtime: polyglot
+    language: bash
+    program: |
+      re-import {{ upload.file_temp }} {{ upload.file }} "json" "json" 0.5
+      re-save pathway ${NAME}.json
```
mstajdohar: Add an empty line at the end of the file. You can configure your editor to do that for you automatically.

resolwe_bio/processes/import_data/pathway_map.yml
```diff
@@ -0,0 +1,34 @@
+- slug: upload-metabolic-pathway
+  name: Metabolic Pathway File
+  requirements:
+    expression-engine: jinja
+    executor:
+      docker:
+        image: resolwebio/legacy:latest
```
mstajdohar: You can probably use a basic `resolwe/base:ubuntu-17.04` image.

resolwe_bio/processes/import_data/pathway_map.yml
```diff
@@ -0,0 +1,34 @@
...
+    resources:
+      network: true
+  data_name: Metabolic Pathway
+  version: 0.1.0
+  type: data:pathway
```
mstajdohar: Change to `data:metabolic-pathway`.

resolwe_bio/processes/import_data/pathway_map.yml
```diff
@@ -0,0 +1,34 @@
...
+  category: upload
+  persistence: RAW
+  scheduling_class: interactive
```
JureZmrzlikar: I am not sure about this, but import processes should not have interactive `scheduling_class`, right, @mstajdohar ?

resolwe_bio/processes/import_data/pathway_map.yml
```diff
@@ -0,0 +1,34 @@
...
+  description: >
+    Upload pathway json.
+  input:
+    - name: upload
```
JureZmrzlikar: We usually use name `src` for files that need to be uploaded. This is not mandatory, but it is nice if all the processes share the same naming convention.

resolwe_bio/processes/import_data/pathway_map.yml
JureZmrzlikar: Newline missing at the end of a file.

resolwe_bio/tests/processes/test_pathway_map.py
```diff
@@ -0,0 +1,15 @@
...
+
+from resolwe_bio.utils.test import BioProcessTestCase
+
+
+class PathwayMapTestCase(BioProcessTestCase):
+
+    @tag_process('upload-metabolic-pathway')
+    def test_upload_metabolic_pathway(self):
+        # This one is the original map file, it's rather big (141 Kb)
+        # Is it ok to leave it, or shall I use a dummy JSON instead?
```
JureZmrzlikar: Please use a dummy json instead.

JureZmrzlikar:
Test tags mark which process is put under test in a test case. But quite often many other processes are run in a test case to prepare inputs for the actual process being tested. These should be under the `preparation_stage` context manager. In your case, you just call the actual process being tested, so no need for `preparation_stage`.

resolwe_bio/processes/import_data/pathway_map.yml
matyasfodor: I think it's not allowed: `data:metabolic-pathway:' does not match '^data:[a-z0-9:]+:$`
Shall I make it camelCase?

mstajdohar: @jkokosar ?

resolwe_bio/processes/import_data/pathway_map.yml
mstajdohar: Maybe they can if they are really, really fast. @dblenkus ?

dblenkus: Interactive scheduling class means that UI depends on the results of the process. To make it efficient, such processes are limited to 30 seconds (excluding the upload). If this process meets both conditions then it can (and should) be interactive.

resolwe_bio/processes/import_data/pathway_map.yml
jkokosar: I such cases, I would usually just concat the words without camelCase: `data:metabolicpathway`

resolwe_bio/processes/import_data/pathway_map.yml
```diff
@@ -0,0 +1,34 @@
...
+  name: Metabolic Pathway File
+  requirements:
+    expression-engine: jinja
+    executor:
+      docker:
+        image: resolwe/base:ubuntu-17.04
+    resources:
+      network: true
+  data_name: Metabolic Pathway
+  version: 0.1.0
```
jkokosar: You can start with 0.0.1

resolwe_bio/processes/import_data/pathway_map.yml
```diff
@@ -0,0 +1,34 @@
...
+  scheduling_class: interactive
+  description: >
+    Upload pathway json.
+  input:
+    - name: src
+      label: Pathway file
+      type: basic:file
+      description: >
+        JSON file. Supported extensions: .json.gz
+      validate_regex: '(\.json)(|\.gz|\.bz2|\.tgz|\.tar\.gz|\.tar\.bz2|\.zip|\.rar|\.7z)$'
```
jkokosar: Match the regex/description info.

resolwe_bio/tests/processes/test_pathway_map.py
```diff
@@ -0,0 +1,13 @@
...
+from resolwe.test import tag_process
+
+from resolwe_bio.utils.test import BioProcessTestCase
+
+
+class PathwayMapTestCase(BioProcessTestCase):
+
+    @tag_process('upload-metabolic-pathway')
+    def test_upload_metabolic_pathway(self):
+        inputs = {'src': 'matabolic-pathway.json.gz'}
```
jkokosar: Rename the test file so that it includes the whitespace, e.g. `matabolic pathway.json.gz`. In the process, you might need to put quotes around the NAME variable: `re-save pathway "${NAME}.json"`

resolwe_bio/processes/import_data/pathway_map.yml
matyasfodor: Apart from upload this process is just saving a JSON to the DB, which is around 0.5 Mb. I think it can be done under 30s.

jkokosar: Will this maps be continuously saved to DB (as per user action), or are they saved once and reused?

matyasfodor: Saved once and reused, so I guess making it non-interactive is also fine.

jkokosar: In that case, it is better to make it non-interactive, so that it does not compete with all the interactive processes for resources.

matyasfodor:
@jkokosar I updated the PR. I think it's ready for the final review.

PR #428, opened 2017-12-01
Add memory-optimized primerclip as a separate process

jkokosar:


resolwe_bio/processes/amplicon/bwa_trim.yml
```diff
@@ -9,11 +9,11 @@
     expression-engine: jinja
     executor:
       docker:
-        image: resolwebio/dnaseq:1.1.0
+        image: resolwebio/legacy:latest
     resources:
       memory: 32768
   data_name: "Alignment ({{ reads|sample_name|default('?') }})"
-  version: 0.1.0
+  version: 0.0.8
```
mstajdohar: The version should be higher then 0.1.0

resolwe_bio/processes/amplicon/bwa_trim.yml
```diff
@@ -193,7 +193,7 @@
       echo "trimming primers"
       primerclip \
         "${NAME}_align_unsorted.sam" \
-        {{master_file.master_file.file}} \
+        {{master_file.primers.file}} \
```
mstajdohar: Will this work with the new masterfile?

resolwe_bio/processes/amplicon/bwa_trim.yml
jkokosar: The process with version 0.1.0 was removed from the database, the current one is 0.0.7, right?

resolwe_bio/processes/amplicon/bwa_trim.yml
jkokosar: This is legacy process, for the old primerclip

resolwe_bio/processes/amplicon/bwa_trim.yml
mstajdohar: Not on all deploys.

jkokosar:
Closed in favour of https://github.com/genialis/resolwe-bio/pull/432

PR #429, opened 2017-12-01
Fix file processing in cutadapt-custom-paired process

jkokosar:

PR #430, opened 2017-12-04
Testing if the same PR will work in a new namespace

mstajdohar:


mstajdohar:
This is the same as https://github.com/genialis/resolwe-bio/pull/429. Testing Jenkins.

mstajdohar:
Included in #431

PR #431, opened 2017-12-04
Use latest Resolwe, fix tests for new total_size field

kostko:
Resolwe 5.1 contains test runner fixes.


dblenkus:
We are trying to stop using `dependency_links` and create an actual alpha release instead. I think @tjanez is just going to create one anyway.

dblenkus:
+1 for fixing `size` fields :)

tjanez:
> We are trying to stop using `dependency_links` and create an actual alpha release instead. I think @tjanez is just going to create one anyway.

Correct. I'll prepare a pre-release of Resolwe (I just need input on the potential backward incompatible changes in Resolwe `master` to set the correct version).

I'll let you know once it is done.

kostko:
> Regarding total_size field test changes, do we want to fix those numbers or would it be better to improve assertFields to allow just checking presence of some fields?

I think the latter would be best as I had to remove some fields due to non-determinism (e.g. file size changed on each test run). In addition to field presence we should also validate field type.

But perhaps we want to handle that in a separate PR?

tjanez:
> I think the latter would be best as I had to remove some fields due to non-determinism (e.g. file size changed on each test run). In addition to field presence we should also validate field type.

Yes, good idea.

> But perhaps we want to handle that in a separate PR?

Yes, agreed.

PR #434, opened 2017-12-08
WIP: DD-DeCaF pathway predictor

sgalkina:


resolwe_bio/processes/modeling/pathway_predictor.yml
```diff
@@ -1,13 +1,16 @@
...
         image: dddecaf/resolwe-pathway-predictor:latest
+    resources:
+      network: true
+      cores: 4
   data_name: "Predict heterologous metabolic pathways"
   version: 0.0.1
-  type: data:similarexpression
+  type: data:pathways
   persistence: TEMP
   scheduling_class: interactive
```
jkokosar: Processes with ``scheduling_class: interactive`` are run in a separate queue and are killed after 30s if not finished in time. This happens in the test case already, so you probably need to remove this.

resolwe_bio/tests/processes/test_pathway_predictor.py
```diff
@@ -4,9 +4,9 @@
...
-    @tag_process('pathway-predictor')
+class PathwayPredictorTestCase(ProcessTestCase):
+    @tag_process('pathways-predictor')
     def test_pathway_predictor(self):
-        inputs = {'model': 'one'}
-        annotation = self.run_process('pathway-predictor', inputs)
-        self.assertFields(annotation, 'pathway', 'l')
+        inputs = {'model': 'iMM904', 'product': 'methanol', 'n_pathways': 2}
+        annotation = self.run_process('pathways-predictor', inputs)
+        self.assertFields(annotation, 'pathway', '')
```
jkokosar: (annotation, 'pathway', '') -> (annotation, 'pathway**s**', '**test_output.json**')

resolwe_bio/tools/pathway_predictor.py
```diff
@@ -1,7 +1,26 @@
 #!/usr/bin/env python3
```
jkokosar: It much better to save all the output to a json file and than call the ``re-save pathways my_file.json`` in the process itself. For each addition to RESULTS list, you could print out something like ``print('{"proc.progress":0.3}'``. The progress value will show in the bioinfo dashboard as an analysis progress (30 % done in that case).

resolwe_bio/tools/pathway_predictor.py
```diff
@@ -1,7 +1,26 @@
...
+    return dict(
+        id=reaction.id,
+        name=reaction.name,
+        reaction_string=reaction.build_reaction_string(use_metabolite_names=True),
+    )
+
+
+def show_results(pathway):
+    RESULTS.append([reaction_to_dict(reaction) for reaction in pathway.reactions])
+    call(['re-save', 'pathways', json.dumps(RESULTS)])
```
jkokosar: [This ](https://github.com/genialis/resolwe-bio/blob/master/resolwe_bio/tools/geneset2storage.py#L30) is an example of code that would create an output json file to be saved with `re-save` in the .yml file

resolwe_bio/tools/pathway_predictor.py
```diff
@@ -1,7 +1,26 @@
...
+
+def reaction_to_dict(reaction):
+    return dict(
+        id=reaction.id,
+        name=reaction.name,
+        reaction_string=reaction.build_reaction_string(use_metabolite_names=True),
+    )
+
+
+def show_results(pathway):
```
jkokosar: Where do you use the function?

tests/settings.py
```diff
@@ -125,6 +125,9 @@
     'resolwe.flow.execution_engines.workflow',
 ]
 
+NOSE_ARGS = ['--nocapture',
```
jkokosar: Why is this needed?

resolwe_bio/processes/modeling/pathway_predictor.yml
```diff
@@ -1,13 +1,16 @@
-- slug: pathway-predictor
+- slug: pathways-predictor
   name: Predict heterologous metabolic pathways
   requirements:
     expression-engine: jinja
     executor:
       docker:
         image: dddecaf/resolwe-pathway-predictor:latest
```
jkokosar: `dddecaf/resolwe-pathway-predictor`

latest is the default, thus not needed here

resolwe_bio/processes/modeling/pathway_predictor.yml
```diff
@@ -1,13 +1,16 @@
...
+    resources:
+      network: true
+      cores: 4
   data_name: "Predict heterologous metabolic pathways"
```
jkokosar: You can construct a more informative data_name out of process inputs. [Example](https://github.com/genialis/resolwe-bio/blob/master/resolwe_bio/processes/go_enrichment/go.yml#L15) that includes gene names in the data_name

jkokosar:
Also, better organize the commits. Squash together what is necessary (modifications to the same file/lines). Add separate commit for settings change. Docker file addition should be in its own pull request (this is a limitation of the current Jenkins testing framework implementation).

Add changelog entry.

resolwe_bio/tools/pathway_predictor.py
sgalkina: but in this case the json file would be saved to MongoDB only after the script finished running, and what is needed is that the entry in MongoDB is updated with new set of reactions during the script execution

mstajdohar:
We have fixed the testing issues on resolwe-bio, so we can now merge your pull request. @jkokosar Please do another review pass. @sgalkina After Janes'z review, please rebase to master.

PR #435, opened 2017-12-11
Refactor amplicon report and multi-sample report

JureZmrzlikar:


JureZmrzlikar:
@jkokosar , @mstajdohar , i made a really ugly trick that works. Check how the template and logo file paths are brought in the process `amplicon-archive-multi-report`. I had no other idea how to get these file paths in the process otherwise. @mzganec suggested to extract this information with SDK script, but in the end, SDK can only extract basename of the file, so we still need to somehow hardcode the `/data_all/` in the path... Let me know what you think. This works locally.

resolwe_bio/processes/reports/amplicon_multi_report.yml
```diff
@@ -89,6 +79,11 @@
       mkdir multisample_report
       # Bash syntax for computing Array size is clashing with jinja comments syntax.
       # With "raw" we tell jinja to leave this line alone.
+      {% set template = 'multi-sample-template'|data_by_slug %}
+      {% set logo = 'genialis-logo'|data_by_slug %}
+      TEMPLATE='/data_all/{{ template | int }}/multireport-template.tex'
```
jkokosar: @mstajdohar @JureZmrzlikar Could we bundle the logo & template as a "resource" to a script in the docker image somehow?

JureZmrzlikar: We talked a bit about this with @dblenkus . He suggests thatwe modify frontend call to this process in way that is is able to accomodate other inputs or get them from configuration. But i guess this will take time...

mstajdohar: > He suggests thatwe modify frontend call to this process in way that is is able to accomodate other inputs or get them from configuration.

This would take too long. We either put this in Docker or write a DSS process that calls this and inserts the logo and template in the inputs. The DSS process would get the logo and template by slug. I do not believe there is and advantage of writing a DSS process over having the logo and template in the Docker.

dblenkus: Putting the template into the Docker image is preferred solution, if there is only one template.

JureZmrzlikar:
Depends on #438 . Otherwise, ready for review.

JureZmrzlikar:
I have extended `test_multisample_report` in `test_generate_report.py` to also test handling of legacy data. Changes were also made in `amplicon_multi_report.yml` . @jkokosar , @mstajdohar @dblenkus , please let me know if the way it is done is acceptable.

resolwe_bio/processes/reports/amplicon_multi_report.yml
```diff
@@ -5,8 +5,8 @@
     expression-engine: jinja
     executor:
       docker:
-        image: resolwebio/latex:1.0.2
-  version: 0.0.1
+        image: resolwebio/latex:test
```
jkokosar: Wrong tag

PR #437, opened 2017-12-12
Public version of modeling docker image

sgalkina:
Install cameo library and its dependencies, no proprietary solvers

### DIFF
--- a/resolwe_bio/docker_images/modeling/Dockerfile
+++ b/resolwe_bio/docker_images/modeling/Dockerfile
@@ -1,0 +1,6 @@
+FROM docker.io/resolwebio/base:ubuntu-16.04
+
+MAINTAINER Resolwe Bioinformatics authors https://github.com/genialis/resolwe-bio
+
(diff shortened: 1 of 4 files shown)


resolwe_bio/docker_images/modeling/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM docker.io/resolwebio/base:ubuntu-16.04
+
+MAINTAINER Resolwe Bioinformatics authors https://github.com/genialis/resolwe-bio
+
+COPY requirements.txt .
```
mzganec: Since this image is derived from `resolwebio/base`, Python modules can also be added like [this](https://github.com/genialis/resolwe-bio/blob/master/resolwe_bio/docker_images/rnaseq/packages-python3.txt). Then you can remove the last two lines from your `Dockerfile`.

mstajdohar:
Please address @mzganec's comment, then rebase. We are ready to merge your code.

PR #439, opened 2017-12-13
Add 'prepare-geo-chipseq' and 'prepare-geo-rnaseq' processes

jvrakor:
@jkokosar

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -29,6 +29,9 @@ Added
   can expect the same set of outputs on all processes of the same type.
 * Add ``requests`` Python package to ``resolwebio/utils`` Docker image
 * Add miRNA workflow
+* Add ``prepare-geo-chipseq`` and ``prepare-geo-rnaseq`` processes that
+  produce a tarball with necessary data and folder structure for GEO
+  upload
 
 Changed
 -------
--- a/resolwe_bio/processes/support_processors/prepare_geo.yml
+++ b/resolwe_bio/processes/support_processors/prepare_geo.yml
@@ -1,0 +1,267 @@
+# ===========
+# Prepare GEO
+# ===========
+---
+
+- slug: prepare-geo-chipseq
+  name: Prepare GEO - ChIP-Seq
+  requirements:
+    expression-engine: jinja
+    resources:
+      cores: 1
+    executor:
+      docker:
+        image: resolwebio/base:ubuntu-17.10
+  data_name: 'Prepare GEO ({{name}} - ChIP-Seq)'
+  version: 0.0.1
+  type: data:other:geo:chipseq
+  category: analyses
+  persistence: CACHED
+  description: >
+    Prepare ChIP-Seq data for GEO upload.
+  input:
+    - name: reads
+      label: Reads
+      type: list:data:reads:fastq
+      description: >
+        List of reads objects. Fastq files will be used.
+    - name: macs14
+      label: Macs14
+      type: list:data:chipseq:macs14
+      description: >
+        List of macs14 objects. Wiggle files will be used.
+    - name: relations
+      label: Relations
+      type: list:basic:string
+      description: >
+        Sample to background mapper. Accepted format <sample name>:<background sample name>.
+    - name: name
+      label: Collection name
+      type: basic:string
+  output:
+    - name: tarball
+      label: GEO folder
+      type: basic:file
+    - name: table
+      label: Annotation table
+      type: basic:file
+  run:
+    runtime: polyglot
+    language: bash
+    program: |
+      NAME={{ name }}_ChIP-Seq
+      mkdir "${NAME}"
+
+      {% set macs_first = macs14|first %}
+
+      declare -A SAMPLE_TO_PAIRED
+      declare -A SAMPLE_TO_BACKGROUND
+
+      {% for relation in relations %}
+        SAMPLE_TO_BACKGROUND[{{ relation.split(':')[0].split('.')[0] }}]={{ relation.split(':')[1].split('.')[0] }}
+      {% endfor %}
+
+      echo -e "File_name\tSample_name\tmd5sum\tSE/PE\tRead_length" > "${NAME}".txt
+
+      {% for read in reads %}
+        SAMPLE_NAME=`basename {{ read|sample_name }} .fastq.gz`
+        PAIRED=SE
+        MATE1_LABEL=""
+        MATE2_LABEL=""
+        {% if read|type|subtype('data:reads:fastq:paired:') %}
+          PAIRED=PE
+          MATE1_LABEL=_R1
+          MATE2_LABEL=_R2
+        {% endif %}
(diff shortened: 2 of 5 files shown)


docs/CHANGELOG.rst
```diff
@@ -10,6 +10,10 @@ This project adheres to `Semantic Versioning <http://semver.org/>`_.
 Unreleased
 ==========
 
+Added
+-----
+* Add ``prepare-geo-chipseq`` and ``prepare-geo-rnaseq`` processes
```
mstajdohar: Write a more informative note, what are this processes for? We have started using the change log to inform our users of the changes with each release so this has become more important.

mstajdohar:
@mzganec Please review.

resolwe_bio/processes/support_processors/prepare_geo.yml
```diff
@@ -0,0 +1,233 @@
...
+        if [ -n "${BACKGROUND}" ] && [ ! -e "${NAME}"/"${BACKGROUND}".wig ]; then
+          gzip -cd "`basename {{ macs.wiggle.file }} .zip`"/control/*.gz > "${NAME}"/"${BACKGROUND}".wig
+          MD5=($(md5sum "${NAME}"/"${BACKGROUND}".wig))
+          echo -e "${BACKGROUND}.wig\t${BACKGROUND}\t${MD5}\t${SAMPLE_TO_PAIRED[${BACKGROUND}]}" >> "${NAME}".txt
+        fi
+      {% endfor %}
+
+      tar -zcvf "${NAME}".tar.gz "${NAME}"
+
+      # BUILD={{ macs14[0].build }}
```
mzganec: Is build already being tracked?

resolwe_bio/processes/support_processors/prepare_geo.yml
```diff
@@ -0,0 +1,233 @@
...
+        SAMPLE_NAME=`basename {{ expression|sample_name }} .fastq.gz`
+
+        cp {{expression.exp.file}} "${NAME}"/"${SAMPLE_NAME}".tab.gz
+        MD5=($(md5sum "${NAME}/${SAMPLE_NAME}.tab.gz"))
+        echo -e ""${SAMPLE_NAME}".tab.gz\t${SAMPLE_NAME}\t${MD5}\t${SAMPLE_TO_PAIRED[${SAMPLE_NAME}]}" >> "${NAME}".txt
+      {% endfor %}
+
+      tar -zcvf "${NAME}".tar.gz "${NAME}"
+
+      EXP={{ expressions[0].exp_type }}
```
mzganec: `(expressions|first).exp_type`. But we may want to check that all expression data objects in the `expressions` list have the same expression type, source, species, build, and feature type.

resolwe_bio/processes/support_processors/prepare_geo.yml
```diff
@@ -0,0 +1,233 @@
...
+
+        {% for fastq in read.fastq %}
+          cp {{ fastq.file }} "${NAME}/${SAMPLE_NAME}${MATE1_LABEL}.fastq.gz"
+          MD5=($(md5sum "${NAME}/${SAMPLE_NAME}${MATE1_LABEL}.fastq.gz"))
+
+          zcat "${NAME}/${SAMPLE_NAME}${MATE1_LABEL}.fastq.gz" |
+          head -n 4000 |
+          awk '(NR + 2) % 4 == 0' > subsample.txt
+
+          LENGTH=`reads_length.py subsample.txt`
```
mzganec: LENGTH=`awk '{print length}' subsample.txt | sort | uniq -c | awk -F ' ' '{print $2}' | sort -r | head -n1`
But add a commentary what this does.

resolwe_bio/processes/support_processors/prepare_geo.yml
```diff
@@ -0,0 +1,267 @@
...
+  run:
+    runtime: polyglot
+    language: bash
+    program: |
+      NAME={{ name }}_ChIP-Seq
+      mkdir "${NAME}"
+
+      {% set macs_first = macs14|first %}
+
+      declare -A SAMPLE_TO_PAIRED
```
mzganec: You can do a pure Jinja implementation of `SAMPLE_TO_PAIRED`, `SAMPLE_TO_BACKGROUND`, and `NAME`.

resolwe_bio/processes/support_processors/prepare_geo.yml
```diff
@@ -0,0 +1,267 @@
...
+      declare -A SAMPLE_TO_BACKGROUND
+
+      {% for relation in relations %}
+        SAMPLE_TO_BACKGROUND[{{ relation.split(':')[0].split('.')[0] }}]={{ relation.split(':')[1].split('.')[0] }}
```
mzganec: Although Jinja doesn't support dict comprehension and our platform doesn't support the `do` Jinja extension, you can still use the last answer [here](https://stackoverflow.com/questions/36886650/how-to-add-a-new-entry-into-a-dictionary-object-while-using-jinja2). The same suggestion applies to `SAMPLE_TO_PAIRED`.

resolwe_bio/processes/support_processors/prepare_geo.yml
```diff
@@ -0,0 +1,267 @@
...
+      {% endfor %}
+
+      echo -e "File_name\tSample_name\tmd5sum\tSE/PE\tRead_length" > "${NAME}".txt
+
+      {% for read in reads %}
+        SAMPLE_NAME=`basename {{ read|sample_name }} .fastq.gz`
```
mzganec: Resolwe implements the `basename` filter [here](https://github.com/genialis/resolwe/blob/master/resolwe/flow/expression_engines/jinja/filters.py#L46).

mstajdohar:
@jvrakor Please address Matjaz's comments.

resolwe_bio/processes/support_processors/prepare_geo.yml
jvrakor: This would become: `{% set x=sample_to_background.__setitem__(relation.split(':')[0].split('.')[0], relation.split(':')[1].split('.')[0]) %}`. I don't see why this is better. @mstajdohar, any thoughts? Most of the code is in bash now. This change would require almost everything to be rewritten into jinja. Some important parts however cannot be rewritten into jinja and would still stay in bash. So we would go from bash heavy code (with some jinja) to jinja heavy code (with some bash). Is that really better and why?

resolwe_bio/processes/support_processors/prepare_geo.yml
jvrakor: This would become: 
```
{% set basename = read|sample_name|basename %}
{% set sample_name = basename.split('.')[0] %}
```

resolwe_bio/processes/support_processors/prepare_geo.yml
mzganec: This was just a suggestion. I don't object merging this PR as it is.

PR #442, opened 2017-12-19
Save original qseq multiplexed inputs

jkokosar:


jkokosar:
We do. Some processes, but no many, use this syntax.

PR #443, opened 2017-12-20
Fix Channels routing for new Resolwe

kostko:


tests/settings.py
```diff
@@ -126,26 +126,21 @@
...
+FLOW_DOCKER_VOLUME_EXTRA_OPTIONS = {
+    'data': 'Z',
+    'data_all': 'z',
+    'upload': 'z',
+    'secrets': 'Z',
+    'users': 'Z',
+    'tools': 'z',
+}
+
+FLOW_DOCKER_EXTRA_VOLUMES = []
```
dblenkus: You can omit this line.

PR #444, opened 2017-12-21
Use resolwebio/dnaseq:1.2.0 docker in filtering_chemut process

tristanbrown:
@jkokosar


resolwe_bio/processes/variant_calling/filtering_chemut.yml
```diff
@@ -9,7 +9,7 @@
     expression-engine: jinja
     executor:
       docker:
-        image: resolwebio/legacy:latest
+        image: docker.io/resolwebio/dnaseq:1.2.0
```
jkokosar: `docker.io` part is not needed. Please remove this.

PR #447, opened 2018-01-08
Expose additional parameters in STAR and cutadapt

JureZmrzlikar:


resolwe_bio/processes/alignment/star.yml
```diff
@@ -144,6 +150,58 @@
...
+          description: >
+            Output of unmapped reads in the SAM format.
+          required: false
+          default: None
+          choices:
+            - label: None
+              value: None
+            - label: Within
+              value: Within
+        - name: outSAMattributes
```
jkokosar: If `outSAMunmapped` is left to `None`, will the use of other related parameters make the process to fail?

resolwe_bio/processes/alignment/star.yml
```diff
@@ -144,6 +150,58 @@
...
+          choices:
+            - label: Local
+              value: Local
+            - label: EndToEnd
+              value: EndToEnd
+            - label: Extend5pOfRead1
+              value: Extend5pOfRead1
+            - label: Extend5pOfReads12
+              value: Extend5pOfReads12
+    - name: output_sam_bam
```
jkokosar: This `.sam` output is not saved with `re-save-file`. When saving, first check if the file was crated. Write a test.

mstajdohar:
@JureZmrzlikar Please address Janez's comments.

resolwe_bio/processes/alignment/star.yml
JureZmrzlikar: Of course it is. By STAR's defaults the core resulting file from alignment is `Aligned.out.sam`. In our process we set `--outSAMtype BAM SortedByCoordinate`, which causes the same file to be named `Aligned.sortedByCoord.out.bam`.

resolwe_bio/processes/alignment/star.yml
JureZmrzlikar: If `--outSAMunmapped` is set to None, there just won't be any unaligned reads in file `Aligned.sortedByCoord.out.bam`. I don't see how this could cause failure.

PR #457, opened 2018-01-18
Enable manager prefix overriding from the env

jberci:
Jenkins needs to be able to override the manager's redis prefix so that multiple resolwe-using test suites can run on the same machine at the same time.


mstajdohar:
@tjanez Please review.

PR #458, opened 2018-01-19
WIP resolwebio/utils Docker image

mzganec:


resolwe_bio/docker_images/utils/packages-manual/sra-toolkit.sh
```diff
@@ -0,0 +1,23 @@
...
+
+download_and_verify \
+    ncbi \
+    sra-toolkit \
+    2.8.2-1 \
+    42b2fc111f55a26be2b245b79da3b57f53f0b7ab56daa71d28eb6652371476f3 \
+    https://ftp-trace.ncbi.nlm.nih.gov/sra/sdk/current/sratoolkit.current-ubuntu64.tar.gz \
+    sratoolkit.\${version}-ubuntu64
+
+mkdir /root/.ncbi/
```
kostko: Please explain why this is in `/root`? Note that under Resolwe, we don't use the `root` user so this may be inaccessible to the process. Move this elsewhere or remove.

mzganec: SRA Toolkit looks for config file in `/root/.ncbi/user-settings.mkfg`. If this file doesn't exist the tool uses cache in `/root/ncbi/` and fails due to missing writing permissions.

The tool needs to be configured either not to use cache to save SRA files or use our mounted `data` directory which is likely the only one with writing permissions when the process is running. Note that also the second option requires changing the configuration. After discussing with @tjanez we decided to set the configuration during image preparation.

kostko: The problem is that `/root` is not readable by the user running the process. So you can't put the configuration file there. Can't you pass the configuration file location to the process when running it?

mstajdohar:
I suggest you include @JenkoB's commit from #460 so that we accept all Docker image changes in one pass.

resolwe_bio/docker_images/utils/packages-manual/sra-toolkit.sh
dblenkus: Keep in mind that SRA Toolkit doesn't use `/root/.ncbi`, but `~/.ncbi`.

mzganec:
All comments have been addressed. @kostko please do another review.

PR #461, opened 2018-01-23
Fix chemut memory limit and data type

tristanbrown:
@jkokosar @dblenkus @tjanez 
Is 32GB enough?

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -93,6 +93,13 @@ Fixed
   output files and expand ``cuffnorm`` tests
 * **BACKWARD INCOMPATIBLE:** Fix ``import-sra`` process
   to work with ``resolwebio/utils`` Docker image and refactor inputs
+* The variant-calling process in the chemical mutagenesis workflow crashed
+  because Picard Tools was trying to use more memory than was available.
+  Therefore, we have explicitly set the process to request 16 GB of RAM, and 
+  limited Picard Tools to use no more than is requested.
+* The chemical mutagenesis workflow was erroneously categorized as
+  ``data:workflow:rnaseq:cuffquant`` type. This is switched to
+  ``data:workflow:chemut`` type.
 
 
 ==================
--- a/resolwe_bio/processes/variant_calling/vc_chemut.yml
+++ b/resolwe_bio/processes/variant_calling/vc_chemut.yml
@@ -6,12 +6,14 @@
 - slug: vc-chemut
   name: Variant calling (CheMut)
   requirements:
+    resources:
+      memory: 16384
     expression-engine: jinja
     executor:
       docker:
         image: "broadinstitute/genomes-in-the-cloud:2.2.4-1469632282"
   data_name: "Called variants (CheMut)"
-  version: 1.1.0
+  version: 1.1.1
   type: data:variants:vcf:chemut
   category: analyses:variants
   persistence: CACHED
@@ -174,6 +176,7 @@
 
         echo "sorting, marking duplicates, indexing"
         java \
+          -Xmx{{ requirements.resources.memory // 1024 }}g \
           -jar /usr/gitc/picard.jar \
           MarkDuplicates \
           I={{ bam.bam.file }} \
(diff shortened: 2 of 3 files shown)


docs/CHANGELOG.rst
```diff
@@ -58,6 +58,8 @@ Changed
 Fixed
 -----
 * Fix ``consensus_subreads`` input option in Subread process
+* Fix memory limit in vc_chemut process
```
mstajdohar: Please write more informative log entries. We have started to use the change log for reporting new features to the end users. Not directly of course, but I am certain @jinlajf and @rafecooks would appreciate a more verbose list of changes. I would suggest to write something like "The chemical mutagenesis analysis crashed with the out-of-box 16 GB limit for some samples <is there a common denominator for the samples where this crashed, like for Human samples or for samples larger then...>. Therefore we have increased the memory limit to 32 GB".

resolwe_bio/processes/variant_calling/vc_chemut.yml
```diff
@@ -6,6 +6,8 @@
 - slug: vc-chemut
   name: Variant calling (CheMut)
   requirements:
+    resources:
+      memory: 32768
```
mstajdohar: Are you sure we need 32 GB? Please test locally how much RAM is really needed (if you can) and how it scale with the sample size or some other dependent variable. If you can't test it locally, ask the infrastructure team to set a memory limit for this process on the server directly. We have the capability to override memory setting. I would try 20 GB next. Maybe that's enough. It's really important to keep this requirement to the minimum as it limits us to scale the number of concurrent jobs. We usually have more CPU then RAM on the worker nodes. Also, it's worth considering if we can reduce the amount of memory in use. Try to identify the command (or script) that uses all this memory and if you can either improve the script or set the input parameters.

Finally, if you have good arguments for extending the memory limit to 32 GB do so and write them in change log and the commit message (you can write more text below the first line in the commit message).

mstajdohar:
OK, try solving this issue with the `-Xmx16g` flag and set the process' memory limit to 16 GB explicitly.

jkokosar:
@mstajdohar please review again

resolwe_bio/processes/variant_calling/vc_chemut.yml
```diff
@@ -174,6 +176,7 @@
 
         echo "sorting, marking duplicates, indexing"
         java \
+          -Xmx16g \
```
dblenkus: Use the memory limit specified in resources.

tristanbrown: Yes, isn't that what I've done here?

mstajdohar: @dblenkus, please elaborate.

mstajdohar:
@tristanbrown Were you able to test the process with the `-Xmx16g` flag? Does it work?

tristanbrown:
@mstajdohar Yes, the -Xmx16g flag works. Regarding the suggestion from @dblenkus , I'm learning some bash and testing how to construct a -Xmx#g flag from the # given by requirements.resources.memory.

dblenkus:
I think it would be easier to do it with Jinja. Something like: `-Xmx{{ (requirements.resources.memory / 1024) | round(method='floor') | int }}g`. Note that brackets around division are needed to get the right order of operations.

    In [1]: from jinja2 import Template

    In [2]: Template("""-Xmx{{ (16384 / 1024) | round(method='floor') | int }}g""").render()
    Out[2]: '-Xmx16g'

http://jinja.pocoo.org/docs/dev/templates/#round

mstajdohar:
Is the `round` needed if you pipe to `int` anyway?

dblenkus:
Yes, because it will use `common` method otherwise instead of `floor`.

tristanbrown:
@dblenkus I was able to get the same functionality using `{{x // y}}`. I've updated the pull request.

mstajdohar:
Quite clever, good job!

PR #462, opened 2018-01-24
Meta processes

JureZmrzlikar:


docs/CHANGELOG.rst
```diff
@@ -13,6 +13,9 @@ Unreleased
 Added
 -----
 * Add Ubuntu 17.10 base Docker image
+* Add processes: ``meta-alignment``, ``meta-annotation``, ``meta-expression``
```
mstajdohar: Please extend the description with more information what value does this give to the end user. For instance, that by introducing this processes, we are making sure that all processes of the same type and subprocesses have the output fields that are defined in the meta process. That this makes the web app and ReSDK scripts more robust, because they can expect the same set of outputs on all processes of the same type...

resolwe_bio/processes/alignment/meta.yml
```diff
@@ -0,0 +1,26 @@
+# =================================
+# Parent of all alignemnt processes
+# =================================
+
+- slug: meta-alignment
+  name: Meta alignment process
+  version: 1.0.0
+  type: data:alignment
```
mstajdohar: Should we add this processes to a dedicated category, say meta or abstract? This will make it easier to exclude them from the UI. @dblenkus How should we name the category?

docs/CHANGELOG.rst
JureZmrzlikar: I've updated this, let me know what you think.

resolwe_bio/processes/alignment/meta.yml
JureZmrzlikar: I would prefer abstract, but somehow meta just stuck in my head during conversations. I'm open to better suggestions...

mstajdohar: Let's call this abstract as this is more usual for this type of concepts in CS terminology. So, change the slugs to abstract- and add a category.

mstajdohar:
Why didn't you change `meta-` to `abstract-` and include a process category (`category: abstract`)?

JureZmrzlikar:
Was waiting if there are any other commentsa and adress them all at once. If this is priority, i can do it now.

mstajdohar:
No, let's wait for @jkokosar's review.

jkokosar:
I think we are missing a meta process `abstract-bed`. @mstajdohar I'm not sure if the `requirements` section in the process can be omitted?

mstajdohar:
@kostko Could we make the requirements section not required. This processes do not need to specify an engine...

kostko:
It is already not required to specify `requirements`? Did you get an error without specifying them?

JureZmrzlikar:
I've adressed your comments: renamed meta to abstract where necessary and added `abstract-bed` process. Please, review again @mstajdohar @jkokosar

docs/CHANGELOG.rst
```diff
@@ -17,6 +17,12 @@ Added
   process
 * Mechanism to override the manager's control channel prefix from the
   environment
+* Add processes ``abstract-alignment``, ``abstract-annotation``,
```
mstajdohar: I suggest: Add abstract processes ``abstract-alignment``, ``abstract-annotation``, ``abstract-expression``, ``abstract-differentialexpression`` and ``abstract-bed``. With abstract processes, we are making sure that all subtype processes have the input and output fields that are defined in the abstract process. For users, it is much more clear and predictable if they can expect the same set of outputs on all processes of the same type.

resolwe_bio/processes/differential_expression/abstract.yml
```diff
@@ -0,0 +1,43 @@
...
+# Parent of all differential expression processes
+# ===============================================
+
+- slug: abstract-differentialexpression
+  name: Abstract differential expression process
+  version: 1.0.0
+  type: data:differentialexpression
+  category: abstract
+  input:
+    - name: case
```
mstajdohar: Why do you need to specify this?

resolwe_bio/processes/alignment/abstract.yml
```diff
@@ -0,0 +1,27 @@
...
+    - name: bai
+      label: Alignment index BAI
+      type: basic:file
+    - name: species
+      label: Species
+      type: basic:string
+    - name: build
+      label: Build
+      type: basic:string
+  run:
```
mstajdohar: The run field is not required so I suggest to delete it.

mstajdohar:
Change the commit message

JureZmrzlikar:
Any update regarding failing tests, @tjanez ?

mstajdohar:
Superseded by #473.

PR #466, opened 2018-01-29
[WIP] Add BaseSpace import processes

hadalin:
Requires https://github.com/genialis/resolwe-bio/pull/475


resolwe_bio/tools/basespace_download.py
```diff
@@ -0,0 +1,112 @@
...
+        raise BaseSpaceDownloadError('Could not save file to {}, due to directory not being found'.format(file_name))
+    except PermissionError:
+        raise BaseSpaceDownloadError('Could not save file to {}, due to insufficient permissions'.format(file_name))
+
+
+if __name__ == "__main__":
+    try:
+        main()
+    except Exception as e:
+        print(e)
```
hadalin: Since this tool is used by processes requiring `secrets`, this line could be problematic.

hadalin:
Assigning @kostko, because processes require `secrets`.

resolwe_bio/processes/import_data/seq_reads.yml
```diff
@@ -218,3 +218,248 @@
...
+        BASESPACE_FILE_ARGUMENT+='--file-id {{file_id}} '
+      {% endfor %}
+
+      SECRET_PATH=/secrets/{{ access_token_secret.handle }}
+      basespace_download.py ${BASESPACE_FILE_ARGUMENT} --access-token-secret-path ${SECRET_PATH}
+      re-checkrc "Failed downloading files from BaseSpace."
+
+      gzip -dk *.gz
+
+      FASTQ=(*.fastq)
```
hadalin: Reused from https://github.com/genialis/resolwe-bio/blob/23e9deb50a8a972b544187a92b0180488b46a7b8/resolwe_bio/processes/import_data/seq_reads.yml#L48

Maybe it would be a good idea to encapsulate this functionality. cc @jkokosar

resolwe_bio/processes/import_data/seq_reads.yml
```diff
@@ -218,3 +218,248 @@
...
+    language: bash
+    program: |
+
+      function extract_filename_from_output() {
+        echo $1 | sed -nE 's/^filename=(.+)/\1/p'
+      }
+
+      SECRET_PATH=/secrets/{{ access_token_secret.handle }}
+
+      FW_FASTQ=()
```
hadalin: Reused from https://github.com/genialis/resolwe-bio/blob/23e9deb50a8a972b544187a92b0180488b46a7b8/resolwe_bio/processes/import_data/seq_reads.yml#L145

Maybe it would be good idea to encapsulate this functionality too. cc @jkokosar

resolwe_bio/tools/basespace_download.py
```diff
@@ -0,0 +1,112 @@
+#!/usr/bin/env python3
+"""Tool to download files from BaseSpace."""
+
+import sys
+import atexit
+import argparse
+import requests
+
+
+class BaseSpaceDownloadError(Exception):
```
mzganec: Populate this class to report errors.

hadalin: What do you mean? This subclass is used to distinguish errors handled in the download tool. The errors are being reported by instantiating the error class with a string in the constructor, e.g. https://github.com/genialis/resolwe-bio/pull/466/files#diff-5d9c604384eec6edf05f1ec1f946a57cR54

mzganec: You're correct, I missed something.

resolwe_bio/processes/import_data/seq_reads.yml
```diff
@@ -218,3 +218,248 @@
...
       re-save-list fastqc_url "${FW_FASTQC_URL[@]}"
       re-save-list fastqc_url2 "${RW_FASTQC_URL[@]}"
+
+- slug: basespace-fastq-single
+  name: BaseSpace Reads (single-end)
+  requirements:
+    expression-engine: jinja
+    executor:
+      docker:
+        image: resolwebio/legacy:latest
```
kostko: I am strongly against running any kind of bioinformatics tools (like fastqc) in processes which have `requirements.resources.secrets = true`. These processes should be using minimal Docker images, which can be more easily audited, without any of these tools. Any post-processing should be handled in another process and the combined flow should be encapsulated in a workflow.

I know that there are currently some missing features with workflows which make this inconvenient, but at least we should put up a big warning sign that this is currently less secure and a fixme that this must be fixed in order to improve security.

cc @tjanez @dblenkus

resolwe_bio/processes/import_data/seq_reads.yml
kostko: As mentioned in my security-related comment above, having this post-processing step as a separate process, encapsulating the two steps in a workflow would solve this duplication issue.

resolwe_bio/processes/import_data/seq_reads.yml
```diff
@@ -218,3 +218,248 @@
...
+      for fqc in "${FASTQC[@]}"
+      do
+        NAME=`basename "$fqc" _fastqc.zip`
+        FASTQC_URL+=("{\"file\":\"fastqc/${NAME}_fastqc/fastqc_report.html\",\"refs\":[\"fastqc/${NAME}_fastqc\"]}")
+      done
+
+      re-save-file-list fastqc_archive "${FASTQC[@]}"
+      re-save-list fastqc_url "${FASTQC_URL[@]}"
+
+- slug: basespace-fastq-paired
```
kostko: Big security-related warning sign as above.

Additionally, this is duplicating basespace import functionality, just because the post-processing step is different. This is another case for why splitting this into two processes and encapsulating the two steps into a workflow should be considered for the future (when we implement workflow features that will make this easier).

resolwe_bio/tools/basespace_download.py
```diff
@@ -0,0 +1,112 @@
...
+
+    for file_id in file_ids:
+        file_name = get_file_name(session, file_id, headers)
+        download_file(session, file_id, file_name, headers)
+        print('filename={}'.format(file_name))
+
+
+def on_exit(session):
+    """Clean up function called on exit."""
+    session.close()
```
kostko: Why is this needed?

resolwe_bio/processes/import_data/seq_reads.yml
```diff
@@ -218,3 +218,248 @@
...
+      RW_FASTQGZ=()
+      RW_FASTQC=()
+      RW_FASTQC_URL=()
+
+      {% for file_id in upstream_file_ids %}
+        BASESPACE_OUTPUT=`basespace_download.py --file-id {{file_id}} --access-token-secret-path ${SECRET_PATH}`
+        re-checkrc "Failed downloading files from BaseSpace."
+
+        GZ_FILE=`extract_filename_from_output ${BASESPACE_OUTPUT}`
+        gzip -dk ${GZ_FILE}
```
mzganec: Filenames can contain spaces. Consistently use quotation marks.

resolwe_bio/processes/import_data/seq_reads.yml
```diff
@@ -218,3 +218,248 @@
...
+      type: list:basic:file
+  run:
+    runtime: polyglot
+    language: bash
+    program: |
+      {% for file_id in file_ids %}
+        BASESPACE_FILE_ARGUMENT+='--file-id {{file_id}} '
+      {% endfor %}
+
+      SECRET_PATH=/secrets/{{ access_token_secret.handle }}
```
hadalin: @kostko `/secrets/{{ access_token_secret.handle }}` should rather be replaced with a template tag?

resolwe_bio/tools/basespace_download.py
kostko: Perhaps add an argument `--verbose` and only output the full backtrace (not just `print(e)`) if this argument is passed and only output `An error occurred while processing the Basespace download request. Use --verbose to see details.`.

resolwe_bio/processes/import_data/seq_reads.yml
kostko: Yes, we could add such a tag to make the location of the `/secrets` directory dependent on the executor.

resolwe_bio/tools/basespace_download.py
hadalin: Not needed per se, but since `Session` instance offers this method, it might be a good idea to call it.

resolwe_bio/processes/import_data/seq_reads.yml
hadalin: Fixed [here](https://github.com/genialis/resolwe-bio/pull/466/files#diff-e860ea00e9e0a400a0646e46faef1896R391), [here](https://github.com/genialis/resolwe-bio/pull/466/files#diff-e860ea00e9e0a400a0646e46faef1896R393), [here](https://github.com/genialis/resolwe-bio/pull/466/files#diff-e860ea00e9e0a400a0646e46faef1896R406) and [here](https://github.com/genialis/resolwe-bio/pull/466/files#diff-e860ea00e9e0a400a0646e46faef1896R408).

resolwe_bio/tools/basespace_download.py
hadalin: > only output the full backtrace

Do you mean not catching the exception or some other way?

kostko: No, use the `traceback` module to print the exception.

hadalin:
I will split the process into multiple parts, one to transfer the file from BaseSpace and one to do the import.

resolwe_bio/processes/import_data/seq_reads.yml
hadalin: It seems `resolwe/base` Docker image does not have `requests` installed.

mstajdohar: @kostko Should we extend the base image or create a new one for this process? @hadalin [This](https://github.com/genialis/resolwe-bio/tree/master/resolwe_bio/docker_images) is where we add Docker images with bioinformatics tools.

hadalin: @dblenkus recommended `resolwebio/utils` and it works.

hadalin: FYI `resolwebio/utils` has Python `2.7.14` installed.

mstajdohar:
#475 was accepted and the new `resolwebio/utils:1.1.0` image was published on Docker Hub.

hadalin:
Closing this PR because of a major rewrite it went through.

PR #475, opened 2018-02-06
Add Requests python package to resolwebio/utils Docker image

hadalin:


docs/CHANGELOG.rst
```diff
@@ -27,6 +27,7 @@ Added
   subtype processes have the input and output fields that are defined in the
   abstract process. For users, it is much more clear and predictable if they
   can expect the same set of outputs on all processes of the same type.
+* ``requests`` Python package to ``resolwebio/utils`` Docker image
```
mzganec: Prepend `Add`. You will need also `Trimmomatic` to do quality control.

mstajdohar: I think quality control is now in another process.

hadalin: >Prepend Add

Isn't this redundant information? It's under _Added_ section. Does @tjanez approve?

hadalin: >I think quality control is now in another process.

This image is used for `data:basespace`. @mzganec will update I think `rnaseq` which will be used for `basespace-fastq-*`.

mzganec:
We regularly use _Add_ in the change log, that's why I suggested it. Otherwise I don't mind.

Locally, the image builds correctly.

docs/CHANGELOG.rst
tjanez: > >Prepend Add

> Isn't this redundant information? It's under _Added_ section. Does @tjanez approve?

I agree with @mzganec:

1. It is strange to start a CHANGELOG entry with some fixed-spaced identifier such as ``requests``.

2. CHANGELOG is not a work of art, so a bit of redundancy, IMHO, can be outweighed by simplicity of using the same sentence as in the commit message.

tjanez:
My opinion added in the [comment thread](https://github.com/genialis/resolwe-bio/pull/475#discussion_r166314952).

docs/CHANGELOG.rst
hadalin: Fixed.

PR #377, opened 2017-10-05
Track organism and genome build

jkokosar:

PR #478, opened 2018-02-07
Add miRNA workflow

romunov:
* fix some typos in bowtie process
* add index files for RSEM genome meant for testing
* modify `prepare_genome` helper function which is now able to use more than the default genome
* add miRNA workflow

miRNA workflow will not build untill [this pull request](https://github.com/genialis/resolwe-bio/pull/377/files#diff-812456a06467163f6e892c958a5dd960R57) which modified some parameter names is accepted.


tjanez:
@mstajdohar, @jkokosar, what happened here?

As @mzganec rightly observed, this PR messed up the CHANGELOG. Please, prepare a fix for this!

resolwe_bio/processes/expression/featureCounts.yml
```diff
@@ -42,11 +42,11 @@
       default: gene_id
       allow_custom_choice: true
       choices:
-        - label: Gene id
+        - label: gene_id
```
dblenkus: Why have you changed this? Label should be human readable, i.e. no underscores and other special characters.

mzganec: Annotation file contains a dictionary with key-value pairs in each line where keys can be somewhat arbitrary strings and depend on the provider of the annotation file (see an example from ENSEMBL [here](https://www.ensembl.org/info/website/upload/gff.html)). The process maps the value of `ID attribute` input onto the dictionaries of the annotation file to obtain values.

If the labels in the `ID attribute` input of the process are different from the corresponding values, a user running the process from the front-end has no way of knowing which attribute is actually being selected prior to running the process. Furthermore, the user can be misled into thinking that the preferred attribute isn't available in the drop-down menu. I therefore believe that this change in the process is justified.

dblenkus: But writing `Gene ID` as `gene_id` is just a naming convention that makes things nicer more useful for computers and means the same thing otherwise.

If the problem is that users don't know which values correspond to listed labels, this is an UI problem and should be fixed there, not by changing the labels.

jkokosar: We actually solved the UI and usability problem in this case, by matching the labels to the actual values that the user needs to be aware of to correctly configure the analysis for the provided annotation file.

PR #479, opened 2018-02-07
Add BaseSpace import processes

hadalin:


tests/settings.py
```diff
@@ -145,6 +145,9 @@
 # Don't pull Docker images if set via the environment variable.
 FLOW_DOCKER_DONT_PULL = strtobool(os.environ.get('RESOLWE_DOCKER_DONT_PULL', '0'))
 
+# Disable SECCOMP if set via environment variable.
+FLOW_DOCKER_DISABLE_SECCOMP = strtobool(os.environ.get('RESOLWE_DOCKER_DISABLE_SECCOMP', '0'))
```
hadalin: cc @tjanez

resolwe_bio/tests/processes/test_upload.py
```diff
@@ -400,3 +400,68 @@ def test_upload_nucl_seq(self):
...
+        fastq_single_inputs = {'reads': basespace_import.pk}
+        reads = self.run_process('basespace-fastq-single', fastq_single_inputs)
+
+        self.assertFiles(reads, 'fastq', ['Test_S1_L001_R1_001.fastq.gz',
+                                          'Test_S1_L002_R1_001.fastq.gz'], compression='gzip')
+        del reads.output['fastqc_url'][0]['total_size']  # Non-deterministic output.
+        del reads.output['fastqc_url'][1]['total_size']  # Non-deterministic output.
+        self.assertFields(reads, 'fastqc_url', [{'file': 'fastqc/Test_S1_L001_R1_001_fastqc/fastqc_report.html',
+                                                 'refs': ['fastqc/Test_S1_L001_R1_001_fastqc'],
+                                                 'size': 309578},
```
hadalin: If `size` is changed to a random number, the tests still pass. How come?

resolwe_bio/processes/import_data/seq_reads.yml
```diff
@@ -218,3 +218,220 @@
...
       re-save-list fastqc_url "${FW_FASTQC_URL[@]}"
       re-save-list fastqc_url2 "${RW_FASTQC_URL[@]}"
+
+- slug: basespace-fastq-single
+  name: BaseSpace Reads (single-end)
+  requirements:
+    expression-engine: jinja
+    executor:
+      docker:
+        image: resolwebio/rnaseq
```
hadalin: The tests won't pass until @mzganec updates `rnaseq` Docker image to include all needed tools.

hadalin:
Old PR: https://github.com/genialis/resolwe-bio/pull/466

docs/CHANGELOG.rst
```diff
@@ -32,6 +32,10 @@ Added
 * Add ``prepare-geo-chipseq`` and ``prepare-geo-rnaseq`` processes that
   produce a tarball with necessary data and folder structure for GEO
   upload
+* BaseSpace file download tool
```
mzganec: Use _Add_ at the beginning of all four added change log entries, like in the rest of this section.

resolwe_bio/processes/import_data/basespace.yml
```diff
@@ -0,0 +1,50 @@
...
+  name: BaseSpace Files Import
+  requirements:
+    expression-engine: jinja
+    executor:
+      docker:
+        image: resolwebio/utils
+    resources:
+      network: true
+      secrets: true
+  data_name: 'BaseSpace Import ({{file_ids|length}} {% if file_ids|length == 1 %}file{% else %}files{% endif %})'
```
mzganec: Use spaces between curly brackets and the content inside. See an example from the [documentation](http://jinja.pocoo.org/docs/2.10/templates/#variables). Fix this in other places as well.

resolwe_bio/processes/import_data/basespace.yml
```diff
@@ -0,0 +1,50 @@
...
+  version: 0.0.1
+  type: data:basespace
+  category: upload
+  persistence: RAW
+  description: >
+    Import files from BaseSpace.
+  input:
+    - name: file_ids
+      label: BaseSpace file IDs
+      type: list:basic:string
```
mzganec: If file IDs are always integers the input type should be set accordingly.

resolwe_bio/processes/import_data/seq_reads.yml
```diff
@@ -218,3 +218,220 @@
...
+      type: list:basic:file:html
+    - name: fastqc_archive
+      label: Download FastQC archive
+      type: list:basic:file
+  run:
+    runtime: polyglot
+    language: bash
+    program: |
+      {% for read in reads.files %}
+        GZ_FILE_PATH={{read.file}}
```
mzganec: Redundant.

resolwe_bio/processes/import_data/seq_reads.yml
```diff
@@ -218,3 +218,220 @@
...
+        cp ${GZ_FILE_PATH} .
+      {% endfor %}
+
+      gzip -dk *.fastq.gz
```
mzganec: No need to extract reads, `FastQC` [works with compressed files](https://www.bioinformatics.babraham.ac.uk/projects/fastqc/Help/2%20Basic%20Operations/2.1%20Opening%20a%20sequence%20file.html) as well.

resolwe_bio/processes/import_data/seq_reads.yml
```diff
@@ -218,3 +218,220 @@
...
+  data_name: 'BaseSpace FASTQ paired-end {{ upstream.files.0.file }} {{ downstream.files.0.file }}'
+  version: 0.0.1
+  type: data:reads:fastq:paired
+  category: upload
+  flow_collection: sample
+  persistence: RAW
+  description: >
+    Import paired-end reads in FASTQ format from BaseSpace.
+  input:
+    - name: upstream
```
mzganec: Consider unifying input names with [upload-fastq process inputs](https://github.com/genialis/resolwe-bio/blob/master/resolwe_bio/processes/import_data/seq_reads.yml#L109), though your input names actually seem better.

resolwe_bio/processes/import_data/seq_reads.yml
```diff
@@ -218,3 +218,220 @@
...
       re-save-list fastqc_url2 "${RW_FASTQC_URL[@]}"
+
+- slug: basespace-fastq-single
+  name: BaseSpace Reads (single-end)
+  requirements:
+    expression-engine: jinja
+    executor:
+      docker:
+        image: resolwebio/rnaseq
+  data_name: 'BaseSpace FASTQ single-end {{ reads.files.0.file }}'
```
mzganec: Put the filename also in parentheses (round brackets), like in [Cutadapt](https://github.com/genialis/resolwe-bio/blob/master/resolwe_bio/processes/reads_processing/cutadapt.yml#L15).

resolwe_bio/processes/import_data/seq_reads.yml
```diff
@@ -218,3 +218,220 @@
...
+      label: Download FastQC archive (Upstream)
+      type: list:basic:file
+    - name: fastqc_archive2
+      label: Download FastQC archive (Downstream)
+      type: list:basic:file
+  run:
+    runtime: polyglot
+    language: bash
+    program: |
+      FW_FASTQ=()
```
mzganec: Try making all those lists in Jinja.

resolwe_bio/tests/processes/test_upload.py
mzganec: File size is not being checked anymore https://github.com/genialis/resolwe/blob/master/resolwe/test/testcases/process.py#L480, you can remove it.

hadalin: I guess `total_size` still needs to be [removed](https://github.com/genialis/resolwe-bio/pull/479/files#diff-c3decfeee6f2d762429890905d989817R418)?

mzganec: Yes.

hadalin:
>Consider restricting basespace-import data object to a singe BaseSpace file ID and allow a list input in basespace-fastq processes instead. Then, the process will closer resemble the two upload-fastq processes in the same YAML file.

Like the idea, but there's going to be an influx of `data:basespace` data objects. Not a problem per se, but still, @mstajdohar what do you think?

resolwe_bio/processes/import_data/basespace.yml
hadalin: BaseSpace serves file IDs as [strings](https://developer.basespace.illumina.com/docs/content/documentation/rest-api/v1-api-reference#GETfilesId), not integers, so I'd rather not change type.

resolwe_bio/processes/import_data/seq_reads.yml
hadalin: Is the `--extract ` in `fastqc "${FASTQGZ[@]}" --extract` necessary if gzipped files are on the input?

mzganec: Yes. The `--extract` flag is not related to the type of input but controls the output of the `FastQC` tool. From the [documentation](https://wiki.gacrc.uga.edu/wiki/FastQC):
```
    --extract       If set then the zipped output file will be uncompressed in
                    the same directory after it has been created.  By default
                    this option will be set if fastqc is run in non-interactive
                    mode.
```

resolwe_bio/processes/import_data/seq_reads.yml
hadalin: I guess the same goes for input name of upload-fastq-single process, naming it `src` instead of `reads`.

hadalin:
>Like the idea, but there's going to be an influx of data:basespace data objects. 

Nevermind. @mzganec's idea is reasonable. I'll go with `list`.

resolwe_bio/processes/import_data/seq_reads.yml
mzganec: Since we're running `FastQC` in non-interactive mode, it seems that the `--extract` flag is actualy redundant because it is set by default (see the last sentence in the pasted excerpt from the documentation).

hadalin: Removing `--extract` produces errors. I'd rather leave that alone and nag bioinformaticians.

resolwe_bio/processes/import_data/seq_reads.yml
mzganec: You're correct.

resolwe_bio/processes/import_data/seq_reads.yml
mzganec: https://github.com/genialis/resolwe-bio/pull/481

hadalin:
@mzganec refactored the processes like we discussed.

resolwe_bio/processes/import_data/seq_reads.yml
```diff
@@ -218,3 +218,218 @@
...
       re-save-list fastqc_url2 "${RW_FASTQC_URL[@]}"
+
+- slug: files-to-fastq-single
+  name: Convert files to Reads (single-end)
+  requirements:
+    expression-engine: jinja
+    executor:
+      docker:
+        image: resolwebio/rnaseq
+  data_name: 'Files to FASTQ single-end ({{ src.0.file.file }})'
```
mzganec: `(src|first).file.file`

resolwe_bio/processes/import_data/seq_reads.yml
```diff
@@ -218,3 +218,218 @@
...
+      type: list:basic:file
+  run:
+    runtime: polyglot
+    language: bash
+    program: |
+      {% for read in src %}
+        cp {{ read.file.file }} .
+      {% endfor %}
+
+      FASTQGZ=(*.fastq.gz)
```
mzganec: What happens if the input files are uncompressed (.fastq instead of .fastq.gz)?

resolwe_bio/processes/import_data/basespace.yml
```diff
@@ -0,0 +1,44 @@
...
+  name: BaseSpace File Import
+  requirements:
+    expression-engine: jinja
+    executor:
+      docker:
+        image: resolwebio/utils
+    resources:
+      network: true
+      secrets: true
+  data_name: 'BaseSpace File Import ({{ file_id }})'
```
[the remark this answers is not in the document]
hadalin: It's less descriptive though, but ok.

resolwe_bio/processes/import_data/seq_reads.yml
hadalin: It won't work with uncompressed ones. Do we need to support `.fastq` also?

resolwe_bio/tools/basespace_download.py
```diff
@@ -0,0 +1,122 @@
...
+    parser.add_argument('--file-id',
+                        dest='file_ids',
+                        action='append',
+                        required=True,
+                        help='BaseSpace file ID. This argument can be repeated to specify multiple files.')
+    parser.add_argument('--access-token-secret-path',
+                        dest='access_token_secret_path',
+                        required=True,
+                        help='BaseSpace access token secret path.')
+    parser.add_argument('--verbose',
```
hadalin: @mzganec `basespace_download` tool accepts `--verbose` flag. Would it make sense to add an input to [basespace-import-file](https://github.com/genialis/resolwe-bio/pull/479/files#diff-9d218a7a04cd5acd878972b5e7ece1cfR18) process?

resolwe_bio/processes/import_data/seq_reads.yml
mzganec: Let's not do this at this point.

resolwe_bio/tools/basespace_download.py
mzganec: We usually don't do this.

hadalin: The reason for this would be to able to debug in case of errors.

mzganec: I understand. You can certainly include this option but put it in the _Advanced_ group and hide it by default (see [here](https://github.com/genialis/resolwe-bio/blob/master/resolwe_bio/processes/import_data/sra_file.yml#L30)) because this option is not for basic users.

mzganec: Also verify how much output is being produced. We've had cases with more than a GB of standard output produced in verbose mode.

hadalin: >You can certainly include this option but put it in the Advanced group and hide it by default (see here) because this option is not for basic users.

Agree, but also be aware this process won't be used by a user directly.

resolwe_bio/processes/import_data/basespace.yml
```diff
@@ -0,0 +1,56 @@
+- slug: basespace-file-import
+  name: BaseSpace File Import
+  requirements:
+    expression-engine: jinja
+    executor:
+      docker:
+        image: resolwebio/utils
+    resources:
+      network: true
+      secrets: true
```
hadalin: @kostko This PR has been refactored to address concerns about `secrets`.

resolwe_bio/tools/basespace_download.py
hadalin: >Also verify how much output is being produced. We've had cases with more than a GB of standard output produced in verbose mode.

It's just a stack trace.

resolwe_bio/processes/import_data/basespace.yml
```diff
@@ -0,0 +1,55 @@
+- slug: basespace-file-import
+  name: BaseSpace File Import
+  requirements:
+    expression-engine: jinja
+    executor:
+      docker:
+        image: resolwebio/utils
```
mzganec: Tag a specific version of `resolwebio/utils` Docker image.

resolwe_bio/processes/import_data/seq_reads.yml
mzganec: Tag a specific version of `resolwebio/rnaseq` Docker image. Do the same for the paired-end process.

resolwe_bio/processes/import_data/seq_reads.yml
```diff
@@ -218,3 +218,218 @@
       re-save-file-list fastqc_archive2 "${RW_FASTQC[@]}"
       re-save-list fastqc_url "${FW_FASTQC_URL[@]}"
       re-save-list fastqc_url2 "${RW_FASTQC_URL[@]}"
+
+- slug: files-to-fastq-single
+  name: Convert files to Reads (single-end)
```
mzganec: Reads -> reads.

hadalin: Fixed.

resolwe_bio/processes/import_data/seq_reads.yml
```diff
@@ -218,3 +218,218 @@
...
+
+      for name in "${RW_FASTQ[@]}"; do
+        READS_NAME_RW=`basename "$name" .fastq`
+        ENCODING=`parse_encoding_type.py "./fastqc/${READS_NAME_RW}_fastqc/fastqc_data.txt"`
+
+        if [ "${ENCODING}" == "Illumina 1.5" ] ||  [ "${ENCODING}" == "Illumina 1.3" ]; then
+
+          echo "Recoding input reads from Phred64 encoding to Phred33 encoding."
+          mv "${READS_NAME_RW}.fastq" input_reads.fastq
+          TrimmomaticSE input_reads.fastq reformated.fastq.gz TOPHRED33
```
JenkoB: Calling Trimmomatic with this command is not working on this docker image. It is working with ```java -jar /opt/usadellab/trimmomatic/trimmomatic-0.36.jar SE```, we should probably change the way trimmomatic is installed in `resolwebio/rnaseq:2.1.0` docker image

mzganec: Do not use the full path to call `trimmomatic`. I created [a script](https://github.com/genialis/resolwe-bio/blob/master/resolwe_bio/docker_images/rnaseq/packages-manual/trimmomatic.sh#L13) `trimmomatich.sh`  which is available on `$PATH` and we should use that instead.

hadalin: Please provide me a code snippet so I can update this like. We also need to update `upload-fastq-single`, etc.

mzganec: In all three cases just replace `TrimmomaticSE` with `trimmomatic.sh SE`. The rest of the parameters remain the same.

resolwe_bio/processes/import_data/seq_reads.yml
```diff
@@ -218,3 +218,218 @@
...
       re-save-list fastqc_url "${FW_FASTQC_URL[@]}"
       re-save-list fastqc_url2 "${RW_FASTQC_URL[@]}"
+
+- slug: files-to-fastq-single
+  name: Convert files to reads (single-end)
+  requirements:
+    expression-engine: jinja
+    executor:
+      docker:
+        image: resolwebio/rnaseq:2.1.0
```
mzganec: Use `resolwebio/rnaseq:2.1.1` Docker image which has `Trimmomatic` installed. The same goes for the paired-end process.

hadalin: Fixed.

mzganec:
This PR looks good already. Put https://github.com/genialis/resolwe-bio/pull/479/commits/b7989df63d5472ae71f74f24f1dabb48f40b52f8 earlier in the commit order though.

hadalin:
>Put b7989df earlier in the commit order though.

b7989df is the first one. Note that commit order on GitHub does not always reflect actual commit order, especially when using rebase.

resolwe_bio/processes/import_data/seq_reads.yml
hadalin: Does this still apply?

mzganec: No. It's correct as you have it now. https://github.com/genialis/resolwe-bio/blob/master/resolwe_bio/docker_images/rnaseq/packages-manual/trimmomatic.sh#L13

hadalin:
Since this PR deals with [secrets](https://github.com/genialis/resolwe-bio/pull/479/files#diff-9d218a7a04cd5acd878972b5e7ece1cfR10), it should be approved by @kostko too.

resolwe_bio/processes/import_data/basespace.yml
```diff
@@ -0,0 +1,55 @@
...
+      {% set SECRET_PATH = '/secrets/' ~ access_token_secret.handle %}
+
+      BASESPACE_OUTPUT=$(basespace_download.py \
+        --file-id {{ file_id }} \
+        --access-token-secret-path {{ SECRET_PATH }} \
+        {% if verbose %} --verbose {% endif %} \
+      )
+      re-checkrc "Failed downloading files from BaseSpace."
+
+      function extract_filename_from_output() {
```
kostko: Why can't `basespace_download.py` also handle this? It seems strange to me that you need to do post-processing in bash for a script that you provide.

resolwe_bio/processes/import_data/basespace.yml
```diff
@@ -0,0 +1,55 @@
...
+  run:
+    runtime: polyglot
+    language: bash
+    program: |
+      {% set SECRET_PATH = '/secrets/' ~ access_token_secret.handle %}
+
+      BASESPACE_OUTPUT=$(basespace_download.py \
+        --file-id {{ file_id }} \
+        --access-token-secret-path {{ SECRET_PATH }} \
+        {% if verbose %} --verbose {% endif %} \
```
kostko: This seems wrong to me. Making verbose easily available like this via a parameter could expose secrets. Verbose should only be used for local script debugging and should not be available in production.

kostko: Another way would be to limit the use of the `verbose` parameter to users with specific privileges. Of course this is not currently supported in Resolwe.

resolwe_bio/processes/import_data/basespace.yml
kostko: Perhaps this could be an output format parameter, like `--output filename` would output just the filename while `--output full` would output everything as it is now.

hadalin: This is used to obtain the filename which is needed in `re-save-file`.

Is there any other way of extracting "return" values from python scripts?

hadalin: This wouldn't be a problem if processes were written in Python.

kostko: > This wouldn't be a problem if processes were written in Python.

Sure, but until they are I think the `--output` flag is not that bad.

resolwe_bio/processes/import_data/basespace.yml
hadalin: The only way is then to remove `verbose` input I guess, but that way we lose the ability to debug in production.

resolwe_bio/processes/import_data/basespace.yml
hadalin: The script can take multiple files on the input and prints downloaded filenames in the form of `filename=foo.fastq.gz`, so by `--output filename` you mean only `filename=...` would be printed?

The thing is that `print('filename=` is the only print statement, unless an error happens in which case filename is not used. Therefore having `--output full` does the same thing as `--output filename`.

kostko: Output with `--output full`:
```
filename=bar
somethingelse=foo
```

Output with `--output filename`:
```
bar
```

The point is that you make the output such that you don't need any post-processing.

kostko: BTW, if the filename is the only thing printed, why doesn't the process just output filenames separated by newlines instead of printing them in a `key=value` format?

hadalin: To make it more robust if anything changes in the script (e.g. someone adds print statement and breaks processes). I like the idea of `key=value` to be honest and don't mind doing a little post processing.

resolwe_bio/processes/import_data/basespace.yml
hadalin: Removed `verbose` from the inputs.

resolwe_bio/processes/import_data/basespace.yml
kostko: In this case, I think that supporting a flag like I mentioned above should be the right way. E.g. by default or if `--output full` is specified, you keep the `key=value` format, but if one specifies `--output filename` (or other keys in the future) only the values for that key are printed. In this way, you don't need any post-processing.

hadalin: Fixed.

resolwe_bio/tools/basespace_download.py
```diff
@@ -0,0 +1,148 @@
...
+    and can be printed without using this function.
+
+    """
+    if output_option == 'full':
+        print(value)
+    elif output_option == 'filename':
+        if value.startswith('filename='):
+            print(value[len('filename='):])
+    else:
+        print("Internal error: output argument {} handling not implemented".format(output_option))
```
kostko: This should be an exception or at least be followed by an exit.

resolwe_bio/tools/basespace_download.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+def make_get_request(session, url, headers, stream=False):
+    """Make a get request."""
+    response = session.get(url, headers=headers, stream=stream)
+
+    if response.status_code == 401:
+        raise BaseSpaceDownloadError('Authentication failed on URL {}'.format(url))
+
+    if response.status_code == 404:
```
kostko: `elif`

resolwe_bio/tools/basespace_download.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+def get_basespace_api_url():
+    """Get base BaseSpace API URL."""
+    return 'https://api.basespace.illumina.com/v1pre3'
+
+
+def get_basespace_api_file_url(file_id):
+    """Get BaseSpace API file URL."""
+    return get_basespace_api_url() + '/files/' + file_id
```
kostko: Use `.format` instead of concatenating strings using the `+` operator.

resolwe_bio/tools/basespace_download.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+def get_basespace_api_file_content_url(file_id):
+    """Get BaseSpace API file contents URL."""
+    return get_basespace_api_file_url(file_id) + '/content'
```
kostko: Use `.format` instead of concatenating strings using the `+` operator.

hadalin: Fixed all.

PR #482, opened 2018-02-12
Add Salmon to resolwebio/rnaseq docker image

jkokosar:


resolwe_bio/docker_images/rnaseq/packages-debian.txt
```diff
@@ -1,5 +1,6 @@
 python-pip
 r-base-core
+salmon
```
mzganec: It would be better to install the latest version of Salmon manually. The version installed here is more than one year old. The more recent versions contain [improvements and bug fixes](https://github.com/COMBINE-lab/salmon/releases) in the indexing process which is to be performed in https://github.com/genialis/resolwe-bio/pull/483. In two months the Ubuntu LTS release is coming out with the same old packaged Salmon version (see [here](https://packages.ubuntu.com/bionic/salmon)).

jkokosar: Noted, thank you. I was aware of newer versions of the tool, but decided not to overcomplicate things at this point, when this tool in used with the limited set of functionality and for the feature still in experimental phase. I suggest to upgrade the tool version when we decide to integrate this for the support of RNA-seq workflows.

PR #453, opened 2018-01-16
Add DEXSeq and Subread_to_DEXSeq to resolwebio/rnaseq

jvrakor:
@jkokosar @mzganec 

This image is built on docker.io/resolwebio/base:ubuntu-18.04 which is build on docker.io/resolwe/base:ubuntu-18.04. These two docker images do not yet exist. Both are built on Ubuntu 18.04.

PR #489, opened 2018-02-15
Change base Docker image of resolwebio/rnaseq to resolwebio/base:ubuntu-18.04

mzganec:
After transition to the `resolwebio/base` Docker image Bioconductor packages stopped working. This is fixed by installing system Ubuntu packages of the affected Bioconductor libraries.

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -100,6 +100,8 @@ Changed
   ``resolwebio/chipseq`` docker image
 * Make ID attribute labels in ``featureCounts`` more informative
 * Change 'source' to 'gene ID database' in labes and descriptions
+* Change base Docker image of ``resolwebio/rnaseq`` to
+  ``resolwebio/base:ubuntu-18.04``
 
 Fixed
 -----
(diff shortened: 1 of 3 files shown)


mzganec:
The same version of `affy` is installed as before due to an earlier mislabeling in the readme file. `tximport` doesn't have a Ubuntu package and works correctly if installed as before. After the upcoming upgrade of the Docker image to Ubuntu 18.04 the installed Bioconductor packages will be automatically upgraded.

mzganec:
We basically have three options:

1. use this PR with Ubuntu 17.10 or 18.04 base image
2. install R with a fix for Bioconductor from source using Ubuntu 17.10 or 18.04 base image and discard this PR
3. transition with `resolwebio/rnaseq` image to Ubuntu 18.04 base and discard this PR

Option 1 downgrades `DESeq2` and `rtracklayer` with Ubuntu 17.10. Option 2 may not be the best idea because we want to revert back to the Ubuntu system R package when transitioning to LTS. I prefer option 3. It comes with the least overhead and I've tested it substantially. @jvrakor is also waiting for Ubuntu 18.04 with https://github.com/genialis/resolwe-bio/pull/453.

mzganec:
We decided for the first option transitioning to `resolwebio/base:ubuntu-18.04` base Docker image. @kostko please review.

PR #493, opened 2018-02-21
Fix custom Trimmomatic script in resolwebio/rnaseq Docker image

mzganec:
Fix passing parameters from the custom script forward to the `Trimmomatic` tool. The `$@` in the custom script were not escaped and didn't survive the script creation process. 

In this PR two custom scripts are created (one for single and one for paired-end reads) to mimic the `trimmomatic` system package in Ubuntu. This allows calling `Trimmomatic` in the same manner as [before](https://github.com/genialis/resolwe-bio/blob/master/resolwe_bio/processes/import_data/seq_reads.yml#L68).

### DIFF
--- a/resolwe_bio/docker_images/rnaseq/packages-manual/trimmomatic.sh
+++ b/resolwe_bio/docker_images/rnaseq/packages-manual/trimmomatic.sh
@@ -10,12 +10,19 @@ download_and_verify \
     http://www.usadellab.org/cms/uploads/supplementary/Trimmomatic/Trimmomatic-\${version}.zip \
     Trimmomatic-\${version}
 
-cat >trimmomatic.sh <<EOL
+cat <<'EOF' >TrimmomaticSE
 #!/bin/bash
 
-java -jar /opt/usadellab/trimmomatic/trimmomatic-0.36.jar "$@"
-EOL
-chmod +x trimmomatic.sh
+java -jar /opt/usadellab/trimmomatic/trimmomatic-0.36.jar SE "$@"
+EOF
+chmod +x TrimmomaticSE
+
+cat <<'EOF' >TrimmomaticPE
+#!/bin/bash
+
+java -jar /opt/usadellab/trimmomatic/trimmomatic-0.36.jar PE "$@"
+EOF
+chmod +x TrimmomaticPE
 
 add_binary_path \
     usadellab \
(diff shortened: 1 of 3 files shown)


resolwe_bio/docker_images/rnaseq/packages-manual/trimmomatic.sh
```diff
@@ -10,12 +10,19 @@ download_and_verify \
     http://www.usadellab.org/cms/uploads/supplementary/Trimmomatic/Trimmomatic-\${version}.zip \
     Trimmomatic-\${version}
 
-cat >trimmomatic.sh <<EOL
+cat >trimmomaticSE <<EOL
```
tjanez: The _here_ documents in Bash that I've seen usually use `EOF` as the limit string and reverse the arguments. Something like:

```
cat << EOF > /tmp/yourfilehere
These contents will be written to the file.
        This line is indented.
EOF
```

resolwe_bio/docker_images/rnaseq/packages-manual/trimmomatic.sh
```diff
@@ -10,12 +10,19 @@ download_and_verify \
...
 #!/bin/bash
 
-java -jar /opt/usadellab/trimmomatic/trimmomatic-0.36.jar "$@"
+java -jar /opt/usadellab/trimmomatic/trimmomatic-0.36.jar SE "\$@"
```
tjanez: You could avoid this and use `'EOF'` to disable interpretation of variables in the whole _here_ document. See [this StackOverflow answer](https://stackoverflow.com/a/2954835/1872673) for more details.

resolwe_bio/docker_images/rnaseq/packages-manual/trimmomatic.sh
```diff
@@ -10,12 +10,19 @@ download_and_verify \
     http://www.usadellab.org/cms/uploads/supplementary/Trimmomatic/Trimmomatic-\${version}.zip \
     Trimmomatic-\${version}
 
-cat >trimmomatic.sh <<EOL
+cat <<EOF >trimmomaticSE
```
JenkoB: To keep consistent we should name it `TrimmomaticSE` and `TrimmomaticPE`. As far as I tried it, it is still not working with input parameters?

PR #497, opened 2018-02-28
Expose advanced parameters in chemut workflow

tristanbrown:
@jkokosar 
chemparams_collapsed
chemparams_vc
chemparams_vf

### DIFF
--- a/resolwe_bio/processes/workflows/chemut.yml
+++ b/resolwe_bio/processes/workflows/chemut.yml
@@ -4,14 +4,14 @@
   data_name: 'CheMut'
   requirements:
     expression-engine: jinja
-  version: 0.0.2
+  version: 0.0.3
   type: data:workflow:chemut
   input:
     - name: analysis_type
       label: Analysis type
       type: basic:string
       description: >
-        Choice of the analysis type. Choose options SNV_CHR2 or INDEL_CHR2 if duplication of CHR2 is to be considered as diploidy when running GATK analysis (-ploidy 2 -L chr2:2263132-3015703).
+        Choice of the analysis type. Choose options SNV_CHR2 or INDEL_CHR2 to run the GATK analysis only on the diploid portion of CHR2 (-ploidy 2 -L chr2:2263132-3015703).
       choices:
         - label: SNV
           value: snv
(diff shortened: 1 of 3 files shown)


resolwe_bio/processes/workflows/chemut.yml
```diff
@@ -22,15 +22,86 @@
...
+            rsIDs from this file are used to populate the ID column of the output. Also, the DB INFO flag will be set when appropriate. dbSNP is not used in any way for the calculations themselves.
+        - name: known_sites
+          label: Known sites (dbSNP)
+          type: data:variants:vcf
+          required: false
+          hidden: "Vc.advanced === false || Vc.br_and_ind_ra === false && Vc.dbsnp === false"
+        - name: known_indels
+          label: Known indels
+          type: list:data:variants:vcf
+          default: []
```
jkokosar: Remove this. The default cannot be set for a `data` type of inputs

resolwe_bio/processes/workflows/chemut.yml
```diff
@@ -22,15 +22,86 @@
...
+      group:
+        - name: advanced
+          label: Advanced options
+          type: basic:boolean
+          default: false
+          required: false
+        - name: br_and_ind_ra
+          label: Do variant base recalibration and indel realignment
+          type: basic:boolean
+          default: false
```
jkokosar: If the `default` value is set and cannot be "removed" by user (boolean field is either `true` or `false`) the `required` field does not need to be set to `false`, since it is redundant.

tristanbrown: In the GUI, anything without a `required: false` tag is marked with an asterisk (defaulting to `required: true`). I've found this confusing when running processes with boolean checkboxes in the GUI, because it seems to indicate that the box needs to be checked.

resolwe_bio/processes/workflows/chemut.yml
tristanbrown: When I remove that value, the variant calling process fails:
```=======================================
Debugging information for data object 6
=======================================

Process' errors:
----------------
Internal error: 'NoneType' object is not iterable```
The issue seems to be `known_indels` evaluating to `True` [here](https://github.com/genialis/resolwe-bio/blob/master/resolwe_bio/processes/variant_calling/vc_chemut.yml#L235) as a result of [this change](https://github.com/genialis/resolwe-bio/pull/497/files#diff-1407ec7872daee7107b79174d3a7b37cR117). 

How can I prevent this?

jkokosar: Ok, since this field is not required, it seems that the only way to overcome this issue is to include the `default: []`

resolwe_bio/processes/workflows/chemut.yml
jkokosar: OK, good point!

PR #499, opened 2018-02-28
Fix cuffnorm to work on sample with dashes in their names and add error messages

jvrakor:
@jkokosar @JenkoB @mstajdohar @dblenkus


resolwe_bio/tools/cuffnorm_group_labels.py
```diff
@@ -1,16 +1,15 @@
...
 """Create group labels from replicates."""
-from __future__ import absolute_import, division, print_function
 
 import argparse
 
 parser = argparse.ArgumentParser(description="Create group labels")
 parser.add_argument('replicates', help="Define replicates and groups")
 args = parser.parse_args()
 
-replicates = set(args.replicates.split(","))
```
dblenkus: Is it ok to sort `replicates` here and not [here](https://github.com/genialis/resolwe-bio/pull/499/commits/5ea34fa6c678bc1070faa165f566847913eea52e#diff-e00ba752209ec3fee45885177c8237f9R15)?

resolwe_bio/tests/files/cuffnorm_all_fpkm_means.txt
```diff
@@ -1,5 +1,5 @@
-Group1	Group3	Group2
-CSNK2A1	114564.5	118186	113357.333333333
-RBCK1	117094.1	140182	109398.133333333
-TBC1D20	41356.25	36121.1	43101.3
-TRIB3	121871.75	89448.5	132679.5
+Group1	Group2	Group3
```
dblenkus: Don't change the order it's an extra test to handle such case.

resolwe_bio/tools/cuffnorm_group_labels.py
jvrakor: Yes. Here they must be sorted because of the 'set', otherwise the group outputs might differ every time the process is ran. In the second file it is correct not to sort them since they must have the same order as other inputs (samples, files) to the process.

jvrakor: Yes. Here they must be sorted because of the 'set', otherwise the group outputs might differ every time the process is ran. In the second file it is correct not to sort them since they must have the same order as other inputs (samples, files) to the process.

resolwe_bio/tests/files/cuffnorm_all_fpkm_means.txt
jvrakor: Before `replicates = set(args.replicates.split(","))` was changed to `replicates = sorted(list(set(args.replicates.split(","))))` this order was different every time the test ran. Now it is the same every time.

jkokosar:
@jvrakor Please rebase

PR #500, opened 2018-03-01
Add DEXSeq related processes

jvrakor:
@jkokosar @mzganec and @romunov for R:)


jvrakor:
The dexseq process tests fail because of a hash mismatch. This is interesting because similar differential expression processes produce similar '.tab.gz' files and there the tests pass. Locally the tests also pass. Any thoughts on that?

mzganec:
The file `diffexp_dexseq.tab.gz` contains floats. Floating-point numbers are not guaranteed to be returned in exactly the same form on different systems. Specifically, the least significant figures can differ. To test process output field values we use [`assertAlmostEqualGeneric`](https://github.com/genialis/resolwe/blob/master/resolwe/test/testcases/__init__.py#L93). However, we don't have an equivalent solution developed for files yet. In such cases we do a weaker test with [`assertFileExists`](https://github.com/genialis/resolwe/blob/master/resolwe/test/testcases/process.py#L630).

resolwe_bio/tests/processes/test_diff_expression.py
```diff
@@ -148,3 +148,121 @@ def test_edger(self):
...
+"""
+                }
+            )
+
+            inputs = {
+                'fc': 'mus_musculus_short_fCount1.txt.gz',
+                'source': 'ENSEMBL',
+                'species': 'Mus musculus',
+                'build': 'GRCm38.p5',
+                'feature_type': 'exon'
```
romunov: Missing comma at the end.

resolwe_bio/tests/processes/test_support_processors.py
```diff
@@ -216,3 +216,18 @@ def test_library_strandedness(self):
...
         self.assertFields(lib_strandedness_paired, 'fragment_ratio', 1.0)
+
+    @tag_process('prepare-annotation-dexseq')
+    def test_prepare_dexseq_gtf(self):
+        with self.preparation_stage():
+            inputs = {
+                'src': 'annotation_ensembl_mus_musculus_short.gtf.gz',
+                'source': 'ENSEMBL',
+                'species': 'Mus musculus',
+                'build': 'GRCm38.p5'
```
romunov: Missing comma at the end.

resolwe_bio/tools/plot_dexseq.R
```diff
@@ -0,0 +1,39 @@
...
+parser = ArgumentParser(description='Run DEXSeq differential expression')
+parser$add_argument('--dxr', help='DEXSeq resuts R object', required=TRUE)
+parser$add_argument('--gene', help='Gene name', required=TRUE)
+parser$add_argument('-t', help='Plot transcript models', action="store_true")
+parser$add_argument('-s', help='Plot individual samples', action="store_true")
+parser$add_argument('-e', help='Plot exon expression', action="store_true")
+
+args = parser$parse_args()
+
+print(str(args))
```
romunov: Not needed any more, remove.

resolwe_bio/processes/plots/dexseq_plot.yml
```diff
@@ -0,0 +1,75 @@
...
+      {% endif %}
+
+      plot_dexseq.R \
+        --dxr {{ dexseq.dxr.file }} \
+        --gene {{ gene }} \
+        ${FLAGS}
+
+      re-checkrc "Error plotting differential expression data (DEXSeq)."
+
+      re-save-file pdf Rplots.pdf
```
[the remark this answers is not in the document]
jvrakor: This process was envisioned to be used in visualizations and not as a stand-alone tool. That's why I didn't put too much effort in the name conventions and wanted to keep it short and not add extra lines to the process (like renaming the file) or the R script. But of course it could be done.

resolwe_bio/processes/support_processors/prepare_dexseq_annotation.yml
```diff
@@ -0,0 +1,55 @@
...
+        re-error "Annotation file must be from ENSEMBL."
+      {% endif %}
+
+      NAME=`basename {{ annotation.annot.file }} .gtf`
+
+      dexseq_prepare_annotation2 -f "${NAME}_dexseq.gtf" {{ annotation.annot.file }} "${NAME}_dexseq.gff"
+
+      re-save-file annot "${NAME}_dexseq.gtf"
+      re-save source ENSEMBL
+      re-save species {{annotation.species}}
```
mzganec: `{{ annotation.species }}`

resolwe_bio/tests/processes/test_diff_expression.py
```diff
@@ -148,3 +148,121 @@ def test_edger(self):
...
+                        'docker': {
+                            'image': 'resolwebio/base:ubuntu-18.04',
+                        },
+                    },
+                },
+                contributor=self.contributor,
+                data_name='{{ fc.file|default("?") }}',
+                type='data:expression:featurecounts',
+                flow_collection="sample",
+                input_schema=[
```
mzganec: The only input that is needed is `fc`. You can hardcode the rest in the `run` section of the process.

resolwe_bio/tests/processes/test_diff_expression.py
```diff
@@ -148,3 +148,121 @@ def test_edger(self):
         self.assertFields(diff_exp, 'species', 'Dictyostelium discoideum')
         self.assertFields(diff_exp, 'build', 'dd-05-2009')
         self.assertFields(diff_exp, 'feature_type', 'gene')
+
+    @tag_process('differentialexpression-dexseq')
+    def test_dexseq(self):
+        with self.preparation_stage():
+            # Mock upload featureCount process
+            process = Process.objects.create(
+                name='Upload featureCount mock process',
```
mzganec: It's _featureCounts_.

resolwe_bio/processes/differential_expression/dexseq.yml
```diff
@@ -0,0 +1,134 @@
...
+  version: 0.0.1
+  type: data:differentialexpression:dexseq
+  category: analyses
+  persistence: CACHED
+  description: >
+    Find differentially expressed exons.
+  input:
+    - name: case
+      label: Case
+      type: list:data:expression
```
mzganec: The type of expression data object is inconsistent with the run section of the process. If only expressions coming from _featureCounts_ are supported the input type should be set more specifically.

resolwe_bio/tests/processes/test_support_processors.py
```diff
@@ -216,3 +216,18 @@ def test_library_strandedness(self):
...
+                'src': 'annotation_ensembl_mus_musculus_short.gtf.gz',
+                'source': 'ENSEMBL',
+                'species': 'Mus musculus',
+                'build': 'GRCm38.p5',
+            }
+            annotation = self.run_process('upload-gtf', inputs)
+
+        inputs = {'annotation': annotation.pk}
+        prepare_dexseq = self.run_process('prepare-annotation-dexseq', inputs)
+        self.assertFileExists(prepare_dexseq, 'annot')
```
mzganec: Why isn't hash checked here? Annotation files don't contain floats.

resolwe_bio/processes/differential_expression/dexseq.yml
```diff
@@ -0,0 +1,134 @@
...
+      {% for e in expression %}
+
+        {% if e.source != (expressions|first).source %}
+          re-warning "All expression data must be annotated by the same genome database."
+          re-error "Expression {{ expressions|first|name }} has {{ (expressions|first).source }} gene IDs, while {{ e|name }} has {{ e.source }} gene IDs."
+        {% endif %}
+
+        {% if e.species != (expressions|first).species %}
+          re-warning "All expressions must be of the same species."
+          re-error "Expression {{ expressions|first|name }} is {{ (expressions|first).species }}, while {{ e|name }} is {{ e.species }}."
```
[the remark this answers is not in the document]
romunov: Why are we issuing a warning and then throwing an error?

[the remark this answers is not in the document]
jvrakor: @romunov  this was a part of the error messages rehaul that @JenkoB made so she is probably best suited to answer this question.

@mzganec I agree this is not the best way. However this does not return sample names but data object names. But yes, these are often the same. I'll look into it and come back with something better. We should have a discussion on rehauling error messages again and come up with a good practice that we can then use everywhere.

resolwe_bio/tests/processes/test_support_processors.py
jvrakor: The third-party script that is used to produce the annotation file returns a different result every time. So the output annotation is different every time. That does not affect further analysis however as what is different is only the order of the gene names in a single row (separated by '+') and not the order of rows.

resolwe_bio/processes/differential_expression/dexseq.yml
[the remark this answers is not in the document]
rafecooks: An easy semantic fix:
re-warning: "All **samples** must be the same species."
re-errorr: "The samples `Sample1_name` is from `organism1` while `Sample2_name` is from `organism2`.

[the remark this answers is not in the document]
JenkoB: I agree with Rafes suggestion, and it would be good to correct it in other processes. The decision of giving Warning and Error is because it displayed better having it that way than error in more lines... I am not committed to this decisions, so I support all improvements. :)

resolwe_bio/processes/differential_expression/dexseq.yml
```diff
@@ -0,0 +1,132 @@
...
+- slug: differentialexpression-dexseq
+  name: DEXSeq
+  requirements:
+    expression-engine: jinja
+    executor:
+      docker:
+        image: resolwebio/rnaseq:2.1.1
+  data_name: 'Differential expression (case vs. control)'
+  version: 0.0.1
+  type: data:differentialexpression:dexseq
```
jkokosar: We have a [meta process](data:differentialexpression) of `data:differentialexpression` type that defines the required fields for the `data:differentialexpression:*` processes. Here, we are missing the `de_json` and `de_file` fields. @JureZmrzlikar how come this was not reported by the failing tests? Were the tests implemented just for migrations? 

Are the `de_json` and `de_file` fields even relevant for this analysis? Is the output of DEX-seq viewable in Volcano plot?

resolwe_bio/processes/plots/dexseq_plot.yml
```diff
@@ -0,0 +1,75 @@
+# =============
+# Plot - DEXSeq
+# =============
+---
+
+- slug: plot-dexeq
```
jkokosar: What is a typical running time for this process? Since its to be used for visualisations, I think its class should be set to `scheduling_class: interactive` and `persistence: TEMP`

resolwe_bio/processes/plots/dexseq_plot.yml
```diff
@@ -0,0 +1,75 @@
...
+  name: DEXSeq - Plot
+  requirements:
+    expression-engine: jinja
+    executor:
+      docker:
+        image: resolwebio/rnaseq:2.1.1
+  data_name: 'Plot - DEXSeq (case vs. control)'
```
jkokosar: Include the `gene` input variable in the object name to make it more informative in BD.

resolwe_bio/processes/plots/dexseq_plot.yml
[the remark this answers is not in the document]
jkokosar: I suggest to make this into `dexseq-plot-<gene_id>.pdf`. This kind of name is far better for the visualization export functionality. The analysis works only with ENSEMBL-based annotations, so there is no worry about the possible special characters e.g. `:` in the file name.

resolwe_bio/processes/support_processors/prepare_dexseq_annotation.yml
```diff
@@ -0,0 +1,55 @@
...
+    runtime: polyglot
+    language: bash
+    program: |
+      {% if annotation.source != "ENSEMBL" %}
+        re-error "Annotation file must be from ENSEMBL."
+      {% endif %}
+
+      NAME=`basename {{ annotation.annot.file }} .gtf`
+
+      dexseq_prepare_annotation2 -f "${NAME}_dexseq.gtf" {{ annotation.annot.file }} "${NAME}_dexseq.gff"
```
jkokosar: Add `re-checkrc` here

resolwe_bio/processes/differential_expression/dexseq.yml
jvrakor: For these outputs we need to parse the dexseq output table. We have several options how to do that:

1.) We change the https://github.com/genialis/resolwe-bio/blob/master/resolwe_bio/tools/parse_diffexp.py tool to run on python3:

1.1.) That would require changing some of the differential expression's docker images to `resolwebio/rnaseq:2.1.1` as it has numpy and pandas installed in python3. For some processes (edgeR, limma) we would have to create a new version of rnaseq docker image because they only live in the legacy image which has python2 numpy and pandas. I prefer this option as we will have to change processes in the future anyway. The only problem is that Ubuntu 18.04 is not "stable" yet.

1.2.) Or we run the script from the older processes with python2.

2.) We keep the parse_diffexp.py script as it is and run it from dexseq process with python3.

resolwe_bio/tools/parse_diffexp.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python3
 """Parse Diff Exp output files."""
 from __future__ import absolute_import, division, print_function
```
jkokosar: This imports are not required

PR #501, opened 2018-03-01
Fix replacing NA values in DE results table

jkokosar:


mzganec:
Although tempting, the proposed fix is not correct. N/As in the `DESeq2` output arise from the test for outliers based on the Cook's distance. Think of N/A as a gene flag that indicates a likely outlier in one or more samples. The corresponding p-values of such genes are not 1. See also the discussion on [seqanswers.com](http://seqanswers.com/forums/showthread.php?t=45526).

Instead I propose disabling detection of outliers in [deseq.R](https://github.com/genialis/resolwe-bio/blob/master/resolwe_bio/tools/deseq.R#L38) with
`result <- results(dds, cooksCutoff=FALSE)`.

If needed, we can display position of genes with outliers in volcano plot with a different point style.

jkokosar:
I think the DESeq2 process should be updated as suggested in _addition_ to the fix proposed in this PR. One might upload NA-containing data table using the upload DE process, and this table would again contain NA values, which should be converted to 1 instead of 0.

mzganec:
You raised a good question. In the upload process we need to deal with missing _p_-values. My point is that the N/A value does not imply _p_ = 1. In fact, if you set `cooksCutoff=FALSE`, you'll get p < 1 for the majority of such genes. _p_ values near 1 mean that expressions of a gene in case and control group are very similar in relation to the error size. On the other hand, _p_ = N/A implies that expressions of a gene across samples don't fit well to negative binomial distribution. Consequently, error size may be mis-estimated which propagates to incorrectly computed _p_-value. However, _p_ = N/A doesn't say anything about the level of similarity of expressions between case and control group, as _p_ = 1 does. That's why I don't favor this approach.

To me, the eqivalent question is what to do, if e.g. _FDR_ = N/A. I would eliminate such genes from any further consideration and issue a warning already in the upload process.

Whatever the final solution, I just want to make sure that we are well aware of scientific consequences of the solution.

codeprimate123:
I would propose next solution for consideration: 1. short term fix: set fdr to 1 so they are not aligned at the top which is far from correct (users have been already asking Amirita about this issue), 2. definitive fix: tag all points with p or padj = 0 (aka FDR=0) as outliers and give user option to toggle them on or off on the plot and rename current toggle option "rescale x - axis" as that is what current toggle does.
That would be scientifically correct and reflect precisely what authors of algorithms wanted to achieve. Only question left is scientific consequences of FDR = padj but im not qualified yet to comment on that more then to bring to your attention that those aren't necessary equal.

PR #515, opened 2018-03-13
Updates to resolwebio/dnaseq docker image

jkokosar:


tjanez:
@jkokosar, @mzganec, in the future, [avoid commits that only update the CHANGELOG](https://github.com/genialis/resolwe-bio/pull/515/commits/aebab61605b5af81c0043a1a84832ea9f3bc4688).

The CHANGELOG should be updated in the commit where the change is made.

Regarding this particular pull request, I don't see the benefit of making `resolwebio/dnaseq` Docker image changes in separate commits, they could all go into 1.

PR #519, opened 2018-03-20
Use name-ordered BAM file in HTSeq-count process by default

mzganec:
Name ordering is used to prevent buffer overflow with large BAM files.

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -19,6 +19,11 @@ Added
 -----
 - Add CNVKit, LoFreq and GATK to ``resolwebio/dnaseq`` docker image
 
+Fixed
+-----
+- Use name-ordered BAM file for counting reads in ``HTSeq-count`` process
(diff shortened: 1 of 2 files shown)


tjanez:
@mzganec, was the BAM file ordering something a user of this process could rely on? In other words, is this a bug fix to prevent buffer overflow or a more substantial change of the process' "API"?

mzganec:
This is just a bug fix to prevent buffer overflow with large BAM files.

tjanez:
> This is just a bug fix to prevent buffer overflow with large BAM files.

Ok, thanks for the explanation.

I would move the CHANGELOG entry under **Fixed** then to signal that this is just a bug fix (even though it changes the default).

mzganec:
The thing is that this is not a proper fix. The default value is set just to circumvent the issue. If the user manually overrides the default value buffer overflow can still occur. I'll adjust the buffer size as well to mitigate the problem.

PR #524, opened 2018-03-26
Support additional file types in download_and_verify script

jkokosar:


resolwe_bio/docker_images/base/common.sh
```diff
@@ -51,6 +51,12 @@ download_and_verify() {
         *.cdf|cdf)
             mv "${name}.${extension}" "${name}/"
             ;;
+        *.css|css)
```
mzganec: A general solution for no extraction regardless of the extension should be preferred here, (1) possibly via another parameter to `download_and_verify()` or (2) as a re-formulation of the final case which now returns an error.

jkokosar: @kostko Please advise on how to best support no-extraction-needed use case

kostko: I think it would be better to just replace the catch-all case with no-extraction instead of emitting an error.

PR #522, opened 2018-03-21
Update amplicon pipeline

jkokosar:
@JenkoB Changes to VCF file outputs are included here

PR #525, opened 2018-03-26
Refactor BBDuk process and RNA-seq pipeline

mzganec:


resolwe_bio/processes/reads_processing/bbduk.yml
```diff
@@ -11,221 +11,363 @@
...
     executor:
       docker:
-        image: resolwebio/legacy:1.0.0
-  data_name: '{{ reads.fastq.0.file|default("?") }}'
-  version: 0.0.6
+        image: resolwebio/rnaseq:3.0.0
+    resources:
+      cores: 10
+  data_name: '{{ reads|sample_name|default("?") }}'
+  version: 0.0.7
```
JureZmrzlikar: This is a breaking change. Raise to `1.0.0`.

resolwe_bio/processes/reads_processing/bbduk.yml
```diff
@@ -11,221 +11,363 @@
...
   flow_collection: sample
   persistence: CACHED
   description: >
-    BBduk is capable of quality-trimming and filtering, adapter-trimming, contaminant-filtering via
-    kmer matching, sequence masking, GC-filtering, length filtering, entropy-filtering, format
-    conversion, histogram generation, subsampling, quality-score recalibration, kmer cardinality
-    estimation, and various other operations in a single pass.
+    BBDuk combines the most common data-quality-related trimming, filtering, and masking operations
+    into a single high-performance tool. It is capable of quality-trimming and filtering,
+    adapter-trimming, contaminant-filtering via kmer matching, sequence masking, GC-filtering,
```
JureZmrzlikar: Isn't it "k-mer"? For also other occurrencies.

resolwe_bio/processes/reads_processing/bbduk.yml
```diff
@@ -11,221 +11,363 @@
...
+          description: >
+            A reads needs at least this fraction of its total kmers to hit a reference in order to
+            be considered a match. If this and 'Minimum number of kmer hits' are set, the greater is
+            used.
+        - name: min_coverage_fraction
+          label: Minimum coverage fraction
+          type: basic:decimal
+          default: 0.0
+          description: >
+            A reads needs at least this fraction of its total bases to be covered by ref kmers to be
```
JureZmrzlikar: A read.

resolwe_bio/processes/workflows/bbduk_star_htseq.yml
```diff
@@ -4,7 +4,7 @@
   data_name: "Pipeline ({{ reads|sample_name|default('?') }})"
   requirements:
     expression-engine: jinja
-  version: 0.0.3
+  version: 0.0.4
```
JureZmrzlikar: Changing the interface of process is a breaking change. Raise the first number.

resolwe_bio/processes/reads_processing/bbduk.yml
mzganec: I used the same spelling as in [the official documentation](https://jgi.doe.gov/data-and-tools/bbtools/bb-tools-user-guide/bbduk-guide/). [Wikipedia](https://en.wikipedia.org/wiki/K-mer) uses your suggestion though.

JureZmrzlikar: True, leave it as it is. I guess it's a matter of preference...

resolwe_bio/processes/reads_processing/bbduk.yml
```diff
@@ -11,221 +11,363 @@
...
           label: Kmer length
           type: basic:integer
           default: 27
           description: >
-            Kmer length used for finding contaminants.  Contaminants shorter than k will not be found.
-            k must be at least 1.
-        - name: rcomp
-          label: Reverse-complements.
+            Kmer length used for finding contaminants. Contaminants shorter than k will not be
+            found. k must be at least 1.
```
JureZmrzlikar: `k --> kmer_length`. Also one line before.

resolwe_bio/processes/reads_processing/bbduk.yml
```diff
@@ -11,221 +11,363 @@
...
       label: Reads
       type: data:reads:fastq:single
-    - name: adapters
-      label: Adapters
+    - name: show_advanced
+      label: Show advanced parameters
+      type: basic:boolean
+      default: false
+    - name: reference
+      label: Reference
```
jkokosar: Missing description of the input field. Its not obvious what the `reference` is. Maybe try to explain that here a user should provide a multi-sequence .fasta file with adapters

resolwe_bio/processes/reads_processing/bbduk.yml
```diff
@@ -11,221 +11,363 @@
...
-      label: Reads file
+      label: Remaining reads
+      type: list:basic:file
+    - name: dumped
+      label: Dumped reads
+      type: list:basic:file
+    - name: statistics
+      label: Statistics
+      type: list:basic:file
+    - name: kmers
```
jkokosar: What is contained in this output files?

resolwe_bio/processes/reads_processing/bbduk.yml
```diff
@@ -236,464 +378,684 @@
...
+        ymin={{ header_parsing.y_min }} \
+        xmax={{ header_parsing.x_max }} \
+        ymax={{ header_parsing.y_max }} \
+        entropy={{ complexity.entropy }} \
+        entropywindow={{ complexity.entropy_window }} \
+        entropyk={{ complexity.entropy_k }} \
+        minbasefrequency={{ complexity.min_base_frequency }} \
+        entropymask={{ bool[complexity.entropy_mask] }} \
+        threads={{ requirements.resources.cores }} \
         -Xmx1g
```
jkokosar: Its better to have the memory in requirements resources and than use jinja formating like [here](https://github.com/genialis/resolwe-bio/pull/461/files#diff-50334eb3c7235ebcce15599e56155610R179)

resolwe_bio/tests/processes/test_reads_filtering.py
```diff
@@ -345,67 +345,47 @@ def test_bbduk_single(self):
...
 
-        inputs = {
-            'reads': reads.id,
-            'adapters': {
-                'bbduk_adapters': 'truseq'},
-            'trimming_par': {
-                'kmask': 'lc'},
-            'barcode_par': {
-                'barcode_seq': ['GCGTTT', 'TATGNNN']}
+        self.assertFiles(filtered_reads, 'fastq', ['bbduk_reads.fastq.gz'], compression='gzip')
```
jkokosar: Please retain the whitespace in the file name. Maybe also split the reads into two files and upload them as two separate seq lanes to cover this part of the process code.

resolwe_bio/processes/reads_processing/bbduk.yml
```diff
@@ -11,221 +11,363 @@
...
       type: data:reads:fastq:single
-    - name: adapters
-      label: Adapters
+    - name: show_advanced
+      label: Show advanced parameters
+      type: basic:boolean
+      default: false
+    - name: reference
+      label: Reference
+      hidden: '!show_advanced'
```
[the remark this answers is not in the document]
mzganec: (1) I ran BBDuk without any parameters to obtain the full set of supported parameters. Note that the version of BBDuk installed in our RNA-seq Docker image is newer than the version on GitHub that you linked which may be the reason why you couldn't find some parameters. Process inputs follow the same order as parameters displayed in the command line and are grouped similarly as well. I agree that matching command-line options with descriptions of inputs in the process is not always trivial. I can (a) add names of command-line parameters to descriptions or (b) change the names of process inputs to match the names of command-line parameters. However, (a) I didn't want the process input schema to look too technical and arouse a command-line-like feeling. Many of our users want to move away from the command line. (b) BBDuk uses a non-standard parameter naming conventions which doesn't play well with the rest of our processes. Lastly, imagine a refactoring where process inputs are recipes-like user-friendly and cannot be mapped one-to-one onto command-line parameters of the tool. I think that we should strive towards improving the descriptions of inputs to the point where the user would be able to use the process without the manual.

(2) This refactoring closely follows the function of the underlying tool. For example, if you call the process without parameters (except reads) it behaves similarly to the tool called from the command-line with just reads. This provides full control of operation and guarantees basic predictability of results: if an input in the process is not set explicitly, the tool's default value is used. This (a) eases integration into workflows because only the parameters explicitly set there have an effect on the operation of the tool and the process doesn't do anything unexpected automatically. This also (b) keeps power users in full control to achieve exactly what they want. The process is not intended to be used by basic users. For them we have recipes, workflows, and easy to use meta-processes named _for biologists_.

Custom input with adapters provided by BBDuk has been removed due to redundancy. The idea is to make those adapters available as public resources on our servers where they can be chosen as a regular data object by all trimming and filtering processes, including Trimmomatic and Cutadapt.

(3) If only reads are provided, they are copied to output and analyzed by FastQC. A warning is displayed that reads remain unmodified. This solution is based on what we've recently implemented in Trimmomatic and follows the operation of the tool itself which also prints out an equivalent warning.

Which parameters are important and which not is to some extent subjective so I decided to hide all non-essential parameters by default which gives a clean and non-intimidating interface to start with. To enable/show reference sequence-related parameters only after such sequences are provided is certainly an interesting idea. I base my decision not to implement this now on the fact that the process is much more likely to be used within a workflow than directly. Given the limited amount of time I deem other tasks more essential but I'm open to implementing this in the future.

resolwe_bio/processes/reads_processing/bbduk.yml
mzganec: The idea is interesting but the implementation is poor. See [this RedHat post](https://developers.redhat.com/blog/2017/04/04/openjdk-and-containers/):
> Specifying -Xmx=1g is telling the JVM to allocate a 1gb heap. It’s not telling the JVM to limit its entire memory usage to 1gb. There are card tables, code caches, and all sorts of other off heap data structures.

Even by specifying `-XX:MaxRAM` I don't think that it's wise to instruct the tool to use all RAM in `requirements.resources.memory` up to the very last byte. There is likely some overhead by the shell etc. as well. @tjanez may want to contribute here.

mzganec: I based the maximum memory setting on the suggestion in [the official documentation](https://jgi.doe.gov/data-and-tools/bbtools/bb-tools-user-guide/bbduk-guide/), paragraph _Memory_. I like the idea of making the memory limit used by Java dependent on `requirements.resources.memory`. Thanks for the suggestion.

resolwe_bio/tests/processes/test_reads_filtering.py
mzganec: This is an output test assertion. If it doesn't handle filenames with spaces correctly, then we have much bigger problems :wink:. The input file is defined a few lines above and already contains a space. I'll follow your suggestion and use two input files though.

resolwe_bio/processes/reads_processing/bbduk.yml
mzganec: A list of matching _k_-mer sequences. I removed this output along with dumped reads because I don't see enough value for a typical user. Additionally, the file in `kmers` is much larger than all other outputs combined.

resolwe_bio/processes/reads_processing/bbduk.yml
```diff
@@ -11,221 +11,363 @@
...
-              value: truseq_rna
-        - name: reference
-          label: List of reference files
+        - name: sequences
+          label: Sequences
           type: list:data:seq:nucleotide
           required: false
-        - name: literal_reference
-          label: List of literal reference sequences
+        - name: literal_sequences
```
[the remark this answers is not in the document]
mzganec: I added basic instructions how to input a list of strings via front-end although I would rather see this implemented as a system-wide solution (a question mark with help or an `onmouseover` event). @hadalin

resolwe_bio/processes/reads_processing/bbduk.yml
mzganec: Good idea! I like the concept of a reference. It encompasses several types of sequences such as adapters, contaminants, and degenerate sequences like primers. I added additional description as you suggested. Note, however, that description of a `group` is not displayed on the front-end. I therefore added the description to the first input in this group.

resolwe_bio/processes/reads_processing/bbduk.yml
jkokosar: A bit naive, but you could assign 2GB (or 1.5GB, ..) using `requirements.resources.memory`, and use only portion of that in the java flag via division/rounding. This would be better than the defaults that are assigned to a whole process now (4 or 8 GB?)

resolwe_bio/tests/processes/test_reads_filtering.py
jkokosar: :+1:

resolwe_bio/processes/workflows/bbduk_star_featurecounts.yml
```diff
@@ -0,0 +1,377 @@
...
+            trim_pairs_evenly: t
+            trim_by_overlap: t
+      - id: alignment
+        run: alignment-star
+        input:
+          reads: '{{ steps.preprocessing }}'
+          genome: '{{ input.alignment.star_index }}'
+          filtering:
+            outFilterType: BySJout
+            outFilterMultimapNmax: 20
```
jkokosar: In this specific workflow, it might make sense to set this parameter to 1, however, Its [not wrong](http://seqanswers.com/forums/showthread.php?t=41991) to leave it at 20. The upside if a more versatile .bam if other tools will potentially be used in downstream analysis. The downside is a bit larger .bam that holds information not useful for FeatureCounts.

resolwe_bio/processes/reads_processing/bbduk.yml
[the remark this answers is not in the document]
JenkoB: With a purpose of reproducibility of the BBduk results I think we should enable user to understand from the label or description which BBduk parameter is in use. Since there is so many parameters it is rather hard to find the one that user wish to define, if he has pre-defined command line pipeline. However I agree that many of our users want to move away from the command line. Why not adding at the end of description a sentence:"BBduk command line parameter `k`"? This is just my concern and suggestion based on user story, which I have described. The decision is up to you. 

To use tool from BioDash it would be much easier if not all parameters would be in advance part and there would be some parameter enabling. However I understand and agree this would take some additional time, and it is not a priority. I have added some of suggestions, where I found it most confusing...

About providing only reads file, I have the missed Warning message at first. I only added a question regarding `Minimum length` parameter.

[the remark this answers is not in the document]
mzganec: I added command-line parameter names in square brackets to avoid confusion with existing round ones in the labels. I followed your other suggestions as well.

resolwe_bio/processes/reads_processing/bbduk.yml
mzganec: I've set `-Xmx` to 85% of memory specified in requirements, as suggested in the help of BBDuk:
> The max is typically 85% of physical memory.

resolwe_bio/processes/workflows/bbduk_star_featurecounts.yml
mzganec: I followed your suggestion and changed the default value of this parameter to 1. Initially, all parameters were copied from `workflow-bbduk-star-htseq`.

resolwe_bio/processes/reads_processing/bbduk.yml
```diff
@@ -236,464 +381,672 @@
...
-          description: >
-            Regions with average quality below this will be trimmed.
-        - name: minlength
-          label: Minlength
+        - name: trim_poly_A
+          label: Minimum length of poly-A or poly-T tails to trim on either end of reads
+          type: basic:integer
+          default: 0
+        - name: min_length
+          label: Minimum length
```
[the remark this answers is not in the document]
mzganec: Yes, short reads are discarded by default. Therefore the warning printed by BBDuk itself is incorrect. I therefore modified the warning displayed by the process and moved it out of _Advanced_ section as you proposed.

resolwe_bio/processes/reads_processing/bbduk.yml
```diff
@@ -236,464 +381,672 @@
...
-    - name: threads
-      label: Number of threads
-      type: basic:integer
-      required: false
-      description: >
-        Set number of threads to use; default is number of logical processors.
+            Literal barcode sequences can be specified by inputting them one by one and pressing
+            Enter after each sequence.
+        - name: x_min
+          label: Minimum X coordinate
```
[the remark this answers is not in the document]
mzganec: Start `resolwebio/rnaseq:3.0.0` Docker container and run `bbduk.sh` which displays the help. I added a description of these inputs to improve user experience.

docs/CHANGELOG.rst
```diff
@@ -16,10 +16,12 @@ Changed
   ``trimmomatic-paired`` processes
 - Bump STAR aligner version in ``resolwebio/rnaseq`` docker image to 2.5.4b
 - Bump Primerclip version in ``resolwebio/dnaseq`` docker image
+- **BACKWARD INCOMPATIBLE:** Refactor ``BBDuk`` process
```
JenkoB: Tadej said that `**BACKWARD INCOMPATIBLE:**` needs to be on the top of changelog section.
https://github.com/genialis/resolwe-bio/pull/522#discussion_r177393288

mzganec:
@jkokosar I changed Docker image of STAR process to `resolwebio/rnaseq`. The index building STAR process could not be transitioned to `resolwebio/rnaseq` due to missing [`file` command](https://github.com/genialis/resolwe-bio/blob/master/resolwe_bio/processes/alignment/star.yml#L475).

mzganec:
That was quite a team effort. Thanks, guys!

PR #526, opened 2018-03-27
Miscellaneous minor improvements

tjanez:
Improving robustness of `goea.py` tool, debuggability of `Feature` and `Mapping` models and ensuring no tests could ever connect to `qa2.genialis.com` server.

### DIFF
--- a/resolwe_bio/tools/goea.py
+++ b/resolwe_bio/tools/goea.py
@@ -6,6 +6,8 @@ import argparse
 import tempfile
 import resdk
 
+from resolwe_runtime_utils import error, warning  # pylint: disable=import-error
+
 
 def parse_arguments():
     """Parse command line arguments."""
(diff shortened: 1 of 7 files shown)


resolwe_bio/tools/goea.py
```diff
@@ -33,27 +34,36 @@ def main():
...
         target_ids = genes
     else:
-        features = res.mapping.filter(
+        mapping_res = res.mapping.filter(
             source_db=args.source_db,
             source_species=args.species,
             target_db=args.target_db,
             target_species=args.species,
-            source_id=genes)
+            source_id=genes
```
dblenkus: Add a trailing comma.

resolwe_bio/tools/goea.py
```diff
@@ -33,27 +34,36 @@ def main():
...
             exit(1)
 
-        target_ids = [str(feature.target_id) for feature in features]
+        mappings = {}
+        for m in mapping_res:
+            if m.source_id in genes:
+                if m.source_id not in mappings:
+                    mappings[m.source_id] = m.target_id
+                else:
+                    print(warning("Mapping {} returned multiple times.".format(m)))
```
dblenkus: Add a single quotes around `{}`.

tjanez: Do you think it is necessary?

Here is an example output of the current code:
```
Mapping <Mapping source_db='NCBI' source_id='56535' target_db='MGI' target_id='MGI:1929646'> returned multiple times.
```

resolwe_bio/tools/goea.py
tjanez: Done.

resolwe_bio/tools/goea.py
dblenkus: Ok, leave it like this.

PR #529, opened 2018-03-30
Handle filenames with spaces in some processes

JureZmrzlikar:
Processes afrfected:
  - alignment-hisat2
  - alignment-bowtie
  - prepare-geo-chipseq
  - prepare-geo-rnaseq


resolwe_bio/processes/alignment/bowtie.yml
```diff
@@ -140,14 +140,14 @@
       reporting={{reporting.r}}
 
       {% for r in reads.fastq %}
-        READS_NAME=`basename {{ r.file }} .fastq.gz`
+        READS_NAME={{ r.file | basename | replace(".fastq.gz", "") | replace(" ", "") }}
```
mzganec: If we want to handle filenames containing spaces by removing spaces, it would be better to do it systematically via input sanitizer for all processes. I suggested this some time ago but the general opinion at the time was that filenames should be conserved, including spaces. Consider [this approach](https://github.com/genialis/resolwe-bio/blob/master/resolwe_bio/processes/reads_processing/bbduk.yml#L410).

resolwe_bio/processes/alignment/bowtie.yml
```diff
@@ -478,7 +478,7 @@
         {% endfor %}
       {% endif %}
 
-      FW_NAME=`basename {{ reads.fastq.0.file }} .fastq.gz`
+      FW_NAME={{ reads.fastq.0.file | basename | replace(".fastq.gz", "") | replace(" ", "") }}
```
mzganec: `(reads.fastq|first).file`

JureZmrzlikar: Is there any benefit of using `first` instead of `0`? Just curious. For me, `reads.fastq[0].file` is even better. Especially, considering that we would like to port this code to Python someday...

mzganec: There is no benefit. This is just an idea, you can leave it as it is.

mzganec:
Please fix also `cufflinks` and `cuffquant`.

PR #532, opened 2018-04-05
Fix genome upload process to correctly handle filenames with dots

mzganec:
Instead of relying on the first dot in index filenames to split basenames into prefix and extension, prefix is determined as the longest common substring among all index files starting at the beginning of index filenames.


mstajdohar:
Do not forget about tests and the change log.

mzganec:
I need to prepare `assertDir()` in resolwe first.

mstajdohar:
I guess we can finally merge this now. Thanks for the `assertDir` feature! Please rebase.

mzganec:
I'll add `assertDir` tests and wrap this up.

resolwe_bio/tests/processes/test_upload.py
```diff
@@ -322,18 +322,13 @@ def test_upload_genome(self):
...
-        self.assertFields(genome, "index_walt", {'dir': 'walt_index'})
+        self.assertDir(genome, 'index_bt', 'index_bt.tar.gz')
+        self.assertDir(genome, 'index_bt2', 'index_bt2.tar.gz')
+        self.assertDir(genome, 'index_bwa', 'index_bwa.tar.gz')
+        self.assertDir(genome, 'index_hisat2', 'index_hisat2.tar.gz')
+        self.assertDir(genome, 'index_subread', 'index_subread.tar.gz')
+        self.assertDir(genome, 'index_walt', 'index_walt.tar.gz')
+
         del genome.output['fasta_track_jbrowse']['total_size']  # Non-deterministic output.
         self.assertFields(genome, 'fasta_track_jbrowse', {'refs': ['seq'],
```
mzganec: @JureZmrzlikar Although not the aim of this pull request, I'm open to suggestions how to improve this line. The field `fasta_track_jbrowse` is of type `basic:file` but `assertFile` doesn't check the `refs` field in the dict.

JureZmrzlikar: Let's leave this until it really becomes an issue. :-)

PR #534, opened 2018-04-09
Change backtick to apostrophe in RNA-seq description

mzganec:
This change is needed to avoid ASCII error in Ansible.


docs/CHANGELOG.rst
```diff
@@ -38,6 +38,7 @@ Changed
 - Run ``vc-realign-recalibrate`` process using multiple cpu cores to optimize
   the processing time
 - Use ``resolwebio/rnaseq`` Docker image in ``alignment-star`` process
+- Change backtick to apostrophe in RNA-seq description to avoid ASCII error
```
dblenkus: I think you don't need to add a change log for this.

tjanez:
@mzganec, just a minor comment, these characters were not backticks (i.e. `` ` ``), but some non-ascii apostrophes.

PR #535, opened 2018-04-10
Fix whitespace handling in STAR indexing tool and upload fasta file processes

jkokosar:


resolwe_bio/processes/alignment/star.yml
```diff
@@ -507,7 +507,7 @@
         {% if advanced.genomeSAsparseD %}--genomeSAsparseD {{ advanced.genomeSAsparseD }}{% endif %} \
         --genomeDir ./star_index/ \
         --genomeFastaFiles genome.fasta \
-        --sjdbGTFfile ${ANNOT} \
+        --sjdbGTFfile "${ANNOT}" \
```
mzganec: This will fail for annotations that are not in GTF format. See how `ANNOT` is set a few lines above.

jkokosar: Nice catch. I have prepared a fix + updated the default .gff3 test file used by `prepare_annotation_gff` to account for whitespace also in this scenario.

mzganec:
Based on the test error messages it seems that `tophat2` and `rsem-prepare-reference` tools cannot handle annotation filenames with spaces. Annotation files need to be renamed in those processes before the tools start using them.

mstajdohar:
I suggest creating a symlink.

jkokosar:
Uh, this really is a can of worms. I fixed some additional processes, but even more fixes will be needed when __all__ test files are changed to include whitespace @mzganec please re-review

resolwe_bio/processes/annotation/cuffmerge.yml
```diff
@@ -87,15 +87,17 @@
 
       re-progress 0.1
 
-      {% if gff.annot.file %}
-        gffread {{gff.annot.file}} -T -o reference_annotation.gtf
+      {% if annotation|type|subtype('data:annotation:gff3') %}
```
mzganec: Input field named `annotation` does not exist. This is not caught by tests.

PR #538, opened 2018-04-16
Add automated library strand detection to featureCounts tool

jkokosar:


resolwe_bio/processes/expression/featureCounts.yml
```diff
@@ -42,14 +42,40 @@
...
+          description: >
+            Transcriptome index file created using the Salmon indexing tool.
+            cDNA (transcriptome) sequences used for index file creation must be
+            derived from the same species as the input sequencing reads to
+            obtain the reliable analysis results.
+        - name: sampling_rate
+          label: Alignment file sampling rate
+          type: basic:decimal
+          default: 0.01
+          required: false
```
mzganec: Setting the `required` attribute is redundant, if `default` is used. The same applies also to the workflow and other associated processes.

resolwe_bio/processes/support_processors/library_strandedness.yml
```diff
@@ -7,9 +7,9 @@
       memory: 8192
     executor:
       docker:
-        image: resolwebio/rnaseq:2.1.0
+        image: resolwebio/rnaseq:3.1.0
   data_name: 'Library strandedness ({{ reads|sample_name }})'
-  version: 0.0.1
+  version: 0.0.3
```
mzganec: Why a double bump?

resolwe_bio/tests/processes/test_expression.py
```diff
@@ -382,3 +382,43 @@ def test_salmon_index(self):
...
+                'source': 'ENSEMBL',
+                'species': 'Homo sapiens',
+                'build': 'ens_90',
+            })
+
+            annotation = self.run_process('upload-gtf', {
+                'src': 'annotation_rsem.gtf.gz',
+                'source': 'ENSEMBL',
+                'species': 'Homo sapiens',
+                'build': 'ens_90'
```
mzganec: Add a trailing comma.

resolwe_bio/processes/expression/featureCounts.yml
```diff
@@ -42,14 +42,40 @@
             Indicate if strand-specific read counting should be performed. For
             paired-end reads, strand of the first read is taken as the strand
             of the whole fragment. FLAG field is used to tell if a read is
-            first or second read in a pair.
+            first or second read in a pair. Automated strand detection is enabled
+            using the [Salmon](https://salmon.readthedocs.io/en/latest/library_type.html)
+            tool's build-in functionality. To use this option, cDNA (transcriptome)
+            index file crated using the Salmon indexing tool must be provided.
           choices:
```
mzganec: Consider changing the `default` attribute to `auto` here as well. The same applies also to the workflow and other associated processes.

jkokosar: I have deliberately left the basic process behaviour to be as similar as is expected from the featureCounts tool. Auto is than used in workflows and elsewhere where it makes sense.

resolwe_bio/processes/expression/featureCounts.yml
```diff
@@ -409,14 +438,33 @@
...
+      {% if alignment.assay_type == 'auto' and alignment.cdna_index %}
+        {% if alignment.aligned_reads.species != alignment.cdna_index.species %}
+          re-error \
+            {{ ('Species of aligned reads (%s) and cDNA index (%s) do not match. ' ~
+              'Please provide aligned reads and cDNA index with the same species.')
+              |format(alignment.aligned_reads.species, alignment.cdna_index.species) }}
+        {% endif %}
+      {% endif %}
+
+      . detect_strandedness.sh \
```
mzganec: The script should be structured in such a way that it can be called without the leading dot.

resolwe_bio/processes/expression/featureCounts.yml
```diff
@@ -409,14 +438,33 @@
...
+        {{ alignment.assay_type }} \
+        {{ alignment.aligned_reads.bam.file }} \
+        {{ alignment.sampling_rate }} \
+        {{ alignment.cdna_index.index.dir }} \
+        {{ requirements.resources.cores }} \
```
mzganec: Remove the trailing backslash.

resolwe_bio/processes/support_processors/library_strandedness.yml
```diff
@@ -78,15 +78,16 @@
         {% set reads_input = '-1 mate1.fastq -2 mate2.fastq' %}
       {% endif %}
 
-      salmon quant \
+      salmon --no-version-check quant \
```
mzganec: Try to re-use the script `detect_strandedness.sh` in this process.

resolwe_bio/processes/expression/featureCounts.yml
jkokosar: If I `echo $STRANDEDNESS` and save it to a variable within the process, than I cannot use `re-save` / `re-checkrc` commands, since they also print to stdout. Its either leading dot or no error reporting, so I took a lesser evil. Any other suggestion, other than doing it in Python when the time is right?

resolwe_bio/processes/support_processors/library_strandedness.yml
jkokosar: This process operates on raw fastq data, the `detect_strandedness.sh` script is written with post-alignment needs in mind

PR #539, opened 2018-04-16
Fix some names and defaults

JureZmrzlikar:
These errors were identified with newly implemented process
validation.

### DIFF
--- a/resolwe_bio/descriptors/reads_detailed.yml
+++ b/resolwe_bio/descriptors/reads_detailed.yml
@@ -1,6 +1,7 @@
 - slug: reads-detailed
   name: Reads annotation (detailed)
   description: Annotation of raw sequencing data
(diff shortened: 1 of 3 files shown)


JureZmrzlikar:
They don't have version defined, so i concluded that they don't need it... :-) Will do. Is `1.0.0` OK? Is any of these schemas deprecated?

jkokosar:
Either `1:0:0` or `0:0:1`  is OK. Deprecated stuff will be removed in one go.

JureZmrzlikar:
If version is not specified, the default is `0.0.0`. I decide to set `0.0.1` for `reads-detailed`, since we just change the default (aka. bugfix change). For `transmart`, i decided to give it `1.0.0`, since it is a breaking change (in a sense that it changes the API of a schema).

PR #546, opened 2018-04-23
Fix Bokeh assets file names to match the Bokeh version installed in Docker

jkokosar:


mstajdohar:
Do we have a test for this?

jkokosar:
`re-checkrc` was added to process code to catch the failing `cp` command to avoid this issue in the future.

PR #553, opened 2018-05-11
Fix calculation of FPKM

romunov:
This fix addresses the bug which affects calculation of FPKM for results from `featureCounts`. The problem was that raw count file from `featureCounts` comes unsorted and this was a problem because division was made as if sorted. The script now uses merge ("left join") to match data from gtf file and raw count data and does all the calculations.


docs/CHANGELOG.rst
```diff
@@ -57,6 +57,7 @@ Fixed
   duplicated amplicon names are included. Validation will now pass also
   for primer sequences in lowercase.
 - Fix allele frequency (AF) calculation in ``snpeff`` process
+- Fix bug in script for calculating FPKM
```
mstajdohar: Provide more info about the bug.

resolwe_bio/tools/expression_fpkm_tpm.R
```diff
@@ -12,31 +13,36 @@ args = parser$parse_args(commandArgs(trailingOnly=TRUE))
...
 annot <- import.gff(args$annotation, format=args$format, genome="NA", feature.type=args$feature_type)
-exon_lengths=width(annot)
-names(exon_lengths)=elementMetadata(annot)[,args$id_attribute]
+exon_lengths <- width(annot)
+names(exon_lengths) <- elementMetadata(annot)[, args$id_attribute]
 
-# calculate gene lengths
-exon_lengths_by_gene=split(exon_lengths, names(exon_lengths))
-gene_lengths=sapply(exon_lengths_by_gene, sum)
+# calculate gene lengths, note that the result is lexographically sorted
```
mzganec: lexicographically

mstajdohar:
How long does the script take for human expressions?

docs/CHANGELOG.rst
mstajdohar: The info should be such that when the CS, product or marketing read about the fix in release notes, they know what it is about. I know you are part of the CS team, but the description should be such that the message does not have to be explained verbally.

romunov: I'm on it.

resolwe_bio/tools/expression_fpkm_tpm.R
```diff
@@ -12,31 +13,36 @@ args = parser$parse_args(commandArgs(trailingOnly=TRUE))
...
 annot <- import.gff(args$annotation, format=args$format, genome="NA", feature.type=args$feature_type)
-exon_lengths=width(annot)
-names(exon_lengths)=elementMetadata(annot)[,args$id_attribute]
+exon_lengths <- width(annot)
+names(exon_lengths) <- elementMetadata(annot)[, args$id_attribute]
 
-# calculate gene lengths
-exon_lengths_by_gene=split(exon_lengths, names(exon_lengths))
-gene_lengths=sapply(exon_lengths_by_gene, sum)
+# calculate gene lengths, note that the result is lexicographically sorted
```
JenkoB: Change to: # calculate gene or transcript lengths

resolwe_bio/tools/expression_fpkm_tpm.R
```diff
@@ -12,31 +13,36 @@ args = parser$parse_args(commandArgs(trailingOnly=TRUE))
...
+names(ncounts) <- c("Gene", "raw_count")
+ncounts$norm_count <- ncounts$raw_count / sum(ncounts$raw_count) # library_size
 
 # obtain CPM
-cpm=ncounts*1e6
+cpm <- ncounts$norm_count * 1e6
+names(cpm) <- ncounts$Gene
+cpm <- cpm[order(names(cpm))] # lexicographically order cpm
 
 # normalise counts by gene length
```
JenkoB: In the next two lines after this comment you don't normalise by gene length, but you are selecting genes in counts table that you will normalise.

romunov: Thanks, changed.

resolwe_bio/tools/expression_fpkm_tpm.R
romunov: Done.

PR #556, opened 2018-05-15
Add tools to support genome browsers to chipseq and resolwebio Docker images 

JenkoB:
@kostko @JureZmrzlikar please review.


resolwe_bio/docker_images/rnaseq/packages-debian.txt
```diff
@@ -19,3 +19,6 @@ libssl-dev
 python3-numpy
 python3-pandas
 python3-sklearn
+
+# required for jbrowse
+# libbio-perl-perl
```
JureZmrzlikar: You probably wish to un-comment this.

JenkoB: Actually, this needs to be removed. Thanks.

PR #564, opened 2018-05-28
Fix rendering of coverage plot html file in multisample download

JureZmrzlikar:


mstajdohar:
Why did this fail before?

PR #567, opened 2018-05-30
Add aligners to resolwebio/rnaseq Docker image

JenkoB:
## Checklist

    before non-breaking changes.
    commit(s).


JenkoB:
@mzganec @JureZmrzlikar please review.

PR #572, opened 2018-06-01
Use resolwebio/rnaseq:3.2.0 Docker image in processes that output reads

mzganec:
This update unifies the version of `FastQC` tool used for quality control of reads.

## Checklist

    before non-breaking changes.
    commit(s).

### Processes ###

  can be performed).
  and publication.

  tests).

### Tests ###

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -42,6 +42,43 @@ Changed
 - **BACKWARD INCOMPATIBLE:** Count fragments (or templates) instead of reads by
   default in ``featureCounts`` process and ``BBDuk - STAR - featureCounts``
   pipeline. The change applies only to paired-end data.
+- **BACKWARD INCOMPATIBLE:** Use ``resolwebio/rnaseq:3.2.0`` Docker image
+  in the following processes that output reads:
+
+  - ``upload-fastq-single``
+  - ``upload-fastq-paired``
+  - ``files-to-fastq-single``
+  - ``files-to-fastq-paired``
+  - ``reads-merge``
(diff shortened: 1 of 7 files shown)


docs/CHANGELOG.rst
```diff
@@ -42,6 +42,43 @@ Changed
...
+  quality control of reads in the aforementioned processes. The new Docker image
+  comes with an updated version of Cutadapt (1.16) which affects
+  the following processes:
+
+  - ``cutadapt-single``
+  - ``cutadapt-paired``
+  - ``cutadapt-custom-single``
+  - ``cutadapt-custom-paired``.
+
+  The new Docker images includes also an updated version of Trimmomatic (0.36)
```
JureZmrzlikar: include

mzganec:
Indeed, data objects produced with the old pipeline still work, but `FastQC` report can be different now. [The changes](https://www.bioinformatics.babraham.ac.uk/projects/fastqc/RELEASE_NOTES.txt) introduced after version 0.11.5 are substantial.

If an output field value of a data object changes this in my opinion represents a backward incompatible change regardless of if this value is used later in the pipeline or not. On the other hand, a performance improvement, for example, would be backward compatible.

JureZmrzlikar:
> The changes introduced after version 0.11.5 are substantial.

I fail to see any substantial changes in the changelog, they claim more or less to be bugfix releases, except for the k-mer part of the report. What exactly do you mean?

mzganec:
Exactly that, _k_-mers and bug fixes. We can discuss it on the daily.

PR #573, opened 2018-06-05
Add package file to resolwebio/rnaseq Docker image

JenkoB:
## Checklist

    before non-breaking changes.
    commit(s).


docs/CHANGELOG.rst
```diff
@@ -33,6 +33,8 @@ Added
   - ``Picard Tools (v2.18.5)``
   - ``bedGraphToBigWig (kent-v365)``
 - Support filtering by type on feature API endpoint
+- Add Debian package file (1:5.32.2) to ``resolwebio/rnaseq:3.3.0`` Docker
```
mzganec: Replace `1:5.32.2` with `1:5.32-2` (note the dash) and put `file` in double backticks.

PR #581, opened 2018-06-14
Fix typo in trimmomatic process

dblenkus:


mstajdohar:
We really should refactor Resolwe not to swallow errors.

PR #585, opened 2018-06-26
[WIP] Refactor sample hierarchical clustering

mzganec:
Improve performance, add informative warning and error messages.

## Checklist

    before non-breaking changes.
    commit(s).
* [X] All output fields have their ``re-save`` calls.

### Processes ###

  can be performed).
  and publication.

  tests).

### Tests ###


mzganec:
@jkokosar @JureZmrzlikar @codeprimate123 Please have a look, if this PR is headed in the right direction. Apart from transitioning to optimal ordering from `scipy` the idea is to provide more informative messages why sample clustering failed. After we align our views I want to amend this PR with similar improvements in gene clustering. Clustering errors became more of an issue after we changed the distance metric on the heatmap to correlation which is more sensitive about the quality of input expressions.

resolwe_bio/processes/clustering/hierarchical_clustering.yml
```diff
@@ -124,9 +117,7 @@
     runtime: polyglot
     language: bash
     program: |
-
       {% for e in exps %}
-
         {% if e.source != (exps|first).source %}
           re-warning "All expression data must be annotated by the same genome database."
           re-error "Sample {{ exps|first|name }} has {{ (exps|first).source }} gene IDs, while sample {{ e|name }} has {{ e.source }} gene IDs."
```
JureZmrzlikar: I know this is not part of your PR, but what is logic behind this? We have a warning and two errors directly afterwards. Why raise a warning if error follows immediately? And why raise two errors, second of which is totally unrelated to the if clause that triggers it...

JureZmrzlikar: Same holds for many error / warning messages all over this process.

resolwe_bio/tools/samplehcluster.py
```diff
@@ -53,36 +49,35 @@ def get_expressions(fnames, sep='\t', gene_set=[]):
...
     """
     if log2:
-        try:
-            expressions = expressions.applymap(lambda x: np.log2(x + const))
-        except:
+        expressions = expressions.applymap(lambda x: np.log2(x + const))
+        if expressions.isnull().values.any():
             msg = 'Cannot apply log2 to expression values.'
             print(error(msg))
             raise ValueError(msg)
```
JureZmrzlikar: I don't see any value in raising ValueError after `print(error(msg))`. Instead of repeating the same message with `raise ValueError(msg)`, just use `sys.exit(1)`

JureZmrzlikar: Same for many examples below.

JureZmrzlikar: Since there are so many lines of form:

    print(error(msg))
    raise ValueError(msg) / sys.exit(1)

consider defining function to do this operation in one line...

resolwe_bio/processes/clustering/hierarchical_clustering.yml
JureZmrzlikar: True, my apologies. I agree: we should use one verbose message and `re-error`.

jkokosar:
@mzganec Error messages look good to me!

resolwe_bio/tools/samplehcluster.py
```diff
@@ -174,24 +141,118 @@ def output_json(result=dict(), fname=None):
...
+        print(error(msg))
+        raise ValueError(msg)
+
+    if len(args.sample_files) < 2:
+        msg = 'Select at least two samples to compute hierarchical clustering of samples.'
+        print(error(msg))
+        raise ValueError(msg)
+
+    if len(args.gene_labels) == 1 and args.distance_metric != 'euclidean':
+        msg = ('Select at least two genes to compute hierarchical clustering of samples with '
```
JenkoB: Does it makes sense to add here to error message: "Select more genes or use Euclidean distance metric."?

JenkoB:
Looks great, very comprehensive work!

PR #586, opened 2018-06-28
Fix bamtobigwig calculation to timeout after defined time

JenkoB:
## Checklist

    before non-breaking changes.
    commit(s).

### DIFF
--- a/resolwe_bio/processes/alignment/bowtie.yml
+++ b/resolwe_bio/processes/alignment/bowtie.yml
@@ -13,7 +13,7 @@
     resources:
       cores: 10
   data_name: "Alignment ({{ reads.fastq.0.file|basename|default('?') }})"
-  version: 1.3.1
+  version: 1.3.2
   type: data:alignment:bam:bowtie1
   category: analyses:alignment
   flow_collection: sample
@@ -124,6 +124,7 @@
     - name: bigwig
       label: BigWig file
       type: basic:file
+      required: false
     - name: species
       label: Species
       type: basic:string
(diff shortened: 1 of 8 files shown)


JenkoB:
@mzganec @JureZmrzlikar

resolwe_bio/tools/bamtobigwig.sh
```diff
@@ -5,15 +5,36 @@ source /etc/profile.d/resolwe-base.sh
 bam_file="$1"
 number_of_cores="$2"
 bin_size="$3"
+time="$4"
```
JureZmrzlikar: Use a more descriptive variable, e.g. `time_limit`. This will make it obviuous to developer that this is a time limit for job execution.

resolwe_bio/tools/bamtobigwig.sh
```diff
@@ -5,15 +5,36 @@ source /etc/profile.d/resolwe-base.sh
...
 
+#get bam size in bytes
+bam_size=$(stat -c%s "${NAME}.bam")
+#if bam_size is less than 1 GB (1073741824bytes) set it to 1 GB
+if [[ ${bam_size} -lt 1073741824 ]]
+then
+  bam_size=1073741824
+fi
+
+time_deeptools=$(echo print ${time}*${bam_size}/1073741824. | python)
```
JureZmrzlikar: Rather use the following line::

    time_deeptools=`python -c "import os; print(max(os.path.getsize('${bam_file}') / 1024**3, 1) * ${time_limit})"`

This also removes the upper part, needed for file bam-file size calculation.

resolwe_bio/tools/bamtobigwig.sh
```diff
@@ -5,15 +5,36 @@ source /etc/profile.d/resolwe-base.sh
...
+
+timeout "${time_deeptools}" \
 bamCoverage \
```
JureZmrzlikar: indent bamCoverage command, since it is nested inside timeout command.

resolwe_bio/processes/alignment/bowtie.yml
```diff
@@ -278,12 +278,14 @@
       re-save species {{genome.species}}
       re-save build {{genome.build}}
 
-      bamtobigwig.sh "${FW_NAME}_mapped.bam" {{ requirements.resources.cores }} 50
-
       if [ -f "${FW_NAME}_unmapped.bam" ]; then
         re-save-file unmapped "${FW_NAME}_unmapped.fastq.gz"
       fi
 
+      #allowed time for bamCoverage is 480s
```
mzganec: Improve this comment, e.g. computation time limit for bamCoverage is 480 s per GB. Do the same in other processes.

PR #608, opened 2018-07-26
Refactor cutadapt and create workflows

JenkoB:
## Checklist

    before non-breaking changes.
    commit(s).

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -22,6 +22,7 @@ Added
   and paired-end reads ``workflow-cutadapt-star-featurecounts-paired``.
   Both workflows consist of preprocessing with Cutadapt, alignment
   with STAR two pass mode and quantification with featureCounts.
+- Add descriptor schema ``rna-seq-cutadapt-star-featurecounts``
 
 Changed
 -------
--- a/resolwe_bio/descriptors/cutadapt-star-featurecounts.yml
+++ b/resolwe_bio/descriptors/cutadapt-star-featurecounts.yml
@@ -1,0 +1,106 @@
+- slug: rna-seq-cutadapt-star-featurecounts
+  name: "Pipeline: Cutadapt - STAR (two pass mode) - featureCounts"
+  version: 0.0.1
+  description: "This RNA-Seq pipeline consists of three steps: preprocessing, alignment and  quantification.
+    \n\nFirstly, reads are preprocessed by __Cutadapt__ which by default drops reads that are shorter
+    than 50 bp and reads having more than 2 'N' bases, then it removes bases from 3' end with quality
+    below 20.\n\nSecondly, trimmed reads are aligned to genome by __STAR__ aligner using two-pass mode.
+    \n\nFinally, aligned reads are summarized to genes by __featureCounts__. Exons that are overlapping
+    to multiple genes are assigned to all of genes. Each overlapping gene receives a count of 1 / y,
+    where y is the total number of genes overlapping with the read."
+  schema:
+    - name: annotation
+      label: Source of annotation
+      type: basic:string
+      default: pgsc_itag
+      choices:
+        - label: PGSC/ITAG
+          value: pgsc_itag
+        - label: PGSC
+          value: pgsc
+        - label: ITAG
+          value: itag
+    - name: advanced
+      label: Show advanced options
+      type: basic:boolean
+      default: false
+      description: >
+        Inspect and modify parameters.
+    - name: trailing
+      label: Quality cutoff from 3' end
+      type: basic:integer
+      default: 20
+      description: >
+        Remove low quality bases from the 3' end. Specifies the minimum quality required to keep a base.
+      hidden: '!advanced'
+    - name: minlen
+      label: Min length
+      type: basic:integer
+      default: 50
+      description: >
+        Drop the read if it is below a specified length.
+      hidden: '!advanced'
+    - name: max_n
+      label: Max number of Ns in read.
+      type: basic:integer
+      default: 2
+      description: >
+        Discard reads having more 'N' bases than specified.
+      hidden: '!advanced'
+    - name: assay_type
+      label: Assay type
+      type: basic:string
+      default: non_specific
+      choices:
+        - label: Strand non-specific
+          value: non_specific
+        - label: Strand-specific forward
+          value: forward
+        - label: Strand-specific reverse
+          value: reverse
+      description: >
+        In strand non-specific assay a read is considered overlapping with a feature regardless
+        of whether it is mapped to the same or the opposite strand as the feature.
+        In strand-specific forward assay and single reads, the read has to be mapped to the
+        same strand as the feature. For paired-end reads, the first read has to be on the same
+        strand and the second read on the opposite strand. In strand-specific reverse
+        assay these rules are reversed.
+      hidden: '!advanced'
+    - name: frac_overlap
+      label: >
+        Minimum fraction of overlapping bases in a read that is required for read assignment
+      type: basic:decimal
+      default: 0.0
+      description: >
+        Value should be within range [0, 1]. Number of overlapping bases is counted from both
+        reads if paired end. Both this option and 'Minimum number of overlapping bases in a read
+        that is required for read assignment' need to be satisfied for read assignment.
+      hidden: '!advanced'
+    - name: allow_multi_overlap
+      label: Assign reads to all their overlapping features or meta-features
+      type: basic:boolean
+      default: true
+      hidden: '!advanced'
+    - name: count_multi_mapping_reads
+      label: Count multi-mapping reads
+      type: basic:boolean
+      default: false
+      description: >
+        For a multi-mapping read, all its reported alignments will be counted. The 'NH' tag in BAM
+        input is used to detect multi-mapping reads.
+      hidden: '!advanced'
+    - name: fraction
+      label: Assign fractional counts to features
+      type: basic:boolean
+      default: true
+      description: >
+        This option must be used together with 'Count multi-mapping reads' or 'Assign reads to
+        all their overlapping features or meta-features' or both. When 'Count multi-mapping reads'
+        is checked, each reported alignment from a multi-mapping read (identified via 'NH' tag)
+        will carry a count of 1 / x, instead of 1 (one), where x is the total number of
+        alignments reported for the same read. When 'Assign reads to all their overlapping
+        features or meta-features' is checked, each overlapping feature will receive a count
+        of 1 / y, where y is the total number of features overlapping with the read. When both
+        'Count multi-mapping reads' and 'Assign reads to all their overlapping features or
+        meta-features' are specified, each alignment will carry a count of 1 / (x * y).
+      hidden: '!advanced'


resolwe_bio/processes/reads_processing/cutadapt.yml
```diff
@@ -205,130 +216,131 @@
...
-      label: Error rate
-      type: basic:decimal
-      required: false
-      default: 0.1
-      description: >
-        Maximum allowed error rate (no. of errors divided by the length of the matching region).
+    - name: adapters
+      label: Adapters
+      group:
+        - name: up_primers_file_fwd
```
JureZmrzlikar: This is a matter of debate, but IMO, it would be better to rename this to `5_prime_mate_1`. Terms `up`, `down`, `forward` and `reverse` can be confusing.

resolwe_bio/processes/reads_processing/cutadapt.yml
```diff
@@ -205,130 +216,131 @@
...
+          type: basic:decimal
+          default: 0.1
+          description: >
+            Maximum allowed error rate (no. of errors divided by the length of the matching region).
+        - name: min_overlap
+          label: Minimal overlap
+          type: basic:integer
+          default: 3
+          description: >
+            Minimum overlap for an adapter match
```
JureZmrzlikar: Add fullstop.

resolwe_bio/processes/reads_processing/cutadapt.yml
```diff
@@ -205,130 +216,131 @@
...
+        - name: match_read_wildcards
+          label: Match read wildcards
+          type: basic:boolean
+          required: false
```
JureZmrzlikar: If there is a default value, `required: false` is not necessary.

resolwe_bio/processes/reads_processing/cutadapt.yml
```diff
@@ -205,130 +216,131 @@
...
+            Interpret IUPAC wildcards in reads.
+    - name: modify_reads
+      label: Modifying reads
+      group:
+        - name: leading
+          label: Leading quality
+          type: basic:integer
+          required: false
+          description: >
+            Remove low quality bases from the beginning. Specifies the minimum quality required to keep a base.
```
JureZmrzlikar: Consider joining the leading and trailg parameters in one, as it is originally done in CLI interface. Explain the difference between notation with and without a comma. Also, use "3/5 prime" notation, leading/trailing can be confusing.

resolwe_bio/processes/reads_processing/cutadapt.yml
```diff
@@ -205,130 +216,131 @@
...
+          required: false
+          description: >
+            Remove low quality bases from the end. Specifies the minimum quality required to keep a base.
+        - name: crop
+          label: Crop
+          type: basic:integer
+          required: false
+          description: >
+            Cut the specified number of bases from the end of the reads
+        - name: headcrop
```
JureZmrzlikar: Consider joining this with `crop` parmeter as it is done in CL interface. Explain the impact of `-` sign.

resolwe_bio/processes/reads_processing/cutadapt.yml
```diff
@@ -205,130 +216,131 @@
...
+          type: basic:integer
+          required: false
+          description: >
+            Remove low quality bases from the end. Specifies the minimum quality required to keep a base.
+        - name: crop
+          label: Crop
+          type: basic:integer
+          required: false
+          description: >
+            Cut the specified number of bases from the end of the reads
```
JureZmrzlikar: Add fulstop.

resolwe_bio/processes/reads_processing/cutadapt.yml
```diff
@@ -205,130 +216,131 @@
...
+            Cut the specified number of bases from the start of the reads
+    - name: filtering
+      label: Filtering of processed reads
+      group:
+        - name: minlen
+          label: Min length
+          type: basic:integer
+          required: false
+          description: >
+            Drop the read if it is below a specified length
```
JureZmrzlikar: Add fullstop.

resolwe_bio/processes/reads_processing/cutadapt.yml
```diff
@@ -205,130 +216,131 @@
...
+          required: false
+          description: >
+            Discard reads having more 'N' bases than specified.
+        - name: pair_filter
+          label: Which of the reads have to match the filtering criterion
+          type: basic:string
+          default: any
+          description: >
+            Which of the reads in a paired-end read have to match the filtering criterion in order for the pair to be
+            filtered. Possible options are 'any' or 'both'.
```
JureZmrzlikar: You can enforce usage of the two options by using the `choices` notation.

resolwe_bio/processes/reads_processing/cutadapt.yml
```diff
@@ -400,78 +410,55 @@
...
-        {% if down_primers_seq_fwd %}$down_primers_fwd{% endif %} \
-        {% if up_primers_seq_fwd %}$up_primers_fwd{% endif %} \
-        {% if down_primers_file_fwd %}-a file:3_primers_fwd.fasta{% endif %} \
-        {% if up_primers_file_fwd %}-g file:5_primers_fwd.fasta{% endif %} \
-        {% if times %}--times {{ times }} {% endif %} \
-        {% if match_read_wildcards %}--match-read-wildcards {% endif %} \
-        {% if error_rate %}--error-rate {{ error_rate }} {% endif %} \
-        -O {{min_overlap_fwd}} \
-        -o "${NAME1}_trimmed.fastq.gz" in_fw_reads.fastq.gz > report.txt
+        {% if filtering.minlen %}-m {{ filtering.minlen }}{% endif %} \
```
JureZmrzlikar: Usage of short option e.g `-m` vs. `--minimum-length` is inconsitent. I would suggest to always use longer notation because it improves readability.

resolwe_bio/processes/reads_processing/cutadapt.yml
```diff
@@ -400,78 +410,55 @@
...
+        {% if modify_reads.headcrop %}-u {{ modify_reads.headcrop }}{% endif %} \
+        {% if modify_reads.crop %}-u -{{ modify_reads.crop }}{% endif %} \
+        {% if modify_reads.leading %}-q {{ modify_reads.leading }}{% endif %} \
+        {% if modify_reads.trailing %}-q {{ modify_reads.trailing }},0{% endif %} \
```
JureZmrzlikar: I think that description of parameters is misleading / implementation is wrong. From the cutadapt manual::

    -q [5'CUTOFF,]3'CUTOFF, --quality-cutoff=[5'CUTOFF,]3'CUTOFF
    Trim low-quality bases from 5' and/or 3' ends of each read before adapter
    removal. Applied to both reads if data is paired. If one value  is  given,  only
    the  3'  end  is  trimmed.  If  two comma-separated cutoffs are given, the 5'
    end is trimmed with the first cutoff, the 3' end with the second.

The roles of leading / trailing shouod be reversed.

resolwe_bio/processes/reads_processing/cutadapt.yml
JenkoB: I agree

resolwe_bio/processes/reads_processing/cutadapt.yml
JenkoB: I think it as it is might be more clear to biologist user and not too confusing for bioinformaticians. However I agree about improving ``label`` notation. I will leave it as it is and correct it when we figure out user stories.

resolwe_bio/processes/reads_processing/cutadapt.yml
JenkoB: Same as above

resolwe_bio/processes/reads_processing/cutadapt.yml
JenkoB: Thanks for noticing this!

resolwe_bio/processes/reads_processing/cutadapt.yml
JenkoB: I have applied the suggestion only in `label` filed, because we don't support numbers in `name`.

JenkoB:
All suggestions were taking into account. Please do another round of reviews.

resolwe_bio/processes/reads_processing/cutadapt.yml
```diff
@@ -205,130 +215,135 @@
...
+          type: list:basic:string
+          required: false
+        - name: down_primers_seq_rev
+          label: 3 prime adapter sequence for Mate 2
+          type: list:basic:string
+          required: false
+        - name: times
+          label: Times
+          type: basic:integer
+          required: false
```
JureZmrzlikar: If default is provided, `required: false` is redundant.

resolwe_bio/processes/reads_processing/cutadapt.yml
```diff
@@ -396,82 +406,55 @@
...
+        {% if modify_reads.leading %}--quality-cutoff={{ modify_reads.leading }},0{% endif %} \
+        {% if adapters.down_primers_seq_fwd %}$down_primers_fwd{% endif %} \
+        {% if adapters.up_primers_seq_fwd %}$up_primers_fwd{% endif %} \
+        {% if adapters.down_primers_file_fwd %}-a file:3_primers_fwd.fasta{% endif %} \
+        {% if adapters.up_primers_file_fwd %}-g file:5_primers_fwd.fasta{% endif %} \
+        {% if adapters.down_primers_seq_rev %}$down_primers_rev{% endif %} \
+        {% if adapters.up_primers_seq_rev %}$up_primers_rev{% endif %} \
+        {% if adapters.down_primers_file_rev %}-A file:3_primers_rev.fasta{% endif %} \
+        {% if adapters.up_primers_file_rev %}-G file:5_primers_rev.fasta{% endif %} \
+        {% if adapters.match_read_wildcards %}--match-read-wildcards {% endif %} \
```
JureZmrzlikar: This parameter is not optional and has a default value. Remove the if-cluase and reformat into form similar to `--pair-filter` / `--error-rate`, `--times`, ...

resolwe_bio/processes/reads_processing/cutadapt.yml
JureZmrzlikar: You could use `mate1_5prime` or something similar... It would be more in sync with the label. But it is also fine as it is now.

resolwe_bio/processes/reads_processing/cutadapt.yml
JenkoB: because the `type: basic:boolean` it should stay like it is.

resolwe_bio/processes/reads_processing/cutadapt.yml
```diff
@@ -165,8 +175,8 @@
       cutadapt \
         {% if headcrop %}-u {{headcrop}}{% endif %} \
         {% if crop %}-u -{{crop}}{% endif %} \
-        {% if leading %}-q {{leading}}{% endif %} \
-        {% if trailing %}-q {{trailing}},0{% endif %} \
+        {% if trailing %}-q {{ trailing }}{% endif %} \
```
jkokosar: @JenkoB @mzganec @JureZmrzlikar For the numerical inputs, the existance of the variable should be tested using `{% if var is defined %}`. If the user inputs 0, the `{% if var %}` evaluates to `False`, which skips the use of the parameter flag in the process altogether. This is an issue in (too) many places, not just in the process being worked on here.

PR #627, opened 2018-08-21
[WIP] Add super-sample descriptor schema

mzganec:
## Checklist

    before non-breaking changes.
    commit(s).

### Processes ###

  can be performed).
  and publication.

  tests).

### Tests ###


resolwe_bio/descriptors/super_sample.yml
```diff
@@ -0,0 +1,263 @@
+- slug: super_sample
```
mstajdohar: This should be just `sample` or `sample2` or `sample-annotation-v2`.

resolwe_bio/descriptors/super_sample.yml
```diff
@@ -0,0 +1,263 @@
...
+  name: Super sample annotation
```
mstajdohar: I'd call this Sample annotation.

resolwe_bio/descriptors/super_sample.yml
```diff
@@ -0,0 +1,263 @@
...
+  description: Sample meta data compliant with GEO and ENCODE databases.
+  version: 0.0.1
```
mstajdohar: Let's make the first one 1.0.0, as it then looks more production ready to the end-users.

mstajdohar:
When can we expect the reads schema?

resolwe_bio/descriptors/super_sample.yml
```diff
@@ -0,0 +1,263 @@
...
+  schema:
+    - name: assay_type
+      label: Assay type
+      type: basic:string
+      choices:
```
jkokosar: If this is to be the new default descriptor schema for sample object, this list is not comprehensive enough. Please see this [example](https://github.com/genialis/resolwe-bio/blob/master/resolwe_bio/descriptors/reads.yml#L12) of the available choices in the GEO system. Explore also the latest and most up-to-date annotation $ submission template that can be dowloaded from GEO site. Use the `allow_custom_choice: true` for the choices menus where this is applicable.

resolwe_bio/descriptors/super_sample.yml
```diff
@@ -0,0 +1,263 @@
...
+        - label: RNA-seq
+          value: rna-seq
+        - label: ChIP-seq
+          value: chip-seq
+        - label: ATAC-seq
+          value: atac-seq
+        - label: ChIPmentation
+          value: chipmentation
+    - name: biosample_type
```
jkokosar: This choices list does not allow for proper classification of other possible types of samples, most notably microorganism cultures, but also other possible types that we might encounter down the road (e.g. metagenomic samples).

resolwe_bio/descriptors/super_sample.yml
```diff
@@ -0,0 +1,263 @@
...
+            - label: RNA binding protein
+              value: rna_binding_protein
+            - label: Chromatin remodeller
+              value: chromatin_remodeller
+            - label: Nucleotide modification
+              value: nucleotide_modification
+        - name: antibody_manufacturer
+          label: Antibody manufacturer
+          type: basic:string
+          choices:
```
jkokosar: This manufacturers list might be the most common, but its also too restrictive and without the `allow_custom_choice` option. There are also other manufacturers, some of them quite [big](https://www.sigmaaldrich.com/life-science/cell-biology/antibodies/antibody-products.html?TablePage=108208823)

resolwe_bio/descriptors/super_sample.yml
```diff
@@ -0,0 +1,263 @@
...
+        - name: antibody_target
+          label: Antibody target
+          type: basic:string
+        - name: antibody_catalog_number
+          label: Antibody catalog number
+          type: basic:string
+        - name: antibody_lot_number
+          label: Antibody lot number
+          type: basic:string
+    - name: organism
```
jkokosar: Consider changing this to `species`. 

Too restrictive list without the `allow_custom_choice` option. The format of the `value` fields as envisioned will break everything that relies on the species name to be in binomial nomenclature. This affects gene knowlege-base, sample basket, existing data objects that will not match the sample basket restrictions (species field on expressions being `Homo sapiens` will not match `homo_sapiens` information on the samples.

resolwe_bio/descriptors/super_sample.yml
```diff
@@ -0,0 +1,263 @@
...
+    - name: genotype
+      label: Genotype
+      type: basic:string
+    - name: biosample_treatment
+      label: Biosample treatment
+      type: basic:string
+    - name: treatment_protocol
+      label: Treatment protocol
+      type: basic:string
+    - name: platform
```
jkokosar: Too restrictive list without the `allow_custom_choice` option.

resolwe_bio/descriptors/super_sample.yml
```diff
@@ -0,0 +1,263 @@
...
+          value: protein
+        - label: Other
+          value: other
+    - name: library_depleted_in
+      label: Library depleted in
+      type: basic:string
+    - name: library_treatment
+      label: Library treatment
+      type: basic:string
+    - name: replication_type
```
jkokosar: What is this field for? In general, I think we lack `description` fields in this schema.

jkokosar:
This schema does not allow for any custom (text) fields. Also, it lacks some commonly used meta-data fields like `gender` and `age`.  We should allow for some optional/general fields to cover cases where descriptor schema fields as envisioned here will turn out to not comprehensive enough.

If the new schema would include other [custom fields that are part of other descriptor schemas](https://github.com/genialis/resolwe-bio/blob/master/resolwe_bio/descriptors/sample_detailed.yml#L76), we could make those schemas obsolete and remove them entirely.

resolwe_bio/descriptors/super_sample.yml
mzganec: We added the option `Other` instead of custom user input.

mzganec:
@jkokosar Please have a look.

resolwe_bio/descriptors/sample2.yml
```diff
@@ -0,0 +1,387 @@
...
+        - name: library_treatment
+          label: Library treatment
+          type: basic:string
+          description: Describe how your library was treated.
+          required: false
+        - name: replication_type
+          label: Replication type
+          type: basic:string
+          description: >
+            Select what replication type you used, for more info, go to
```
jkokosar: Use hyperlinks

resolwe_bio/descriptors/sample2.yml
```diff
@@ -0,0 +1,387 @@
...
+              value: tissue
+            - label: Primary cell
+              value: primary_cell
+            - label: Other
+              value: other
+        - name: primary_cell_line
+          label: Primary cell line
+          type: basic:string
+          description: >
+            Search for your tissue type: http://www.ontobee.org/ontology/UBERON and enter the UBERON
```
jkokosar: Use hyperlinks instead

resolwe_bio/descriptors/sample2.yml
```diff
@@ -0,0 +1,387 @@
...
+          type: basic:string
+          description: >
+            Genotype of your sample, e.g. P53+/-, Tg2576, etc.
+          required: false
+        - name: biosample_type
+          label: Biosample type
+          type: basic:string
+          required: false
+          description: >
+            This is where you put what your samples are derived from, e.g. cell line, tissue,
```
jkokosar: This is where you put what your samples are derived.. Maybe it would be better "The source of samples/biological material, e.g. ..."

resolwe_bio/descriptors/sample2.yml
mzganec: We don't support hyperlinks in description of inputs.

PR #629, opened 2018-08-30
Add MultiQC and Seqtk to the resolwebio/utils Docker image

jkokosar:
* MultiQC
* Seqtk

## Checklist

    before non-breaking changes.
    commit(s).

### Processes ###

  can be performed).
  and publication.

  tests).

### Tests ###


resolwe_bio/docker_images/utils/packages-python3.txt
```diff
@@ -4,6 +4,7 @@ pandas==0.23.0 --hash=sha256:29960ea1783c59fe86b323799b130592149e0f7bb739149e7fc
 requests==2.18.4 --hash=sha256:6a1b267aa90cac58ac3a765d067950e7dbbf75b1da07e895d1f594193a40a38b
 resdk==4.0.0 --hash=sha256:19b5409e9ac4e68fdd9fa8f15b295c1eb417842fa72fb605b143390451e7fedf
 scipy==1.1.0 --hash=sha256:729f8f8363d32cebcb946de278324ab43d28096f36593be6281ca1ee86ce6559
+multiqc==1.6.0 --hash=sha256:1e9f0480b7f6590a4ac159f69c1ee6c04aed0e3c156538b4994e3df451a6244e
```
JureZmrzlikar: You could put this entry a few rows higher, to keep alphabetical ordering. Not urgent.

resolwe_bio/docker_images/utils/packages-python3.txt
```diff
@@ -24,3 +25,20 @@ urllib3==1.22 --hash=sha256:06330f386d6e4b195fbfc736b297f58c5a892e4440e54d294d70
 numpy==1.14.4 --hash=sha256:f29a9c5607b0fded7a9f0871dbd06918a88cb0a465acfac5c67f92d1a4115d48
 python-dateutil==2.6.1 --hash=sha256:95511bae634d69bc7329ba55e646499a842bc4ec342ad54a8cdb65645a0aad3c
 pytz==2017.2 --hash=sha256:d1d6729c85acea5423671382868627129432fba9a89ecbb248d8d1c7a9f01c67
+
+# Multiqc dependencies.
+click==6.7 --hash=sha256:29f99fc6125fbc931b758dc053b3114e55c77a6e4c6c3a2674a2dc986016381d
```
JureZmrzlikar: Also here: order alphabetically, if possible.

PR #635, opened 2018-09-03
Add tools for genome browsers to rnaseq docker

JenkoB:
## Checklist

    before non-breaking changes.
    commit(s).

### DIFF
--- a/resolwe_bio/docker_images/rnaseq/packages-manual/bedtobigbed.sh
+++ b/resolwe_bio/docker_images/rnaseq/packages-manual/bedtobigbed.sh
@@ -1,0 +1,40 @@
+#!/bin/bash -e
+
+#
+# Licensing conditions stated in http://hgdownload.soe.ucsc.edu/admin/exe/
+# ========================================================================
+#
+# (updated Mar 29, 2018)
+#
+# The UCSC Genome Browser and Blat software are free for academic, nonprofit, and personal use. A
+# license is required for commercial download and installation of these binaries, with the exception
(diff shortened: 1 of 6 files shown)


mstajdohar:
Rebase

resolwe_bio/docker_images/rnaseq/packages-manual/bedtobigbed.sh
```diff
@@ -0,0 +1,42 @@
...
+. /var/cache/build/packages-manual/common.sh
+
+download_and_verify \
+    kent \
+    bedToBigBed \
+    kent-v365 \
+    a66027e1a51f803fef6bf234b90d0e4d6deee6027acfe2c0a65712f3157cca5d \
+    https://drive.google.com/uc?id=1N2pUaJuHgCUS4gztTcoybrRWw1m8WNtv#.elf
+
+wget https://raw.githubusercontent.com/ENCODE-DCC/kentUtils/master/src/hg/lib/encode/narrowPeak.as
```
JureZmrzlikar: Please explain why do we first pull something from google drive and then also perform wget. Are there two programs to be installed? If there are, they should be put in separate commits...

JenkoB: This is a short file that is needed by this program in some cases. For this case is actually redundant (it was important for chipseq), so I will delete it. Thanks

JenkoB:
@JureZmrzlikar @mzganec please review.

PR #650, opened 2018-09-25
Set new default sample descriptor schema

mzganec:
Change slug from `sample2` to `sample`, modify group names, and remove the original `sample` and `sample-detailed` descriptor schemas.


mzganec:
Fixed failing expression aggregator test.

PR #655, opened 2018-10-03
Add novel splice junctions in BED12 format output

JenkoB:
## Checklist

    before non-breaking changes.
    commit(s).

### DIFF
--- a/resolwe_bio/tools/novel_splice_junctions.py
+++ b/resolwe_bio/tools/novel_splice_junctions.py
@@ -1,0 +1,31 @@
+#!/usr/bin/env python3
+"""Filter novel splice junctions in BED12 format."""
+
+import argparse
+import os
+import sys
+
+import pandas as pd
+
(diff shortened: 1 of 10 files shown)


resolwe_bio/tools/novel_splice_junctions.py
```diff
@@ -0,0 +1,32 @@
...
+
+    if os.stat(bed_file).st_size == 0:
+        msg = 'Bed file has no entries.'
+        print(warning(msg))
+        os.rename(bed_file, 'novel_sj.bed')
+        sys.exit(0)
+
+    bed = pd.read_csv(args.bed_file, delimiter='\t', header=None, dtype=str)
+    novel_sj = pd.read_csv(args.novel_sj, delimiter='\t', dtype=str)
+    bed_novel_sj = bed[bed[3].isin(novel_sj["name"])]
```
JureZmrzlikar: Nice!

PR #659, opened 2018-10-09
WIP: Increase test timeout to 90 minutes

mstajdohar:
Testing the hypothesis that `resolwe-bio` tests fail simply because the runtime has increased (and not because of some other error, e.g. the semaphore issue).


mstajdohar:
We'll raising the timeout did not help. @jberci Please check if it's semaphore resurrection.

PR #665, opened 2018-10-12
Remove obsolete processes

jkokosar:
## Checklist

    before non-breaking changes.
    commit(s).

### Processes ###

  can be performed).
  and publication.

  tests).

### Tests ###


resolwe_bio/processes/workflows/amplicon.yml
```diff
@@ -3,7 +3,7 @@
   data_name: "{{ reads|sample_name|default('?') }}"
   requirements:
     expression-engine: jinja
-  version: 3.0.1
+  version: 3.1.0
```
JureZmrzlikar: IMO, removing a process from workflow is equal to removing an output from the process, which is a breaking change. Raise version accordingly and prepend changelog entry with "backward-incompatible"...

resolwe_bio/tests/processes/test_expression.py
```diff
@@ -18,13 +18,13 @@ def test_cufflinks(self):
...
 
-            inputs = {
+            aligned_reads = self.run_process('alignment-hisat2', {
                 'genome': genome.pk,
                 'reads': reads.pk,
-                'annotation': annotation_gtf.pk,
-                'PE_options': {
-                    'library_type': "fr-unstranded"}}
-            aligned_reads = self.run_process('alignment-tophat2', inputs)
+                'spliced_alignments': {
```
JureZmrzlikar: How is this change connected to the process removal?

resolwe_bio/tests/workflows/test_heat_seq.py
```diff
@@ -1,53 +0,0 @@
...
-    @tag_process('workflow-heat-seq')
-    def test_heatseq_workflow(self):
-        with self.preparation_stage():
-            inputs = {
-                'src1': ['heat_seq_mate1.fq.gz'],
-                'src2': ['heat_seq_mate2.fq.gz']}
-            reads = self.run_process('upload-fastq-paired', inputs)
-
-            inputs = {
-                'src': 'chr21_small.fasta.gz',
```
JureZmrzlikar: Also remove this file.

resolwe_bio/tests/processes/test_expression.py
jkokosar: `alignment-tophat2` was removed as a process and was used in a few of the tests. `alignment-hisat2` is more modern replacement aligner to be used in this workflows, so this replacement was made.

docs/CHANGELOG.rst
```diff
@@ -12,6 +12,18 @@ Unreleased
 
 Changed
 -------
+- **BACKWARD INCOMPATIBLE:** Remove obsolete processes: ``bsmap``,
```
JenkoB: Great cleaning!! Did you remove ``wgbs.yml`` workflow? It should be also removed and stated in changelog.

PR #666, opened 2018-10-15
Handle missing gene expression in hierarchical clustering of genes

mzganec:
    before non-breaking changes.
    commit(s).
* [X] All output fields have their ``re-save`` calls.


mstajdohar:
Included in https://github.com/genialis/resolwe-bio/pull/670

PR #675, opened 2018-10-24
Add bedGraphToBigWig to bamtobigwig script

JenkoB:
## Checklist

    before non-breaking changes.
    commit(s).


resolwe_bio/tools/bamtobigwig.sh
```diff
@@ -16,29 +17,56 @@ fi
...
 
-exit_status="$?"
+  exit_status="$?"
+
+  if [ "${exit_status}" == 124 ]
+  then
+    re-warning "BigWig took too long to process."
+  elif [ "${exit_status}" == 1 ]
+  then
+    re-checkrc "Calculation of BigWig failed."
```
mzganec: `re-checkrc` is not intuitive here. Either use `re-error` or use `re-checkrc` outside of the `if` statement.

It seems to me that the status of the data object should be the same, if BigWig times out or if it fails. If we consider BigWig to be a less important output, we can issue a warning. If we find it important, we can raise an error. But in my view we should do the same thing in both timeout and fail.

resolwe_bio/processes/amplicon/bwa_trim.yml
```diff
@@ -9,12 +9,12 @@
     expression-engine: jinja
     executor:
       docker:
-        image: resolwebio/dnaseq:3.3.0
+        image: resolwebio/dnaseq:bw1
```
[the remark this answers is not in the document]
JenkoB: Local :) , I am waiting for Janez's PR

resolwe_bio/tools/bamtobigwig.sh
JenkoB: Thanks Matjaz. I will put `re-warning` in both cases. I have just noticed that the python script ``bigwig_chroms_to_ucsc.py`` will fail in case of warnings because there will be empty input for `infile`. I need to restructure the script.

JenkoB:
@JureZmrzlikar please review, second commit is new

resolwe_bio/tools/check_bam_source.py
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/python3
+"""Check if the source of RNA ae rRNA or globin RNA."""
+
+import argparse
+import os
+import sys
+
+MAPPINGS_DIR = '/opt/chrom_mappings/globin_rrna/'
```
JureZmrzlikar: `GLOBIN_RRNA_DIR`

resolwe_bio/tools/check_bam_source.py
```diff
@@ -0,0 +1,72 @@
...
+QC_FILES = {
```
JureZmrzlikar: These are definitely not QC files. I suggest naming this `GLOBIN_RRNA_FILES` and add a comment "These are files that contain names of globin and rRNA sequences"

resolwe_bio/tools/check_bam_source.py
```diff
@@ -0,0 +1,72 @@
...
+def parse_arguments():
+    """Parse command line arguments."""
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument('--chromosome_names', help="File with chromosome names")
+    parser.add_argument('--species', help="Species")
+    return parser.parse_args()
+
+
+def parse_file(filename):
+    """Parse qc files."""
```
JureZmrzlikar: Parse file with chromosome names.

resolwe_bio/tools/check_bam_source.py
```diff
@@ -0,0 +1,72 @@
...
+    qc_file = set()
```
JureZmrzlikar: This is neither a file, nor it contains a qc sequences. These are chromosome names. Name it `chromosomes` or something `chromosome_names`

resolwe_bio/tools/check_bam_source.py
```diff
@@ -0,0 +1,72 @@
...
+    with open(filename) as handle:
+        for line in handle:
+            line = line.strip().split()
```
JureZmrzlikar: If chromosome name will include spaces this will fail. Rather use `split('\t')`

resolwe_bio/tools/check_bam_source.py
```diff
@@ -0,0 +1,72 @@
...
+            qc_file.add(line[0])
+    return qc_file
+
+
+def parse_qc(species):
+    """Parse file with chromosome names for globin RNA and rRNA."""
```
JureZmrzlikar: Get names of globin RNA and rRNA sequences.

resolwe_bio/tools/check_bam_source.py
```diff
@@ -0,0 +1,72 @@
...
+    qc_file = set()
```
JureZmrzlikar: This si not a file. Find a more descriptive variable.

resolwe_bio/tools/check_bam_source.py
```diff
@@ -0,0 +1,72 @@
...
+
+def check_rna_source(chrom_names, qc_files):
+    """
+    Check if the source of RNA ae rRNA or globin RNA.
+
+    If chromosome names are subset of rRNA or globin RNA chromosome names,
+    exit with status 0 since you don't need bigwig for this file.
+    """
+    chrom_names = parse_file(chrom_names)
+    if chrom_names.issubset(qc_files):
```
JureZmrzlikar: Rename variables:
`if chromosomes.issubset(globin_rrna_sequences):`
`...`
This make's it much clearer what is actually happening and makes docstring almost redundant.

PR #680, opened 2018-11-06
Fix tests

mstajdohar:


docs/CHANGELOG.rst
```diff
@@ -43,6 +43,7 @@ Added
 
 Fixed
 -----
+- Dummy fix
```
dblenkus: Change this.

resolwe_bio/expression_filters/relation.py
```diff
@@ -8,9 +8,7 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 from resolwe.flow.expression_engines.jinja.filters import id_
-from resolwe.flow.models.entity import RelationPartition
-
-from resolwe_bio.models import Sample
+from resolwe.flow.models.entity import Entity, RelationPartition
```
dblenkus: Is this part of the fix?

mstajdohar: Not really, but using Sample is ugly.

PR #683, opened 2018-11-06
Handle empty gene name in expression files in PCA

mzganec:


resolwe_bio/tools/pca.py
```diff
@@ -94,11 +94,19 @@ def save_pca(result={}, sample_ids=[], output_fn=None, max_size=10):
...
+        1: float,
     }
-    csv = pd.read_csv(fname, sep='\t', header=0, dtype=dtype, low_memory=False)
-    return csv.set_index('Gene')
+    csv = pd.read_csv(
+        filepath_or_buffer=fname,
+        sep='\t',
+        header=0,
+        index_col=0,
+        dtype=dtype,
```
JureZmrzlikar: Any reason why you define variable `dtype` and use it below instead of just using the dictionary directly?

resolwe_bio/tests/processes/test_pca.py
```diff
@@ -58,6 +72,16 @@ def test_pca(self):
...
 
+        inputs = {
+            'exps': [
+                expression_noname_1.pk,
+                expression_noname_2.pk,
+            ],
+            'source': 'DICTYBASE',
+            'species': 'Dictyostelium discoideum',
+        }
+        pca = self.run_process('pca', inputs)
```
JureZmrzlikar: Is the fact that process completes without error a sign of success or do you want to make some additional assertions? For example asserting there are no warnings, etc. I am fine with the way it is, but you know best weather this is bulletproof...

PR #693, opened 2018-11-20
Release 15

mstajdohar:


dblenkus:
And there are some unnecessary change log entries, like fixing tests, etc.

docs/CHANGELOG.rst
```diff
@@ -5,6 +5,19 @@ Change Log
 All notable changes to this project are documented in this file.
 This project adheres to `Semantic Versioning <http://semver.org/>`_.
 
+==========
```
dblenkus: There should be 2 empty lines above.

mstajdohar:
> there is inconsistent use of dots at the end of bullet points

Dots are used if there are more then 2 sentences in the bullet line. Is this OK?

mstajdohar:
Should we always use dots?

mstajdohar:
> Few lines in the change log are longer than 79 characters and there is inconsistent use of dots at the end of bullet points.

Fixed

> And there are some unnecessary change log entries, like fixing tests, etc.

Fixed

> There should be 2 empty lines above.

Fixed

dblenkus:
> Should we always use dots?

I think we should never use them.

> Dots are used if there are more then 2 sentences in the bullet line. Is this OK?

I agree if multiple sentences are inevitable. In some cases they can be easily reworded in to a single sentence, e.g. ```Add ``resolwebio/snpeff`` Docker image. The image includes SnpEff (4.3K) tool.```

PR #702, opened 2018-12-03
Add ERCC table to docker

JenkoB:
## Checklist

    before non-breaking changes.
    commit(s).

### Processes ###

  can be performed).
  and publication.

  tests).

### Tests ###


JenkoB:
@JureZmrzlikar

docs/CHANGELOG.rst
```diff
@@ -38,6 +38,7 @@ Added
 - Add ``background_pairs`` Jinja expressions filter that accepts a list of
   data objects and orders them in a list of pairs (case, background) based on
   the background relation between corresponding samples.
+- Add ERCC table to ``resolwebio/rnaseq`` docker image.
```
JureZmrzlikar: Add version of docker image: it will be `4.1.0`. Please put exact version also in the commit message.

JureZmrzlikar:
@mzganec There are some additional commits. Please make another review pass.

resolwe_bio/docker_images/rnaseq/packages-manual/ercc.sh
```diff
@@ -0,0 +1,15 @@
...
+# Original file is available at https://tools.thermofisher.com/content/sfs/manuals/cms_095046.txt
+
+download_and_verify \
+    resolwebio \
+    assets \
+    ERCC_table \
+    5151b394332477301851404d37b907cb6f27c9aaa2569e288531a839d6ca2b3b \
+    https://drive.google.com/uc?id=193PpxSoNhFwYUpTMh9nGt44K72tc0F9f \
+    '' \
+    ERCC_table.txt
```
mstajdohar: Would be nice to have a more informative file name.

resolwe_bio/docker_images/rnaseq/Dockerfile
```diff
@@ -1,3 +1,6 @@
 FROM docker.io/resolwebio/common:1.0.0
 
 MAINTAINER Resolwe Bioinformatics authors https://github.com/genialis/resolwe-bio
+
+# Force matplotlib backend to a headless one.
+ENV MPLBACKEND Agg
```
mstajdohar: I suggest we add this to the common image where the matplotlib is installed.

mstajdohar: Why not? Add it to common and we'll build and publish common:1.0.1 which we can reference here.

mstajdohar: you can do both in the same PR

mstajdohar:
You need to whitelist some of the files in setup.py/MANIFEST.in. See the testing report.

resolwe_bio/docker_images/rnaseq/packages-manual/ercc.sh
JureZmrzlikar: How do you like `ERCC_spikein_concentrations.txt`?

mstajdohar: I was thinking more in line of mentioning Thermofisher or the type of kits using this... Just cause there are more and more files on the Google Drive and we need to be able to know what's what. Anyway, it's not super important and I was not sure whether I should point it out at all or not. Your call how you handle it.

resolwe_bio/docker_images/rnaseq/Dockerfile
JureZmrzlikar: Good point, will do.

resolwe_bio/docker_images/rnaseq/packages-manual/ercc.sh
JureZmrzlikar: Let's just leave it for now. Files belonging in Docker images are quite well separated and this is the first on regarding ERCC. If and when we have more of them we should have a better idea how to rename/organize them.

resolwe_bio/docker_images/rnaseq/assets/genialis.css
```diff
@@ -0,0 +1,135 @@
+body {
+  font-family: 'Roboto';
+  font-size: 12px;
+  font-style: normal;
+  color: #212121;
+  background-color: #fff;
+  margin: 2em 2% 4em 2%;
+  width: 80%!important;
```
lukaw3d: I would remove `width` and just update `margin` with it's final value `2em 18% 4em 2%`

resolwe_bio/docker_images/rnaseq/assets/genialis.css
```diff
@@ -0,0 +1,135 @@
...
+  text-align: justified;
```
lukaw3d: `justify`
that will fix it

resolwe_bio/docker_images/rnaseq/assets/genialis.css
```diff
@@ -0,0 +1,135 @@
...
+}
+
+
+
+/* roboto-300 - latin-ext_latin */
+@font-face {
+  font-family: 'Roboto';
+  font-style: normal;
+  font-weight: 300;
+color: green;
```
lukaw3d: This got lost :D

resolwe_bio/docker_images/rnaseq/assets/genialis.css
```diff
@@ -0,0 +1,135 @@
...
+  font-family: 'Roboto';
+  font-style: normal;
+  font-weight: 500;
+  src: url('fonts/roboto-v16-latin-ext_latin-500.eot'); /* IE9 Compat Modes */
+  src: local('Roboto Medium'), local('Roboto-Medium'),
+       url('fonts/roboto-v16-latin-ext_latin-500.eot?#iefix') format('embedded-opentype'), /* IE6-IE8 */
+       url('fonts/roboto-v16-latin-ext_latin-500.woff2') format('woff2'), /* Super Modern Browsers */
+       url('fonts/roboto-v16-latin-ext_latin-500.woff') format('woff'), /* Modern Browsers */
+       url('fonts/roboto-v16-latin-ext_latin-500.ttf') format('truetype'), /* Safari, Android, iOS */
+       url('fonts/roboto-v16-latin-ext_latin-500.svg#Roboto') format('svg'); /* Legacy iOS */
```
lukaw3d: My browser console has an issue with parsing some `*-500` files.
Can you check your console / re-download them from google / re-commit

resolwe_bio/docker_images/rnaseq/assets/spikeins_qc_template.html
```diff
@@ -0,0 +1,68 @@
+<!DOCTYPE html>
```
lukaw3d: Will these files be downloaded as zip usually, or will they be viewed directly on website?

mstajdohar: Ideally we should be able to open them in a separate tab. Should we change the `report` output  type from `basic:file` to `basic:url:view`?

mstajdohar: @jkokosar ?

jkokosar: Opening this kind of reports in separate tab is always preferred over downloading.

@mstajdohar @JureZmrzlikar I think the type for the output field should be `basic:file:html`

resolwe_bio/docker_images/rnaseq/assets/genialis.css
JureZmrzlikar: The issue with *.500 files is fixed, thanks.

resolwe_bio/docker_images/rnaseq/assets/spikeins_qc_template.html
lukaw3d: Discussion from daily:
- direct view link in `basic:file:html` is fine
- direct download link in `basic:file`/`basic:file:html` will **break external** css, fonts, and images
- downloading all files as zip should work

Workaround
- we can add a zipped output
  - can download when viewing a Sample or Data object
  - can't view in new tab
- we can add a hidden `basic:file:html` output and add `Open report` column in Sample Table
  - can open in new tab

PR #708, opened 2018-12-13
Add feature symbols to expressions in archive samples process

mzganec:


resolwe_bio/processes/support_processors/archive.yml
```diff
@@ -5,8 +5,8 @@
     expression-engine: jinja
     executor:
       docker:
-        image: resolwebio/utils:1.2.0
-  version: 0.1.5
+        image: resolwebio/utils:1.5.0
```
jkokosar: Rather use `resolwebio/common:1.0.0` image

PR #712, opened 2019-01-04
Validate the input FASTQ files

jkokosar:
Requires https://github.com/genialis/resolwe-bio/pull/711


resolwe_bio/tools/validate_fastq.py
```diff
@@ -0,0 +1,95 @@
...
+        print(error('The number of mate-pair files in split-lane samples must match. '
+                    '{} and {} input files were given for the -fq and -fq2 inputs, '
+                    'respectively.'.format(len(args.fq), len(args.fq2))))
+        exit(1)
+
+    if args.fq2:
+        for mate1, mate2 in zip(args.fq, args.fq2):
+            try:
+                paired_reads = dnaio.open(mate1, file2=mate2, fileformat='fastq')
+                if not any(True for r1, r2 in zip(paired_reads.reader1._iter, paired_reads.reader2._iter)):
```
mzganec: Simplify.

resolwe_bio/tools/validate_fastq.py
```diff
@@ -0,0 +1,95 @@
...
+                        )
+                    )
+                else:
+                    for read in paired_reads:
+                        continue
+                    print('Successfully validated mate-pair files {} and {}.'.format(basename(mate1), basename(mate2)))
+
+            except (FastqFormatError, FileFormatError) as dnaio_error:
+                print(error(
+                    'Error in mate-pairs {} and {}. {}'.format(basename(mate1), basename(mate2), str(dnaio_error))
```
mzganec: Add the error type. I suggest _Format error in mate-pairs ..._

resolwe_bio/tools/validate_fastq.py
JureZmrzlikar: Just use `if not any(list(paired_reads))`.

resolwe_bio/tools/validate_fastq.py
```diff
@@ -0,0 +1,95 @@
...
+        for mate1, mate2 in zip(args.fq, args.fq2):
+            try:
+                paired_reads = dnaio.open(mate1, file2=mate2, fileformat='fastq')
+                if not any(True for r1, r2 in zip(paired_reads.reader1._iter, paired_reads.reader2._iter)):
+                    print(error(
+                        'Mate-pair files {} and {} contain no read sequences.'.format(basename(mate1), basename(mate2))
+                        )
+                    )
+                else:
+                    for read in paired_reads:
```
JureZmrzlikar: Just use list(paired_reads) and remove the `continue` line below. Same goes for single-end case.

resolwe_bio/tools/validate_fastq.py
```diff
@@ -0,0 +1,95 @@
...
+
+            except (FastqFormatError, FileFormatError) as dnaio_error:
+                print(error(
+                    'Error in mate-pairs {} and {}. {}'.format(basename(mate1), basename(mate2), str(dnaio_error))
+                    )
+                )
+    else:
+        for fq in args.fq:
+            try:
+                reads = dnaio.open(fq, fileformat='fastq')
```
JureZmrzlikar: `any(list(reads))`

resolwe_bio/tools/validate_fastq.py
```diff
@@ -0,0 +1,95 @@
...
+                if not any(True for _ in reads._iter):
```
JureZmrzlikar: `any(list(reads))`

resolwe_bio/tools/validate_fastq.py
jkokosar: I know this is ugly, but if one would use ``list(reads_reader_object)`` here, the memory consumptions would explode for the large input FASTQ files. @mzganec did you have some other simplifications in mind?

JureZmrzlikar: True, indeed, But you could still use simpler syntax for iteration over `paired_reads` object:
`any(True for r1, r2 in paired_reads))`

mzganec: > @mzganec did you have some other simplifications in mind?

`if not any(paired_reads):`

resolwe_bio/tools/validate_fastq.py
mzganec: `if not any(reads):`

resolwe_bio/tools/validate_fastq.py
jkokosar: @mzganec Thanks. This works for the single-end parsing. I had to retain the ``zip()`` for mate-pair parsing, since the iterator gets out of sync for some reason in the ``else`` part.

PR #713, opened 2019-01-10
Improve project's infrastructure

tjanez:


resolwe_bio/tools/etc.py
```diff
@@ -35,10 +35,10 @@
         if ndx > 0:
             try:
                 times.append(int(name[ndx + 2:ndx + 4]))
-            except:  # pylint: disable=bare-except
+            except Exception:  # pylint: disable=bare-except
```
dblenkus: Is pylint disable still needed?

mstajdohar: Good point, fixed.

PR #725, opened 2019-02-11
Persistence of basespace-file-import process changed from RAW to TEMP

hadalin:


jkokosar:
@hadalin @mstajdohar Should `files-to-fastq-single` and `files-to-fastq-paired` also have changed persistence?

jkokosar:
Isn't this process downloading large files? TEMP queue will kill the job if it takes more then 30s

mstajdohar:
> Should `files-to-fastq-single` and `files-to-fastq-paired` also have changed persistence?

No, because this processes generate reads data objects and we need to keep those.

> Isn't this process downloading large files? TEMP queue will kill the job if it takes more then 30s

The `scheduling_class: interactive` defined the queue. Persistence only means we can safely delete this objects at some point in the future.

PR #731, opened 2019-02-27
Add featureCounts-based 3' mRNA-seq pipeline 

jkokosar:
## Checklist

    before non-breaking changes.
    commit(s).

### Processes ###

  can be performed).
  and publication.

  tests).

### Tests ###


docs/CHANGELOG.rst
```diff
@@ -14,6 +14,9 @@ Changed
 -------
 - Bump MultiQC to 1.7.0 and add QoRTs (1.3.0) package to
   ``resolwebio/common:1.1.0`` Docker image
+- Save CPM-normalized expressions in ``feature_counts`` process. control
```
JureZmrzlikar: "control" --> "Control"

resolwe_bio/processes/support_processors/qorts_qc.py
```diff
@@ -0,0 +1,105 @@
...
+            '--generatePlots',
+            inputs.alignment.bam['file'],
+            inputs.annotation.annot['file'],
+            'qorts_output',
+        ]
+
+        optional_args = []
+
+        # Detect if aligned reads in BAM file are of single or paired-end type
+        if Cmd['samtools']('view', '-c', '-f', '1', inputs.alignment.bam['file']).strip() == '0':
```
JureZmrzlikar: Are you sure that this `samtools` command differs between single and paired end reads? Could you provide some more explanation?

resolwe_bio/processes/workflows/bbduk_star_featurecounts_quant.yml
```diff
@@ -0,0 +1,343 @@
...
+  category: Pipeline
+  description: >
+    This 3' mRNA-Seq pipeline is comprised of QC, preprocessing,
+    alignment and quantification steps.
+
+    Reads are preprocessed by __BBDuk__ which removes adapters, trims
+    reads for quality from the 3'-end, and discards reads that are too short
+    after trimming. Preprocessed reads are aligned by __STAR__
+    aligner. For read-count quantification, the __FeatureCounts__ tool
+    is used.
```
JureZmrzlikar: It would be nice to also add a sentence or two regarding the QC steps...

resolwe_bio/processes/expression/featureCounts.yml
```diff
@@ -581,13 +600,27 @@
...
+        expression2storage.py --output json.txt {{ name ~ '_tpm.tab.gz' }}
+        {% set default_exp_output = name ~ '_tpm.tab.gz' %}
+      {% elif normalization_type == 'CPM' %}
+        expression2storage.py --output json.txt {{ name ~ '_cpm.tab.gz' }}
+        {% set default_exp_output = name ~ '_cpm.tab.gz' %}
+      {% elif normalization_type == 'FPKM' %}
+        expression2storage.py --output json.txt {{ name ~ '_cpm.tab.gz' }}
+        {% set default_exp_output = name ~ '_fpkm.tab.gz' %}
+      {% else %}
+        re-error Unsupported expression normalization type used.
```
mzganec: Put the error message in quotation marks.

resolwe_bio/processes/expression/featureCounts.yml
```diff
@@ -123,6 +123,19 @@
             output table. In GTF files this is usually 'gene_id', in GFF3 files
             this is often 'ID', and 'transcript_id' is frequently a valid
             choice for both annotation formats.
+    - name: normalization_type
```
mzganec: Consider moving this setting to advanced options.

resolwe_bio/processes/support_processors/qorts_qc.py
```diff
@@ -0,0 +1,105 @@
...
+            'docker': {
+                'image': 'resolwebio/common:1.1.0',
+            },
+        },
+        'resources': {
+            'cores': 1,
+            'memory': 4096,
+        },
+    }
+    data_name = "QoRTs QC report ({{alignment|sample_name}})"
```
mzganec: Just curious: can we avoid jinja expressions in python processes?

resolwe_bio/processes/workflows/bbduk_star_featurecounts_quant.yml
```diff
@@ -0,0 +1,343 @@
...
+              label: Seed
+              type: basic:integer
+              default: 11
+            - name: fraction
+              label: Fraction
+              type: basic:decimal
+              required: false
+              range: [0, 1.0]
+              description: >
+                Use the fraction of reads [0 - 1.0] from the orignal input file instead
```
mzganec: Use a comma in specification of range. Consider _Use the fraction of reads in range [0.0, 1.0] from ..._

resolwe_bio/processes/workflows/bbduk_star_featurecounts_quant.yml
```diff
@@ -0,0 +1,343 @@
...
+          label: Number of reads
+          type: basic:integer
+          default: 1000000
+        - name: advanced
+          label: Advanced options
+          group:
+            - name: seed
+              label: Seed
+              type: basic:integer
+              default: 11
```
mzganec: :smile:

resolwe_bio/processes/support_processors/qorts_qc.py
jkokosar: There is currently no other way of getting to `sample_name` (it works for `data_name` only). When this will change, we will be able to avoid jinja

resolwe_bio/processes/support_processors/qorts_qc.py
jkokosar: I have documented this a bit better. Tnx!

PR #742, opened 2019-03-22
Add single cell processes

jvrakor:
@jkokosar 

## Checklist

* [x] Make the processes and tests work
    before non-breaking changes.
    commit(s).
* [x] All output get saved.


mstajdohar:
cc @JakaKokosar

mstajdohar:
@JakaKokosar It would be great if @jvrakor's work would be compatible with yours. Please review this PR and sync with @jvrakor on Slack.

resolwe_bio/processes/import_data/seq_reads_10x.py
```diff
@@ -0,0 +1,68 @@
...
+    ListField,
+    Process,
+    SchedulingClass,
+)
+
+
+class ImportScRNA10x(Process):
+    """Import 10x scRNA reads in FASTQ format."""
+    slug = 'upload-sc-10x'
+    name = 'Uload scRNA 10x fastqs'
```
jkokosar: Uload -> Upload

resolwe_bio/processes/import_data/seq_reads_10x.py
```diff
@@ -0,0 +1,68 @@
...
+            }
+        }
+    }
+    data_name = '{{ reads.0.file|default("?") }}'
+
+    class Input:
+        """Input fields to process ImportScRNA10x."""
+        barcodes = ListField(
+            FileField(
+                required=True,
```
jkokosar: `required=True` is the default, so this is redundant. The same applies bellow.

resolwe_bio/processes/scseq/cellranger.py
```diff
@@ -0,0 +1,264 @@
...
+class CellRangerMkref(Process):
+    """Cell ranger mkref."""
+
+    slug = 'cellranger-mkref'
+    name = 'Cell Ranger Mkref'
+    process_type = 'data:genomeindex:10x'
+    version = '1.0.0'
+    category = 'scRNA-Seq'
+    sheduling_class = SchedulingClass.BATCH
+    entity = {
```
jkokosar: Genome index should not be of entity type Sample

resolwe_bio/processes/scseq/cellranger.py
```diff
@@ -0,0 +1,264 @@
...
+        bai_path = os.path.join(output_dir, 'possorted_genome_bam.bam.bai')
+        os.symlink(bam_path, bam_name)
+        os.symlink(bai_path, bai_name)
+        process_inputs = {
+            'src': bam_name,
+            'src2': bai_name,
+            'species': inputs.genome_index.species,
+            'build': inputs.genome_index.build,
+        }
+        self.run_process('upload-bam-indexed', process_inputs)
```
jkokosar: This will likely create a new, separate instance of Sample. Is this as intended?

jvrakor: Good point. That is not intended. Any workaround without writing a new process?

jkokosar: Not that I can think of. We used redundant processes like [this](https://github.com/genialis/resolwe-bio/blob/master/resolwe_bio/processes/import_data/expressions.yml#L211) before to solve the same issue.

resolwe_bio/processes/scseq/cellranger.py
```diff
@@ -0,0 +1,260 @@
...
+        os.symlink(bam_path, bam_name)
+        os.symlink(bai_path, bai_name)
+        process_inputs = {
+            'src': bam_name,
+            'src2': bai_name,
+            'reads': inputs.reads.id,
+            'species': inputs.genome_index.species,
+            'build': inputs.genome_index.build,
+        }
+        self.run_process('upload-bam-scseq-indexed', process_inputs)
```
jvrakor: The process does not get spawned and I do not know why. It works if I run it by itself.

PR #763, opened 2019-04-30
Support automatic releasing to PyPI

dblenkus:


mstajdohar:
Nice!

mstajdohar:
Update contributing docs.

romunov:
Can we do something about this PR? After 4 years, it may be ready to be closed.

PR #766, opened 2019-05-09
Add bamclipper tool to dnaseq docker and fix building issues

romunov:
A bamclipper tool is added to `resolwebio/dnaseq`. GATK has been bumped to latest version 4.1.2.0. I would recommend tagging this docker image as 4.2.0.

In addition, building of the image was interrupted when building `chemut` R package dependency `assertthat`.

```
Error in run(bin, args = real_cmdargs, stdout_line_callback = real_callback(stdout),  : 
  lazy-load database '/usr/local/lib/R/site-library/assertthat/R/assertthat.rdb' is corrupt
Calls: install_github ... with.default -> eval -> eval -> with_envvar -> force -> run
In addition: Warning message:
In run(bin, args = real_cmdargs, stdout_line_callback = real_callback(stdout),  :
  internal error -3 in R_decompress1
Execution halted
xargs: bash: exited with status 255; aborting
```

The root cause is not exactly known (possibly due to updating of an already loaded package). In any case, the solution from the internet was to restart the R session. This is not really feasible in a non-interactive session. To avoid this, we explicitly install certain dependencies before installing `chemut` package.

Some Bioconductor dependencies listed in [chemut](https://github.com/jkokosar/chemut/blob/master/DESCRIPTION#L9) package are now installed explicitly through the new `BiocManager` package. This may overwrite (update) [`Rsamtools` package](https://github.com/genialis/resolwe-bio/blob/master/resolwe_bio/docker_images/common/packages-r-bioconductor.txt#L1) from `common` docker image.

## Checklist

    before non-breaking changes.
    commit(s).
* [ ] ~~Bump the process version~~:
* [ ] ~~All inputs are used in process~~.
* [ ] ~~All output fields have their ``re-save`` calls~~.

### Processes ###

  can be performed).
  and publication.

  tests).

### Tests ###

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -25,6 +25,11 @@ Added
 - Support Python 3.7
 - Add Tabix (1.7-2) to ``resolwebio/wgbs`` docker image
 - Add JBrowse index output to ``hmr`` process
+- Add ``bamclipper`` tool and ``parallel`` package to ``resolwebio/dnaseq:4.2.0`` image
+
+Fixed
+-----
+- Fix building problems of ``resolwebio/dnaseq`` docker
 
 
 =================
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -19,6 +19,7 @@ Changed
   ``resolwebio/common:1.1.3`` Docker image
 - Use ``resolwebio/common:1.1.3`` Docker image as a base image for
   ``resolwebio/rnaseq:4.4.2``
+- Bump GATK4 version to 4.1.2.0 in ``resolwebio/dnaseq:4.2.0``
 
 Added
 -----
--- a/resolwe_bio/docker_images/dnaseq/README.md
+++ b/resolwe_bio/docker_images/dnaseq/README.md
@@ -9,6 +9,6 @@ Included bioinformatics tools:
 ------------------------------
 * bamclipper (1.1.1)
 * CheMut
-* GATK (4.0.11.0)
+* GATK (4.1.2.0)
 * LoFreq (2.1.3.1)
 * Primerclip (v180308)
--- a/resolwe_bio/docker_images/dnaseq/README.md
+++ b/resolwe_bio/docker_images/dnaseq/README.md
@@ -7,6 +7,7 @@ https://hub.docker.com/r/resolwebio/base/) image.
 
 Included bioinformatics tools:
 ------------------------------
+* bamclipper (1.1.1)
 * CheMut
 * GATK (4.0.11.0)
 * LoFreq (2.1.3.1)
--- a/resolwe_bio/docker_images/dnaseq/packages-manual/bamclipper.sh
+++ b/resolwe_bio/docker_images/dnaseq/packages-manual/bamclipper.sh
@@ -1,0 +1,17 @@
+#!/bin/bash -e
+
+. /var/cache/build/packages-manual/common.sh
+
+download_and_verify \
+    tommyau \
+    bamclipper \
+    1.1.1 \
+    081a6008b16209ed27bb7fd6e1e86e89efc431e8cb958990f84a969b301485e2 \
+    'https://github.com/tommyau/bamclipper/archive/master.zip'
+
+mv bamclipper-master bamclipper
+
+add_binary_path \
+    tommyau \
+    bamclipper \
+    bamclipper
--- a/resolwe_bio/docker_images/dnaseq/packages-manual/r-packages.sh
+++ b/resolwe_bio/docker_images/dnaseq/packages-manual/r-packages.sh
@@ -2,9 +2,18 @@
 
 . /var/cache/build/packages-manual/common.sh
 
+# These steps need to be performed because certain packages produce a "database corrupt error". The solution is to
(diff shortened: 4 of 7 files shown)


resolwe_bio/docker_images/dnaseq/packages-manual/bamclipper.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash -e
+
+. /var/cache/build/packages-manual/common.sh
+
+download_and_verify \
+    genialis \
```
jkokosar: Bamclipper was obtained at https://github.com/tommyau/bamclipper. We should reference that and maybe install this to `tommyau` folder instead of `genialis`

docs/CHANGELOG.rst
```diff
@@ -6,6 +6,21 @@ All notable changes to this project are documented in this file.
 This project adheres to `Semantic Versioning <http://semver.org/>`_.
 
 
+=================
+Unreleased
+=================
+
+Changed
+-------
+- Add bamclipper tool to ``resolwebio/dnaseq:4.3.0`` image
```
jkokosar: Bamclipper requires `parallel` package to run. Append this to `resolwebio/dnaseq` Docker installation files.

romunov: `parallel` come shipped with R (since 2.14.0) so I'm not sure this is really needed.

jkokosar: Try to run the tool in locally build image to test this and be sure.

romunov: Mea culpa, I now see you were referring to the linux's `parallel` package, not R's.

resolwe_bio/docker_images/dnaseq/packages-manual/bamclipper.sh
romunov: Done.

docs/CHANGELOG.rst
romunov: Added parallel package to ubuntu installation.

resolwe_bio/docker_images/dnaseq/packages-manual/r-packages.sh
```diff
@@ -2,9 +2,18 @@
 
 . /var/cache/build/packages-manual/common.sh
 
+# These steps need to be performed because certain packages produce a "database corrupt error". The solution is to
+# restart R, but this is not feasible during the docker built or building of dependencies. This chunk also installs
+# packages from bioconductor. This entire step should be considered temporary until we can get packrat to work for
+# packages from all universes (CRAN, Bioconductor).
+Rscript --slave --no-save --no-restore-history -e "install.packages('assertthat')"
```
mzganec: All three `Rscript` commands can be joined not to repeat the first line.

docs/CHANGELOG.rst
```diff
@@ -19,7 +19,11 @@ Changed
 Added
 -----
 - Support Python 3.7
+- Add ``bamclipper`` tool and ``parallel`` package to ``resolwebio/dnaseq:4.2.0`` image
```
mzganec: Add `bamclipper` to the readme file of the Docker image as well.

resolwe_bio/docker_images/dnaseq/packages-manual/r-packages.sh
romunov: My reasoning for leaving them in three chunks is that they belong to three conceptual groups. The first one is a hack, that emulates restarting of R after installing the offending `assertthat` package. Second chunk is the explicit installation of the Bioconductor packages, which are a dependency in `chemut` package. And finally, the last chunk installs the package that we really want.
Of course I could always run things in one/two chunks and place comments in there, but since it's just an aesthetic thing and there's no performance hit, I would leave it the way it is. If you really insist on this, I can merge the chunks.

mzganec: You can leave it as it is. It was just an idea.

docs/CHANGELOG.rst
romunov: Done.

docs/CHANGELOG.rst
```diff
@@ -19,7 +19,11 @@ Changed
...
 
+Fixed
+-------
+- Fix building problems of ``dnaseq`` docker
```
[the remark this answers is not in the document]
romunov: Done.

romunov:
@mzganec @jkokosar I have added another commit bumping GATK4 version to 4.1.2.0. Please make another round of reviews.

resolwe_bio/docker_images/dnaseq/README.md
```diff
@@ -8,6 +8,7 @@ https://hub.docker.com/r/resolwebio/base/) image.
 Included bioinformatics tools:
 ------------------------------
 * CheMut
-* GATK (4.0.11.0)
+* GATK (4.1.2.0)
 * LoFreq (2.1.3.1)
 * Primerclip (v180308)
+* bamclipper
```
mzganec: Add version information (1.1.1) and move to the top (we use alphabetical ordering).

romunov: Nice catch. Added.

PR #771, opened 2019-05-21
Support RPKUM normalization in featureCounts

jkokosar:
## Checklist

    before non-breaking changes.
    commit(s).

### Processes ###

  can be performed).
  and publication.

  tests).

### Tests ###


resolwe_bio/tools/expression_rpkum.py
```diff
@@ -0,0 +1,77 @@
...
+def main():
+    """Invoke when run directly as a program."""
+    args = parse_arguments()
+
+    mappability = parse_mapability_file(args.mappability)
+    expression = parse_expression_file(args.counts)
+    lib_size = expression['Expression'].sum()
+
+    # Compute RPKUM expression
+    expression['Expression'] = expression.apply(
```
mzganec: If we read the expression file in a clever way, `expression` is a `pandas.Series` instead of `pandas.DataFrame` because we have just one column (`Expression`) and an index (`Gene`). Then we can use here [pandas.Series.apply](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.apply.html).

resolwe_bio/tools/expression_rpkum.py
```diff
@@ -0,0 +1,77 @@
...
+        lambda row: rpkum(row.name, row.Expression, get_coverage(row.name, mappability), lib_size),
```
mzganec: What happens if a gene in expression file is missing in mappability file?

resolwe_bio/tools/expression_rpkum.py
```diff
@@ -0,0 +1,77 @@
...
+    return parser.parse_args()
+
+
+def parse_expression_file(exp_file):
+    """Parse expression file to a Pandas dataframe."""
+    with gzip.open(exp_file) as exp:
+        df = pd.read_csv(exp, sep='\t')
+
+        ALLOWED_COLUMNS = ["Gene", "Expression"]
+        if not all(column_label in ALLOWED_COLUMNS for column_label in df.columns.values):
```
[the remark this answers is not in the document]
jkokosar: This are all good suggestions, but were made obsolete by the use of `usecols` in the `pd.read_csv`. Error handling for the missing columns is handled there instead.

resolwe_bio/tools/expression_rpkum.py
```diff
@@ -0,0 +1,77 @@
...
+            print(error('Invalid column headers {} in file {}.'.format(df.columns.values, exp_file)))
+            sys.exit(1)
+
+        # Cast Gene column to string
+        df['Gene'] = df['Gene'].astype('str')
```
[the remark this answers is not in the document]
jkokosar: `skip_blank_lines` is `True` by default and removes only empty rows. `dropna()` is still necessary to remove rows with empty column data (e.g. missing expression value).

resolwe_bio/tools/expression_rpkum.py
jkokosar: Accessing index value for the use in the `get_coverage()` gets harder, so there is no clear benefit for changing this to `Series.apply()`. Also, input files are not extremely large, so the performance is not an issue.

resolwe_bio/tools/expression_rpkum.py
jkokosar: Good catch. I have added additional error handling to `get_coverage()`.

jkokosar:
@mzganec Thanks for taking the time to produce such a quality review!

resolwe_bio/tools/expression_rpkum.py
mzganec: Replace with
```
    df = pd.concat([expression, mappability], axis=1)
    df['Expression'] = 10 ** 9 * df['Expression'] / df['coverage'] / lib_size
```

resolwe_bio/tools/expression_rpkum.py
```diff
@@ -0,0 +1,99 @@
...
+    mappability = parse_mapability_file(args.mappability)
+    expression = parse_expression_file(args.counts)
+    lib_size = expression['Expression'].sum()
+
+    # Compute RPKUM expression
+    expression['Expression'] = expression.apply(
+        lambda row: rpkum(row.Expression, get_coverage(row.name, mappability), lib_size),
+        axis=1
+    )
+    expression.sort_index(inplace=True)
```
mzganec: Are you sure? I presume that the input expression file with raw counts is already sorted. In this case, this line doesn't change anything. If the expression file with raw counts is unsorted it makes more sense to keep the same order of genes in RPKUM expression file as well.

Otherwise do not forget to change `expression` to `df` here.

resolwe_bio/tools/expression_rpkum.py
```diff
@@ -0,0 +1,99 @@
...
+    expression.to_csv(args.output, columns=['Expression'], header=True, index=True, sep='\t', compression='gzip')
```
mzganec: `df.to_csv`

PR #779, opened 2019-06-07
Add WES pipeline

romunov:
This PR is a work in progress and the final workflow doesn't pass tests yet.

Workflow consists of the following steps (processes):
- trimming (Trimmomatic)
- aligning (bwa mem)
- clipping (bamclipper) *
- optional removal of duplicates (GATK4 RemoveDuplicates) *
- recalibration of quality scores (GATK4 BaseQualityScoreRecalibration) *
- calling of variants (GATK4 HaplotypeCaller)

\* process new with this workflow

Input to the workflow are 
- raw (paired-end) reads in fastq format
- interval (bed) file which will limit analysis to a certain region (whole exon analysis part)
- bedpe file for bamclipper - can be parsed from bed file
- known sites of variation (VCF)
- dbSNP database of variants (VCF)
- various parameters for above mentioned processes

Output is a vcf file of called variants.

Test files are for gene STK11 and b37 genome cut from chromosome 19 from 1:1207200 (it just includes the STK11 exon). Files are placed into a folder structure where each process has its own folder. Each folder has input and output folders. Input folder holds data used on input and output is used to store data for checking the process outputs. Some of these outputs are used as input for the next process. E.g. data in output folder of markduplicates is passed as input data for the next process in line which recalibrates quality scores. Its result is used for variant calling...

## Checklist

    before non-breaking changes.
    commit(s).

### Processes ###

  can be performed).
  and publication.

  tests).

### Tests ###


resolwe_bio/processes/reads_processing/markduplicates.py
```diff
@@ -0,0 +1,174 @@
...
+            md_inputs = [
+                '--INPUT', f'{inputs.bam.bam.path}',
+                '--VALIDATION_STRINGENCY', f'{inputs.validation_stringency}',
+                '--OUTPUT', f'{bam}',
+                '--METRICS_FILE', f'{metrics_file}']
+
+            if inputs.remove_duplicates:
+                md_inputs.extend(['--REMOVE_DUPLICATES', f'{rmd}'])
+
+            if inputs.assume_sort_order is not '':
```
jkokosar: `if inputs.assume_sort_order:`

resolwe_bio/processes/reads_processing/markduplicates.py
```diff
@@ -0,0 +1,174 @@
...
+                md_inputs.extend(['--ASSUME_SORT_ORDER', f'{inputs.input.assume_sort_order}'])
+
+            print(md_inputs)
```
jkokosar: Remove this print. Parameters used are listed in the `Cmd` stdout

resolwe_bio/tests/processes/test_reads_filtering.py
```diff
@@ -223,3 +223,127 @@ def test_bbduk_paired(self):
...
+            }
+            input_bam = {
+                'src': './markduplicate/output/STK11.primerclipped.markduplicates.bam',
+                'species': species,
+                'build': build
+            }
+
+            ks_dbsnp = []
+            # dbSNP vcf has been cut to only specific region using VCFtools - 0.1.15. dbSNP database has
+            # been downloaded from Genialis platform.
```
jkokosar: Remove a reference to Genialis platform

resolwe_bio/tests/workflows/test_illumina_wes.py
```diff
@@ -0,0 +1,101 @@
...
+
+            intervals = []
+            for i in ['./bqsr/input/STK11.bed']:
+                intervals.append(
+                    self.run_process('upload-bed', {'src': i, 'species': species, 'build': build})
+                )
+
+            adapters = self.prepare_adapters()
+
+        print([i.id for i in ks])
```
jkokosar: Remove for final PR

jkokosar:
Fix pylint and pydocstyle errors.

resolwe_bio/processes/workflows/illumina_wes.yml
```diff
@@ -0,0 +1,312 @@
+- slug: workflow-illumina-wes
```
jkokosar: It's dubious what Illumina stands for here: is this for Illumina WES kits or for any WES done using Illumina sequencing technologies? In this regard, I am thinking it would be best to use just `worklfow-wes` and than use description field to explain the specifics. Please reflect any changes you make to `slug` also in the `Name` field

resolwe_bio/processes/workflows/illumina_wes.yml
```diff
@@ -0,0 +1,312 @@
...
+  name: Perform Illumina whole exome sequencing analysis
```
jkokosar: Omit "Perform".

resolwe_bio/processes/workflows/illumina_wes.yml
jkokosar: ..and `type` field.

resolwe_bio/processes/reads_processing/bqsr.py
```diff
@@ -0,0 +1,153 @@
...
+
+from resolwe.process import Process, Cmd, SchedulingClass, DataField, \
+    FileField, StringField, ListField
+
+
+class BQSR(Process):
+    """
+    A two pass process of BaseRecalibrator and ApplyBQSR from GATK.
+
+    See https://software.broadinstitute.org/gatk/documentation/tooldocs/current/org_broadinstitute_hellbender_tools_walkers_bqsr_BaseRecalibrator.php
```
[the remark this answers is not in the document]
lukaw3d: Yes, should work

resolwe_bio/processes/reads_processing/bqsr.py
```diff
@@ -0,0 +1,153 @@
...
+        bam_rg = f'{file_name}_RG.bam'
+
+        species = inputs.bam.species
+
+        # Parse read_group argument from a string, delimited by a \t and space
+        # into a form that will be accepted by AddOrReplaceReadGroups tool.
+        # E.g. '-LB DAB\t-PL Illumina\t-PU barcode -SM sample1' should become
+        # ['-LB', 'DAB', '-PL', 'Illumina', '-PU', 'barcode', '-SM', 'sample1']
+        # prepended by INPUT and OUTPUT.
+        if len(inputs.read_group) > 0:
```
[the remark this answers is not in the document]
romunov: To preserve the logic, I used `if not inputs.read_groups:`.

resolwe_bio/processes/reads_processing/markduplicates.py
romunov: Remnants of my debugging attempts, sorry.

resolwe_bio/processes/reads_processing/bqsr.py
```diff
@@ -0,0 +1,153 @@
...
+
+    See https://software.broadinstitute.org/gatk/documentation/tooldocs/current/org_broadinstitute_hellbender_tools_walkers_bqsr_BaseRecalibrator.php
+    for more information.
+    """
+
+    slug = 'bqsr'
+    name = 'BaseQualityScoreRecalibrator'
+    process_type = 'data:alignment:bam:bqsr:'
+    version = '1.0.0'
+    category = 'Clipping'  # TODO: decide where to put it
```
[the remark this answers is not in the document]
romunov: I haven't invested enough time to thinking more deeply about this. Yes, your suggestion sounds more in tune of what is going on in this process.

resolwe_bio/processes/reads_processing/bqsr.py
```diff
@@ -0,0 +1,153 @@
...
+            data_type='bed',
+            label='One or more genomic intervals over which to operate. This field is opotional, but it can speed up '
+                  'the process by restricting calculations to specific genome regions.',
+            required=True
+        )
+        read_group = StringField(
+            label='Replace read groups in a BAM file.This argument enables the user to replace all read groups in the '
+                  'INPUT file with a single new read group and assign all reads to this read group in the OUTPUT BAM '
+                  'file. Addition or replacement is performed using Picard\'s AddOrReplaceReadGroups tool. Input '
+                  'should take the form of -<name><space><value> delimited by a tab, e.g. "-ID 1\t-PL Illumina\t-SM '
```
lukaw3d: It will be very hard for users to type a <kbd>tab</kbd> into an input

resolwe_bio/processes/reads_processing/bqsr.py
```diff
@@ -0,0 +1,153 @@
...
+        # ['-LB', 'DAB', '-PL', 'Illumina', '-PU', 'barcode', '-SM', 'sample1']
+        # prepended by INPUT and OUTPUT.
+        if len(inputs.read_group) > 0:
+            arrg = [
+                '--INPUT', f'{inputs.bam.bam.path}',
+                '--OUTPUT', f'{bam_rg}'
+            ]
+
+            for x in inputs.read_group.split('\t'):
+                arrg.extend(x.split(' '))
```
lukaw3d: Should this be split up at only the first space encountered?
e.g. is this valid, `'-LB DAB 123\t-PL Illumina 123'` and should it parse to `['-LB', 'DAB 123', '-PL', 'Illumina 123']`?

resolwe_bio/processes/reads_processing/bqsr.py
romunov: It's not <kbd>TAB</kbd> per se, but rather its representation as `\t`. I will reflect this in the description.

resolwe_bio/processes/reads_processing/bqsr.py
romunov: The string should split only on the first space. Your suggestion doesn't parse well with my current solution.

```
x = '-LB DAB 123\t-PL Illumina 123'
x.split(' ')
['-LB', 'DAB', '123\t-PL', 'Illumina', '123']
```
Do you happen to have something more robust on your mind or should I think some more about this?

lukaw3d: I don't know.
One option is basic:text:, to have newlines as separator. And something less common than `<space>` as delimiter. E.g. `=`

(split has a `max` option)

romunov: I think having a `=` instead of space solves the issue with spaces in the name, which we should respect because it's the only [non-printing character](https://software.broadinstitute.org/gatk/documentation/tooldocs/4.1.2.0/picard_sam_AddOrReplaceReadGroups.php) allowed in the name. 

`basic:text` is a nice idea, but I don't see a whole lot of benefits, other than having the explicit new line. My guess would be that initially it would probably be a bit less intuitive. I would leave the `\t` as a separator for now. I have thought of adding something else, like a pipe (`|`), but some people may have problems finding that on their keyboards.

jkokosar: @romunov What about handling read info information [this](https://github.com/genialis/resolwe-bio/blob/master/resolwe_bio/processes/variant_calling/vc_chemut.yml#L53) way? You can define the type (and value range) of each input, and in the process control the delimiters. Its more inputs, but in the end it will be more robust.

romunov: If we go down that road, we would have to implement (and test) 13 tags whereas the current implementation is agnostic to the number of tags used. Perhaps we can do this for version 2.0?

jkokosar: It was just a suggestion, as I remembered that we were handling this information using separate fields in the past.

romunov: A valid suggestion nonetheless. Another case against this type of input would be the long list of parameters, which would be a PITA to fill out if you want to repeat this over and over. Until we get the auto-prefilled-parameters-based-on-previous-run feature, this would be somewhat cumbersome.

resolwe_bio/processes/reads_processing/bqsr.py
```diff
@@ -0,0 +1,153 @@
...
+        read_group = StringField(
+            label='Replace read groups in a BAM file.This argument enables the user to replace all read groups in the '
+                  'INPUT file with a single new read group and assign all reads to this read group in the OUTPUT BAM '
+                  'file. Addition or replacement is performed using Picard\'s AddOrReplaceReadGroups tool. Input '
+                  'should take the form of -<name><space><value> delimited by a tab, e.g. "-ID 1\t-PL Illumina\t-SM '
+                  'sample_1". See https://software.broadinstitute.org/gatk/documentation/tooldocs/4.1.2.0'
+                  '/picard_sam_AddOrReplaceReadGroups.php for more information on tag names. Note that PL, LB, '
+                  'PU and SM are require fields.',
+            required=False,
+            default=''
```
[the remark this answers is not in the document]
jkokosar: Please check that you fixed/evaluated issues that were already marked as resolved.

PR #789, opened 2019-07-02
Fix merging of expressions in archive-samples process

mzganec:


mzganec:
Parsing is a computationally intensive task. I used 91 samples from one of our commercial projects as a test dataset and got the following timings on my laptop equipped with 4 cores and 8 threads:

| threads | read [s] | concat [s] |
|---------|------|--------|
| 1 | 25.6 | 13.9 |
| 8 | 8.6 | 7.1 |

In the table **read** refers to the line `multiprocessing.Pool().map(read_csv, args.file_paths)` and **concat** to `multiprocessing.Pool().map(merge_expressions, items)`.

JureZmrzlikar:
Interesting, thanks!

PR #799, opened 2019-07-17
Fix spikeins-qc process to handle expressions without spikeins

JureZmrzlikar:
* [x] Update CHANGELOG.rst for each commit separately
* [x] Bump the process version


resolwe_bio/tools/spikein_pairwise.py
```diff
@@ -244,6 +247,15 @@ def main():
             exp_type=exp_type,
         )
 
+    if min_one_has_spikeins:
+        for message in warnings:
+            print(warning(message))
+    else:
```
jkokosar: Nice!

PR #804, opened 2019-07-26
Refactor Cutadapt, add 3' mRNA-seq workflow

jkokosar:
## Checklist

    before non-breaking changes.
    commit(s).

### Processes ###

  can be performed).
  and publication.

  tests).

### Tests ###


resolwe_bio/processes/reads_processing/cutadapt.yml
```diff
@@ -130,24 +145,28 @@
...
-       and not trailing
-       and not crop
-       and not headcrop
-       and not minlen %}
+      {% if not adapters.up_primers_file
+       and not adapters.down_primers_file
+       and not adapters.up_primers_seq
+       and not adapters.down_primers_seq
+       and not adapters.polya_tail
+       and not modify_reads.leading
```
mzganec: Consider adding `modify_reads.nextseq_trim` here.

resolwe_bio/processes/reads_processing/cutadapt.yml
```diff
@@ -25,93 +21,112 @@
...
+    - name: modify_reads
+      label: Read trimming
+      group:
+        - name: nextseq_trim
+          label: NextSeq-specific quality trimming
+          type: basic:integer
+          required: false
+          description: |
+            NextSeq-specific quality trimming (each read). Trims also dark
+            cycles appearing as high-quality G bases. This option is mutually
```
mzganec: In one of the future versions we should make a drop-down menu for selection of quality trimming type. This would make mutual exclusion mentioned in the description more obvious.

resolwe_bio/processes/workflows/cutadapt_star_fc_quant_tf.yml
```diff
@@ -0,0 +1,171 @@
...
+    This 3' mRNA-Seq pipeline is comprised of QC, preprocessing,
+    alignment and quantification steps.
+
+    Reads are preprocessed by __Cutadapt__ which removes adapters, trims
+    reads for quality from the 3'-end, and discards reads that are too short
+    after trimming. Preprocessed reads are aligned by __STAR__
+    aligner. For read-count quantification, the __FeatureCounts__ tool
+    is used.
+
+    QC steps include downsampling, QoRTs QC analysis and alignment of
```
mzganec: Consider rephrasing the downsampling part. Downsampling does not contribute any information, i.e. report, to QC. I would therefore make a distinction from the rest of the tools. Consider using something along the lines _QC operates on downsampled reads and includes ..._

resolwe_bio/processes/workflows/cutadapt_star_fc_quant_tf.yml
```diff
@@ -0,0 +1,171 @@
...
+    - name: globin_reference
+      label: Indexed Globin reference sequence
+      type: data:genomeindex:star
+      description: |
+        Reference sequence index prepared by STAR aligner indexing tool.
+    - name: show_advanced
+      label: Show advanced parameters
+      type: basic:boolean
+      default: false
+    - name: downsampling
```
mzganec: Listing downsampling options before cutadapt options is misleading. It may lead the user into thinking that filtering and trimming is done on downsampled FASTQ files.

resolwe_bio/processes/reads_processing/cutadapt_3prime.py
```diff
@@ -0,0 +1,145 @@
...
+                            "cycles appearing as high-quality G bases. This option is mutually "
+                            "exclusive with the use of standard quality-cutoff trimming and is "
+                            "suitable for the use with data generated by the recent Illumina "
+                            "machines that utilize two-color chemistry to encode the four bases.",
+                default=10,
+            )
+
+            quality_cutoff = IntegerField(
+                label="Quality cutoff",
+                description="Trim low-quality bases from 3' ends of each read before adapter "
```
mzganec: ends -> end

resolwe_bio/processes/workflows/cutadapt_star_fc_quant_tf.yml
```diff
@@ -0,0 +1,171 @@
...
+    - name: cutadapt
+      label: Cutadapt filtering
+      hidden: '!show_advanced'
+      group:
+      - name: quality_cutoff
+        label: Reads quality cutoff
+        type: basic:integer
+        required: false
+        description: |
+          Trim low-quality bases from 3' ends of each read before
```
mzganec: ends -> end

resolwe_bio/processes/reads_processing/cutadapt_3prime.py
```diff
@@ -0,0 +1,145 @@
...
+            '-m', inputs.options.min_len,
+            '-O', inputs.options.min_overlap,
+            '-g', 'truseq=A{18}AGATCGGAAGAGCACACGTCTGAACTCCAGTCAC',
+            '--discard-trimmed',
+            '-o', '{}_trimmed.fastq.gz'.format(name),
+            '-',
+        ]
+
+        # Run Cutadapt, write analysis reports into a report file
+        (Cmd['cutadapt'][first_pass_input] | Cmd['cutadapt'][second_pass_input] | Cmd['cutadapt'][third_pass_input]) \
```
mzganec: Why don't we run cutadapt in multiple threads (`-j` option) here as well?

PR #805, opened 2019-07-29
Add skip parameter to bamclipper process and wes-pipeline workflow

romunov:
This PR adds a skip parameter to `bamclipper` tool. This enables the WES pipeline to process not only data from amplicon based kits, but also hybridization based kits.

## Checklist
    before non-breaking changes.
    commit(s).

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -24,6 +24,8 @@ Added
 -----
 - Add ``cutadapt-3prime-single`` process
 - Add ``workflow-cutadapt-star-fc-quant-single`` process
+- Add argument ``skip`` to ``bamclipper`` which enables skipping of
+  the said process
 
 
 ===================
--- a/resolwe_bio/processes/reads_processing/bamclipper.py
+++ b/resolwe_bio/processes/reads_processing/bamclipper.py
@@ -1,7 +1,8 @@
 """Remove primer sequences from BAM alignments by soft-clipping."""
 import os
 
-from resolwe.process import Process, Cmd, SchedulingClass, DataField, FileField, StringField
+from resolwe.process import Process, Cmd, SchedulingClass, DataField, FileField, StringField, BooleanField
+from shutil import copy2
 
 
 class Bamclipper(Process):
@@ -13,7 +14,7 @@ class Bamclipper(Process):
     slug = 'bamclipper'
     name = 'Bamclipper'
     process_type = 'data:alignment:bam:bamclipped:'
-    version = '1.0.0'
+    version = '1.1.0'
     category = 'Clipping'
     shaduling_class = SchedulingClass.BATCH
     entity = {'type': 'sample'}
@@ -31,7 +32,12 @@ class Bamclipper(Process):
         """Input fields to process Bamclipper."""
 
         alignment = DataField('alignment:bam', label='Alignment BAM file')
-        bedpe = DataField('bedpe', label='BEDPE file')
+        bedpe = DataField('bedpe', label='BEDPE file', required=False)
+        skip = BooleanField(
+            label='Skip Bamclipper step',
+            description='Use this option to skip Bamclipper step.',
+            default=False
+        )
 
     class Output:
         """Output fields to process Bamclipper."""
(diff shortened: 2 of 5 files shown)


resolwe_bio/tests/workflows/test_wes.py
```diff
@@ -93,3 +97,10 @@ def test_wes(self):
         self.run_process('workflow-wes', input_workflow)
         wes = Data.objects.last()
         self.assertFileExists(wes, 'vcf')
+
+        # Test skipping bamclipper
+        input_workflow_skipbc = input_workflow
+        input_workflow_skipbc['advanced']['bamclipper']['skip'] = True
+        self.run_process('workflow-wes', input_workflow_skipbc)
+        wes_skip = Data.objects.filter(process_info=['Skipping bamclipper step.'])
```
jkokosar: It would be better to filter for the correct object based on process type, and then check if the correct process info is present.

resolwe_bio/processes/reads_processing/bamclipper.py
```diff
@@ -46,38 +52,51 @@ class Output:
...
-        bai = f'{name}.primerclipped.bam.bai'
 
-        self.progress(0.5)
+        # If so specified, skip bamclipper step. Prepare outputs to match those of as if bamclipping proceeded.
+        if inputs.skip:
+            bam = f'{name}.bamclipper_skipped.bam'
+            bai = f'{bam}.bai'
+            bigwig = f'{bam[:-4]}.bw'
+
+            copy2(inputs.alignment.bam.path, bam)
```
jkokosar: What about the copy of `bai` and `bigwig` files? I don't see where those get created in the case of `skip = True`

jkokosar: Oh, I see they are created either way.

resolwe_bio/tests/workflows/test_wes.py
romunov: `.get` doesn't work because there are two objects available which throws an error. I use `filter()` and test for the last (highest ID) object.

resolwe_bio/tests/workflows/test_wes.py
```diff
@@ -44,7 +44,7 @@ def test_wes(self):
         input_workflow = {
             'reads': reads.pk,
             'genome': genome.id,
-            'bamclipper_bedpe': bc_bedpe.id,
+            # 'bamclipper_bedpe': bc_bedpe.id,
```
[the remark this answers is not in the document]
romunov: Done.

docs/CHANGELOG.rst
```diff
@@ -20,6 +20,8 @@ Added
 - Add pigz package to ``resolwebio/common:1.2.0`` Docker image
 - Add StringTie and UMI-tools to ``resolwebio/rnaseq:4.7.0`` Docker
   image
+- Add argument ``skip`` to ``bamclipper`` which enables skipping of
```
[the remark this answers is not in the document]
romunov: Please see if it looks correct.

resolwe_bio/tests/workflows/test_wes.py
```diff
@@ -93,3 +96,16 @@ def test_wes(self):
...
         wes = Data.objects.last()
         self.assertFileExists(wes, 'vcf')
+
+        # Test skipping bamclipper
+        input_workflow_skipbc = input_workflow
+        input_workflow_skipbc['advanced']['bamclipper']['skip'] = True
+        self.run_process('workflow-wes', input_workflow_skipbc)
+        wes_skip = Data.objects.filter(process__slug='bamclipper')
+
+        # wes_skip returns two objects. Find the one with the highest ID (which
```
jkokosar: This can probably be solved by `wes_skip = Data.objects.filter(process__slug='bamclipper').last()`

docs/CHANGELOG.rst
```diff
@@ -5,7 +5,6 @@ Change Log
 All notable changes to this project are documented in this file.
 This project adheres to `Semantic Versioning <http://semver.org/>`_.
 
-
```
jkokosar: Don't remove this line. Is supposedly there on purpose ;)

resolwe_bio/tests/workflows/test_wes.py
jkokosar: @romunov have use tried using `.last()`? It would be more concise.

romunov: We're not interested in the last object as bamclipper is the second step (+ the uploads).

jkokosar:
```
wes_skip = Data.objects.filter(process__slug='bamclipper').last()
self.assertEqual(wes_skip.process_info, ['Skipping bamclipper step.'])
```
Test pass OK for me. 
Using `process__slug` filter will return 2 object, and `.last` will return the object with higher ID, making the code below unnecessary.

docs/CHANGELOG.rst
romunov: Done.

PR #808, opened 2019-08-01
Workaround for warehouse issue with requirements

dblenkus:


dblenkus:
This was fixed in warehouse.

PR #816, opened 2019-08-15
Fix warning message in hierarchical clustering of genes

mzganec:
@jinlajf Thanks for reporting the issue.


resolwe_bio/tools/genehcluster.py
```diff
@@ -213,7 +213,6 @@ def main():
                    'samples. However, hierarchical clustering of genes cannot be computed with '
                    'just one gene.'.format(gene_names[0]))
             set_error(msg)
-        removed = [name for i, name in enumerate(expressions.index) if not matches[i]]
```
mzganec: This line was the root of the problem: `expressions` contained an already filtered set of genes but `matches` did not, `len(expressions) <= len(matches)`. Therefore, the number and the order of genes in the two variables didn't match.

PR #823, opened 2019-08-28
Index GTF/GFF features upon annotation file upload

jkokosar:
Requires https://github.com/genialis/resolwe-bio/pull/822

### DIFF
--- a/resolwe_bio/processes/import_data/annotation.yml
+++ b/resolwe_bio/processes/import_data/annotation.yml
@@ -9,13 +9,13 @@
     expression-engine: jinja
     executor:
       docker:
-        image: resolwebio/rnaseq:4.0.0
+        image: resolwebio/rnaseq:4.9.0
     resources:
       cores: 1
-      memory: 4096
+      memory: 8192
       network: true
   data_name: '{{ src.file }}'
-  version: 3.2.1
+  version: 3.3.0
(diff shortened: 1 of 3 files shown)


resolwe_bio/processes/import_data/annotation.yml
```diff
@@ -107,14 +101,26 @@
...
+        --trackLabel "annotation" \
+        --nameAttributes "name,id,alias,gene_id,gene_name"
       re-checkrc "Annotation track processing for JBrowse failed."
 
+      # Prepare Seq folder
+      prepare-refseqs.pl --gff-sizes "${NAME}"_sorted.gff3 --out .
+      re-checkrc "Refseq track processing for JBrowse failed."
+
+      # Index features
+      generate-names.pl --out .
```
mzganec: Running the corresponding test with `resolwebio/rnaseq:4.9.0` Docker image I'm getting `No reference sequences defined in configuration, nothing to do.` The process fails in the next line.

jkokosar: @mzganec `gt gff3validator` was replaced with `gt gff3 -tidy` mode, since the latter command reports and attempts to fix the file format errors, while the former command just reports the GFF3 format issues and does not stop the process, leading into issues downstream.

PR #826, opened 2019-09-09
Add network access to upload-bedpe process

romunov:
This pull request grants access to network for process `upload-bedpe` to enable uploading of files from the internet.

## Checklist
    before non-breaking changes.
    commit(s).

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -28,6 +28,9 @@ Added
 - Add ``workflow-bbduk-salmon-qc-single`` and
   ``workflow-bbduk-salmon-qc-paired`` workflows
 
+Fixed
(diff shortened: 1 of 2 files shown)


docs/CHANGELOG.rst
```diff
@@ -22,6 +22,7 @@ Changed
 - Support Salmon reports in MultiQC and expose ``dirs_depth`` parameter
 - Expose transcript-level expression file in the ``salmon-quant``
   process
+- Give process ``upload-bedpe`` access to network
```
jkokosar: This can be considered a bug fix, so please move under Fixed section. Also, bump the process version at the last digit, to `1.0.2`. Otherwise looks good.

PR #830, opened 2019-09-24
Increase memory and CPU limit in PCA process

mzganec:
Parallelize extraction of expressions.


jkokosar:
@mzganec Is this PR still relevant?

mzganec:
There is no immediate need for this to be merged. The problem has been fixed by changing Docker parameters in `resolwe`. Closing.

PR #835, opened 2019-10-18
Robust BaseSpace file download

hadalin:
## Checklist

    before non-breaking changes.
    commit(s).

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -14,6 +14,10 @@ Added
 -----
 - Add ``resolwebio/slamdunk`` Docker image
 
+Changed
+-------
+- Make BaseSpace file download more robust
+
 
 ===================
 23.1.1 - 2019-10-11
--- a/resolwe_bio/tools/basespace_download.py
+++ b/resolwe_bio/tools/basespace_download.py
@@ -75,12 +75,12 @@ def main():
             file_name, file_size = get_file_properties(session, file_id, headers)
             download_file_repeatedly(args.tries, session, file_id, file_name, file_size, headers)
             output(args.output, f'filename={file_name}')
-    except Exception:
+    except Exception as error:
         if args.verbose:
             traceback.print_exc()
         else:
-            print("An error occurred while processing the Basespace download request. "
-                  "Use --verbose to see details.")
+            print(str(error))
+            print("Use --verbose to see traceback.")
 
         sys.exit(1)
 
--- a/resolwe_bio/tools/basespace_download.py
+++ b/resolwe_bio/tools/basespace_download.py
@@ -134,24 +134,24 @@ def make_get_request(session, url, headers, stream=False):
     return response
 
 
-def get_basespace_api_url():
+def get_api_url():
     """Get base BaseSpace API URL."""
     return 'https://api.basespace.illumina.com/v1pre3'
 
 
-def get_basespace_api_file_url(file_id):
+def get_api_file_url(file_id):
     """Get BaseSpace API file URL."""
-    return '{}/files/{}'.format(get_basespace_api_url(), file_id)
+    return '{}/files/{}'.format(get_api_url(), file_id)
 
 
-def get_basespace_api_file_content_url(file_id):
+def get_api_file_content_url(file_id):
     """Get BaseSpace API file contents URL."""
-    return '{}/content'.format(get_basespace_api_file_url(file_id))
+    return '{}/content'.format(get_api_file_url(file_id))
 
 
 def get_file_properties(session, file_id, request_headers):
     """Get file name and size (in bytes)."""
-    response = make_get_request(session, get_basespace_api_file_url(file_id), request_headers)
+    response = make_get_request(session, get_api_file_url(file_id), request_headers)
     info = response.json()['Response']
     return info['Name'], info['Size']
 
@@ -171,7 +171,12 @@ def download_file_repeatedly(tries, session, file_id, file_name, expected_file_s
 
 def download_file(session, file_id, file_name, expected_file_size, request_headers):
     """Download BaseSpace file."""
-    response = make_get_request(session, get_basespace_api_file_content_url(file_id), request_headers, stream=True)
+    response = make_get_request(
+        session,
+        get_api_file_content_url(file_id),
+        request_headers,
+        stream=True
+    )
 
     try:
         with open(file_name, 'wb') as f:
--- a/resolwe_bio/tools/basespace_download.py
+++ b/resolwe_bio/tools/basespace_download.py
@@ -1,10 +1,14 @@
 #!/usr/bin/env python3
 """Tool to download files from BaseSpace."""
 
-import sys
-import traceback
-import atexit
 import argparse
+import atexit
+import gzip
+import os
+import sys
+import time
+import traceback
+
 import requests
 
 
@@ -20,6 +24,18 @@ def main():
     atexit.register(on_exit, session)
 
     parser = argparse.ArgumentParser(description='Download file from Illumina BaseSpace.')
+
+    def check_tries_within_range(arg):
+        try:
+            value = int(arg)
+        except ValueError:
+            raise argparse.ArgumentTypeError("Not an integer")
+
+        if not 1 <= value <= 10:
+            raise argparse.ArgumentTypeError("Not between 1 and 10")
+
+        return value
+
     parser.add_argument('--file-id',
                         dest='file_ids',
                         action='append',
@@ -37,6 +53,11 @@ def main():
                         help="Sets what is printed to standard output. "
                              "Argument 'full' outputs everything, "
                              "argument 'filename' outputs only file names of downloaded files.")
+    parser.add_argument('--tries',
+                        dest='tries',
+                        type=check_tries_within_range,
+                        default=3,
+                        help="Number of tries to download a file before giving up.")
     parser.add_argument('--verbose',
                         dest='verbose',
                         action='store_true',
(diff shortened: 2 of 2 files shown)


resolwe_bio/tools/basespace_download.py
```diff
@@ -51,8 +72,8 @@ def main():
         headers = {'x-access-token': access_token}
 
         for file_id in file_ids:
-            file_name = get_file_name(session, file_id, headers)
-            download_file(session, file_id, file_name, headers)
+            file_name, file_size = get_file_properties(session, file_id, headers)
+            download_file_repeatedly(args.tries, session, file_id, file_name, file_size, headers)
             output(args.output, 'filename={}'.format(file_name))
     except Exception:
         if args.verbose:
```
dblenkus: You shouldn't hide the error.

docs/CHANGELOG.rst
```diff
@@ -14,6 +14,7 @@ Changed
 -------
 - Renamed ``workflow-trim-align-quant`` workflow to make the name more
   informative
+- Made BaseSpace file download more robust
```
dblenkus: This should go under unreleased.

resolwe_bio/tools/basespace_download.py
hadalin: This is a security measure. You don't want to print anything because of the secrets. There's a verbose argument for this reason.

resolwe_bio/tools/basespace_download.py
```diff
@@ -125,25 +148,58 @@ def get_basespace_api_file_content_url(file_id):
...
-    """Get file name."""
+def get_file_properties(session, file_id, request_headers):
+    """Get file name and size (in bytes)."""
     response = make_get_request(session, get_basespace_api_file_url(file_id), request_headers)
-    return response.json()['Response']['Name']
+    info = response.json()['Response']
+    return info['Name'], info['Size']
+
 
+def download_file_repeatedly(tries, session, file_id, file_name, expected_file_size, request_headers):
```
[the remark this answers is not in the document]
hadalin: This is then less readable.

resolwe_bio/tools/basespace_download.py
dblenkus: Can you give me an example when a secret is printed?

resolwe_bio/tools/basespace_download.py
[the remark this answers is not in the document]
dblenkus: Actually it is more readable, because you know what both functions do, what is the differenc, which one to use,...

resolwe_bio/tools/basespace_download.py
```diff
@@ -125,25 +148,58 @@ def get_basespace_api_file_content_url(file_id):
...
-    response = make_get_request(session, get_basespace_api_file_content_url(file_id), request_headers, True)
+    response = make_get_request(session, get_basespace_api_file_content_url(file_id), request_headers, stream=True)
 
     try:
         with open(file_name, 'wb') as f:
-            for chunk in response.iter_content(chunk_size=1024):
+            chunk_size = 1024 * 1024 * 10
+            for chunk in response.iter_content(chunk_size=chunk_size):
                 f.write(chunk)
     except FileNotFoundError:
```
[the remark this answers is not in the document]
hadalin: When `open(file_name, 'wb')` is unsuccessful.

resolwe_bio/tools/basespace_download.py
```diff
@@ -125,25 +148,58 @@ def get_basespace_api_file_content_url(file_id):
...
         raise BaseSpaceDownloadError('Could not save file to {}, due to directory not being found'.format(file_name))
     except PermissionError:
```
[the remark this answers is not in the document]
hadalin: Same as above.

resolwe_bio/tools/basespace_download.py
dblenkus: You don't need to print the whole traceback, but the error message should definitely be printed.

hadalin: I can't, but it's a possibility. Kostko was pretty adamant about it.

resolwe_bio/tools/basespace_download.py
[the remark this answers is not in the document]
hadalin: I disagree, because `_` tells yo uless than `repeatedly`.

resolwe_bio/tools/basespace_download.py
hadalin: It's a broad exception catch so you don't know what going to be printed. There could be sensitive information even in the message.

resolwe_bio/tools/basespace_download.py
```diff
@@ -125,25 +148,58 @@ def get_basespace_api_file_content_url(file_id):
...
+    return info['Name'], info['Size']
+
 
+def download_file_repeatedly(tries, session, file_id, file_name, expected_file_size, request_headers):
+    """Attempt to download BaseSpace file numerous times in case of errors."""
+    for index in range(tries):
+        try:
+            download_file(session, file_id, file_name, expected_file_size, request_headers)
+            break
+        except BaseSpaceDownloadError as error:
```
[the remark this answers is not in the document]
hadalin: If I remember correctly `requests` don't raise network errors. You check if `status_code == 200` and file integrity at the end. I have to make sure though.

resolwe_bio/tools/basespace_download.py
dblenkus: I am pretty sure that Python never prints any values in tracebacks - at most it reveals dict's keys. And I am totally certain that it doesn't print anything sensitive in the error message if you don't put it there 🙂

The user should know why the download failed, so you need to print the actual message.

resolwe_bio/tools/basespace_download.py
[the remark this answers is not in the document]
dblenkus: Also check what happens if reading from (streamed) response fails.

resolwe_bio/tools/basespace_download.py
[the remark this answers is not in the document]
dblenkus: It tells you everything you need to know - don't use that function.

It is also mentioned in [PEP8](https://www.python.org/dev/peps/pep-0008/):
> `_single_leading_underscore`: weak "internal use" indicator.

docs/CHANGELOG.rst
```diff
@@ -6,6 +6,15 @@ All notable changes to this project are documented in this file.
 This project adheres to `Semantic Versioning <http://semver.org/>`_.
 
 
+===================
```
dblenkus: Match the length with the title.

resolwe_bio/tools/basespace_download.py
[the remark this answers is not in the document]
hadalin: But you could use it. Maybe you don't care about repeated downloading. This is a personal preference and has nothing to do with any functionality.

resolwe_bio/tools/basespace_download.py
```diff
@@ -125,25 +149,58 @@ def get_basespace_api_file_content_url(file_id):
...
-            for chunk in response.iter_content(chunk_size=1024):
+            chunk_size = 1024 * 1024 * 10
+            for chunk in response.iter_content(chunk_size=chunk_size):
                 f.write(chunk)
     except FileNotFoundError:
         raise BaseSpaceDownloadError('Could not save file to {}, due to directory not being found'.format(file_name))
     except PermissionError:
         raise BaseSpaceDownloadError('Could not save file to {}, due to insufficient permissions'.format(file_name))
 
+    # Check file size.
```
dblenkus: It would make sense to move validation to a separate function. And I would call it from the wrapper function.

resolwe_bio/tools/basespace_download.py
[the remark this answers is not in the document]
hadalin: >Is BaseSpaceDownloadError raised if network error occurs? I assume it is the primary reason for failed downloads.

Failed downloads happened when the process ended successfully. That means requests didn't raise any exceptions.

resolwe_bio/tools/basespace_download.py
```diff
@@ -141,12 +141,12 @@ def get_api_url():
 
 def get_api_file_url(file_id):
     """Get BaseSpace API file URL."""
-    return '{}/files/{}'.format(get_api_url(), file_id)
+    return f'{get_api_url()}/files/{file_id}'
```
dblenkus: It would be much more readable if you don't use functions in strings

resolwe_bio/tools/basespace_download.py
```diff
@@ -51,14 +72,16 @@ def main():
...
         for file_id in file_ids:
-            file_name = get_file_name(session, file_id, headers)
-            download_file(session, file_id, file_name, headers)
-            output(args.output, 'filename={}'.format(file_name))
-    except Exception:
+            file_name, file_size = get_file_properties(session, file_id, headers)
+            download_file_repeatedly(args.tries, session, file_id, file_name, file_size, headers)
+            output(args.output, f'filename={file_name}')
+    except Exception as error:
+        print(str(error))
```
dblenkus: Put this print in the `else` below, so you get traceback or this message, not both.

docs/CHANGELOG.rst
```diff
@@ -6,6 +6,15 @@ All notable changes to this project are documented in this file.
 This project adheres to `Semantic Versioning <http://semver.org/>`_.
 
 
+==========
+Unreleased
+==========
+
+Changed
+-------
+- Made BaseSpace file download more robust
```
dblenkus: Change log should be in present tense.

PR #840, opened 2019-11-04
Add reverse complement process

AnzeLovse:
## Checklist

    before non-breaking changes.
    commit(s).

### Processes ###

  can be performed).
  and publication.

  tests).

### Tests ###


resolwe_bio/processes/support_processors/seqtk_reverse_complement.py
```diff
@@ -0,0 +1,134 @@
...
+        },
+        'resources': {
+            'cores': 1,
+            'memory': 16384,
+        },
+    }
+    data_name = '{{ reads|sample_name|default("?") }}'
+    version = '1.0.0'
+
+
```
romunov: Too many blank lines.

resolwe_bio/processes/support_processors/seqtk_reverse_complement.py
```diff
@@ -0,0 +1,134 @@
+"""Reverse complement reads with Seqtk."""
+import os
+
+from resolwe.process import Process, DataField, FileField, Cmd, ListField, StringField
+
+
+class ReverseComplementSingle(Process):
+    """Reverse complement FASTQ reads file using Seqtk."""
```
romunov: Comment could perhaps include that this process is for single-end.

resolwe_bio/processes/support_processors/seqtk_reverse_complement.py
```diff
@@ -0,0 +1,134 @@
...
+
+    slug = 'seqtk-rev-complement-single'
+    process_type = 'data:reads:fastq:single:seqtk:comp'
+    name = 'Reverse complement FASTQ'
```
romunov: Name could include that this is a single-end process. My suggestion would be to use `(single-end)`, as do other processes in resolwe-bio.
Screenshot from 2019-11-04 21-05-38

resolwe_bio/processes/support_processors/seqtk_reverse_complement.py
```diff
@@ -0,0 +1,134 @@
...
+        (Cmd['cat'][[reads.path for reads in inputs.reads.fastq]] > 'input_reads.fastq.gz')()
+
+        # Reverse complement reads
+        (Cmd['seqtk']['seq', '-r', 'input_reads.fastq.gz'] > f'{name}_complemented.fastq')()
+
+        (Cmd['gzip'][f'{name}_complemented.fastq'])()
+
+        # Save the outputs
+        outputs.fastq = [f'{name}_complemented.fastq.gz']
+
```
romunov: Add one more blank line.

resolwe_bio/processes/support_processors/seqtk_reverse_complement.py
```diff
@@ -0,0 +1,134 @@
...
+class ReverseComplementPaired(Process):
+    """Reverse complement FASTQ reads file using Seqtk."""
```
romunov: Comment could include that this process works on paired-end data.

resolwe_bio/processes/support_processors/seqtk_reverse_complement.py
```diff
@@ -0,0 +1,134 @@
...
+
+    slug = 'seqtk-rev-complement-paired'
+    process_type = 'data:reads:fastq:paired:seqtk:comp'
+    name = 'Reverse complement FASTQ (Paired)'
```
romunov: My suggestion would be to use `(paired-end)` as do other processes on the platform.

resolwe_bio/processes/support_processors/seqtk_reverse_complement.py
```diff
@@ -0,0 +1,134 @@
...
+    requirements = {
+        'expression-engine': 'jinja',
+        'executor': {
+            'docker': {
+                'image' : 'resolwebio/common:1.1.0'
```
romunov: Redundant space in front of the colon.

resolwe_bio/processes/support_processors/seqtk_reverse_complement.py
```diff
@@ -0,0 +1,134 @@
...
+
+    class Input:
+        """Input fields to process ReverseComplementPaired."""
+
+        reads = DataField('reads:fastq:paired', label='Reads')
+        select_mate = StringField(
+            label='Select mate',
+            description='Select the which mate should be reverse complemented.',
+            choices=[
+                ('Mate 1', 'Mate 1'),
```
romunov: Would computer readable name be in any way better, i.e. `('mate1', 'Mate 1')`?

resolwe_bio/processes/support_processors/seqtk_reverse_complement.py
```diff
@@ -0,0 +1,134 @@
...
+                ('Mate 2', 'Mate 2'),
+                ('Both', 'Both')
+            ],
+            default='Mate 1',
+        )
+
+    class Output:
+        """Output fields."""
+
+        fastq = ListField(FileField(), label='Reverse complemented FASTQ file')
```
romunov: I would name it `fastq1` just to be a little bit more explicit as you seem to follow the 1/2 convention in other variable names.

resolwe_bio/processes/support_processors/seqtk_reverse_complement.py
```diff
@@ -0,0 +1,134 @@
...
+        elif inputs.select_mate == 'Mate 2':
+            # Reverse complement mate 2
+            (Cmd['seqtk']['seq', '-r', f'{name_mate2}_original.fastq.gz'] > f'{name_mate2}_complemented.fastq')()
+            (Cmd['gzip'][f'{name_mate2}_complemented.fastq'])()
+
+            # Save the outputs
+            outputs.fastq = [f'{name_mate1}_original.fastq.gz']
+            outputs.fastq2 = [f'{name_mate2}_complemented.fastq.gz']
+
+        else:
```
romunov: Personally I would explicitly do three `if` statements, something along the lines of

```
if inputs.select_mate == "Mate 1":
    pass

if inputs.select_mate == "Mate 2":
    pass

if inputs.select_mate == "Both":
    pass
```
... just because it explicitly speaks to choices in the process preamble. I had to take a few minutes and see what the last `else` does (because there's a lot of repeating code, differing in small way between `if/else` clauses) because I wasn't very diligent in reading what choices there are. Once I understood the code I checked the `select_mate` options, it made sense.

resolwe_bio/processes/support_processors/seqtk_reverse_complement.py
```diff
@@ -0,0 +1,134 @@
...
+            # Reverse complement mate 1
+            (Cmd['seqtk']['seq', '-r', f'{name_mate1}_original.fastq.gz'] > f'{name_mate1}_complemented.fastq')()
+            (Cmd['gzip'][f'{name_mate1}_complemented.fastq'])()
+            
+            # Reverse complement mate 2
+            (Cmd['seqtk']['seq', '-r', f'{name_mate2}_original.fastq.gz'] > f'{name_mate2}_complemented.fastq')()
+            (Cmd['gzip'][f'{name_mate2}_complemented.fastq'])()
+
+            # Save the outputs
```
romunov: Comment IMO redundant.

resolwe_bio/tests/processes/test_support_processors.py
```diff
@@ -548,3 +548,33 @@ def test_umi_tools_dedup(self):
...
+            }
+
+            revcomp_single = self.run_process('seqtk-rev-complement-single', inputs_single)
+
+            self.assertFiles(revcomp_single, 'fastq', ['hs_slamseq_R1_complemented.fastq.gz'],
+                             compression='gzip')
+
+            inputs_paired = {
+                'reads' : paired_reads.id,
+                'select_mate' : 'Mate 1'
```
romunov: Redundant space in front of the colon.

resolwe_bio/tests/processes/test_support_processors.py
```diff
@@ -548,3 +548,33 @@ def test_umi_tools_dedup(self):
 
         dedup_paired = self.run_process('umi-tools-dedup', {'alignment': bam_paired.id})
         self.assertFile(dedup_paired, 'stats', './corall/output/dedup_paired_stats.txt')
+
+    @tag_process('seqtk-rev-complement-single', 'seqtk-rev-complement-paired')
+    def test_reverse_complement(self):
+        with self.preparation_stage():
+            reads = self.run_process('upload-fastq-single', {
```
romunov: Use `single_reads` to discern it explicitly from `paired_reads`.

romunov: You can use `self.prepare_reads` helper function which will match your usage of `self.prepare_paired_reads`.

resolwe_bio/tests/processes/test_support_processors.py
```diff
@@ -548,3 +548,33 @@ def test_umi_tools_dedup(self):
...
+                'reads': reads.id
+            }
+
+            revcomp_single = self.run_process('seqtk-rev-complement-single', inputs_single)
+
+            self.assertFiles(revcomp_single, 'fastq', ['hs_slamseq_R1_complemented.fastq.gz'],
+                             compression='gzip')
+
+            inputs_paired = {
+                'reads' : paired_reads.id,
```
romunov: Redundant space in front of the colon.

resolwe_bio/processes/support_processors/seqtk_reverse_complement.py
```diff
@@ -0,0 +1,134 @@
...
+    """Reverse complement FASTQ reads file using Seqtk."""
+
+    slug = 'seqtk-rev-complement-single'
+    process_type = 'data:reads:fastq:single:seqtk:comp'
+    name = 'Reverse complement FASTQ'
+    requirements = {
+        'expression-engine': 'jinja',
+        'executor': {
+            'docker': {
+                'image' : 'resolwebio/common:1.1.0'
```
romunov: Redundant space in front of the colon.

resolwe_bio/processes/support_processors/seqtk_reverse_complement.py
```diff
@@ -0,0 +1,134 @@
...
+import os
+
+from resolwe.process import Process, DataField, FileField, Cmd, ListField, StringField
+
+
+class ReverseComplementSingle(Process):
+    """Reverse complement FASTQ reads file using Seqtk."""
+
+    slug = 'seqtk-rev-complement-single'
+    process_type = 'data:reads:fastq:single:seqtk:comp'
```
JureZmrzlikar: I think `data:reads:fastq:single:seqtk` should suffice. Other read processing processes have types:

- `data:reads:fastq:single:bbduk`
- `data:reads:fastq:single:cutadapt`
- ...

The point of process type is to restrict data usage for upstream processes. IMO, one level od depth is enough, but i might be missing something.

resolwe_bio/processes/support_processors/seqtk_reverse_complement.py
JureZmrzlikar: Any reason not to use the [latest version](https://hub.docker.com/r/resolwebio/common/tags) of `resolwebio/common`?

resolwe_bio/processes/support_processors/seqtk_reverse_complement.py
```diff
@@ -0,0 +1,134 @@
...
+    requirements = {
+        'expression-engine': 'jinja',
+        'executor': {
+            'docker': {
+                'image' : 'resolwebio/common:1.1.0'
+            },
+        },
+        'resources': {
+            'cores': 1,
+            'memory': 16384,
```
JureZmrzlikar: What is the memory footprint of this process? We will be dealing with really large files ~10GB and more, so make sure that reserved memory is sufficient. It is difficult to perform such test locally, but did you maybe checked how memory requirements grow when increasing file size? Linearly, constant?

resolwe_bio/processes/support_processors/seqtk_reverse_complement.py
```diff
@@ -0,0 +1,134 @@
...
+
+    class Input:
+        """Input fields to process ReverseComplementSingle."""
+
+        reads = DataField('reads:fastq:single', label='Reads')
+
+    class Output:
+        """Output fields."""
+
+        fastq = ListField(FileField(), label='Reverse complemented FASTQ file')
```
JureZmrzlikar: We might want to add same outputs as in [other processes](https://github.com/genialis/resolwe-bio/blob/master/resolwe_bio/processes/import_data/seq_reads.yml#L33) of type `data:reads:fastq`, i.e `fastqc_url`, `fastqc_archive`. This is the whole point of processes that share the same process type: to have the same set of outputs.
@jkokosar an comment more on this.

resolwe_bio/processes/support_processors/seqtk_reverse_complement.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def run(self, inputs, outputs):
+        """Run the analysis."""
+        name = os.path.basename(inputs.reads.fastq[0].path).strip('.fastq.gz')
```
JureZmrzlikar: Very dangerous. Check what is the output of `'gaaaaaz_my_fastq.fastq.gz'.strip('fastq.gz')`.
One way of handling this::

    basename = os.path.basename(inputs.reads.fastq[0].path)
    assert basename.endswith('fastq.gz')
    name = basename[:-8]

resolwe_bio/processes/support_processors/seqtk_reverse_complement.py
```diff
@@ -0,0 +1,134 @@
...
+
+        # Concatenate multilane reads
```
JureZmrzlikar: Usually, comments are not describing what code does, but why it does if not obvious from the context. If it is obvious what the code does (it should be, 99% of the time if code is well written) the comment is unnecessary. IMO in this and the cases below it is safe to remove the comments.

resolwe_bio/processes/support_processors/seqtk_reverse_complement.py
JureZmrzlikar: Not really. We use the same notation in [other places](https://github.com/genialis/resolwe-bio/blob/master/resolwe_bio/descriptors/sample.yml#L16) too.

resolwe_bio/processes/support_processors/seqtk_reverse_complement.py
JureZmrzlikar: I disagree. This is consistent with other processes of the same type, for example  [upload-fastq-paired](https://github.com/genialis/resolwe-bio/blob/master/resolwe_bio/processes/import_data/seq_reads.yml#L136)

resolwe_bio/processes/support_processors/seqtk_reverse_complement.py
```diff
@@ -0,0 +1,134 @@
...
+
+    class Output:
+        """Output fields."""
+
+        fastq = ListField(FileField(), label='Reverse complemented FASTQ file')
+        fastq2 = ListField(FileField(), label='Remaining mate')
+
+    def run(self, inputs, outputs):
+        """Run the analysis."""
+        name_mate1 = os.path.basename(inputs.reads.fastq[0].path).strip('.fastq.gz')
```
JureZmrzlikar: See comment above.

resolwe_bio/processes/support_processors/seqtk_reverse_complement.py
JureZmrzlikar: I disagree. You suggestion implies that all three options could be true at the same time. The nature of if-elif-else is handling three exclusive disjunct options and this is exactly what we are handling with.

resolwe_bio/tests/processes/test_support_processors.py
```diff
@@ -548,3 +548,33 @@ def test_umi_tools_dedup(self):
...
+    def test_reverse_complement(self):
+        with self.preparation_stage():
+            reads = self.run_process('upload-fastq-single', {
+                'src': ['hs_slamseq_R1.fastq.gz']
+            })
+
+            paired_reads = self.prepare_paired_reads(['hs_slamseq_R1.fastq.gz'],
+                                                     ['hs_slamseq_R2.fastq.gz'])
+
+            inputs_single = {
```
JureZmrzlikar: It makes sense to put inputs close to the process that uses them. I suggest the following syntax:

    revcomp_single = self.run_process('seqtk-rev-complement-single', {'reads': reads.id})

and:

    revcomp_paired = self.run_process('seqtk-rev-complement-paired', {
        'reads' : paired_reads.id,
        'select_mate' : 'Mate 1',
    })

resolwe_bio/tests/processes/test_support_processors.py
```diff
@@ -548,3 +548,33 @@ def test_umi_tools_dedup(self):
...
+            self.assertFiles(revcomp_single, 'fastq', ['hs_slamseq_R1_complemented.fastq.gz'],
+                             compression='gzip')
+
+            inputs_paired = {
+                'reads' : paired_reads.id,
+                'select_mate' : 'Mate 1'
+            }
+
+            revcomp_paired = self.run_process('seqtk-rev-complement-paired', inputs_paired)
+            self.assertFiles(revcomp_paired, 'fastq', ['hs_slamseq_R1_complemented.fastq.gz'],
```
JureZmrzlikar: This line can fit into 119 chars limit, which is the limit for this repo. It is more compact and easier to read if it is just in one line IMO. But this is a matter of style, so i leave it up to you.

resolwe_bio/processes/support_processors/seqtk_reverse_complement.py
romunov: My reasoning was that this is handled on the `StringField` as I am assuming only one choice can be selected at a time.

resolwe_bio/processes/support_processors/seqtk_reverse_complement.py
AnzeLovse: I was only able to test it locally on read files with 5GB size.  According to the results of [this](https://bioinf.shenwei.me/seqkit/benchmark/) benchmarking the RAM usage is usually smaller than the file size.

JureZmrzlikar: OK.

resolwe_bio/tests/processes/test_support_processors.py
```diff
@@ -548,3 +548,23 @@ def test_umi_tools_dedup(self):
...
         self.assertFile(dedup_paired, 'stats', './corall/output/dedup_paired_stats.txt')
+
+    @tag_process('seqtk-rev-complement-single', 'seqtk-rev-complement-paired')
+    def test_reverse_complement(self):
+        with self.preparation_stage():
+            single_reads = self.prepare_reads(['hs_slamseq_R1.fastq.gz'])
+
+            paired_reads = self.prepare_paired_reads(['hs_slamseq_R1.fastq.gz'], ['hs_slamseq_R2.fastq.gz'])
+
+            revcomp_single = self.run_process('seqtk-rev-complement-single', {'reads': single_reads.id})
```
JureZmrzlikar: This should be one indentation level less than it is. The `preparation_stage` context manager only prepares input data objects, but the process that is tested should be put outside of it (and also all assert statements that follow it).

docs/CHANGELOG.rst
```diff
@@ -13,6 +13,7 @@ Unreleased
 Added
 -----
 - Add ``resolwebio/slamdunk`` Docker image
+- Add ``seqtk-rev-complement-single`` process
```
jkokosar: Two processes were added, so modify to something like: Add ``seqtk-rev-complement-single`` and ``seqtk-rev-complement-paired`` processes

resolwe_bio/processes/support_processors/seqtk_reverse_complement.py
```diff
@@ -0,0 +1,185 @@
...
+        """Run the analysis."""
+        basename = os.path.basename(inputs.reads.fastq[0].path)
+        assert basename.endswith('.fastq.gz')
+        name = basename[:-9]
+
+        # Concatenate multilane reads
+        (Cmd['cat'][[reads.path for reads in inputs.reads.fastq]] > 'input_reads.fastq.gz')()
+
+        # Reverse complement reads
+        (Cmd['seqtk']['seq', '-r', 'input_reads.fastq.gz'] > f'{name}_complemented.fastq')()
```
romunov: Think about putting `f'{name}_complemented.fastq'` into a variable as it repeats several times.

resolwe_bio/processes/support_processors/seqtk_reverse_complement.py
```diff
@@ -0,0 +1,185 @@
...
+        assert basename_mate1.endswith('.fastq.gz')
+        assert basename_mate2.endswith('.fastq.gz')
+        name_mate1 = basename_mate1[:-9]
+        name_mate2 = basename_mate2[:-9]
+
+        (Cmd['cat'][[reads.path for reads in inputs.reads.fastq]] > f'{name_mate1}_original.fastq.gz')()
+        (Cmd['cat'][[reads.path for reads in inputs.reads.fastq2]] > f'{name_mate2}_original.fastq.gz')()
+
+        if inputs.select_mate == 'Mate 1':
+            (Cmd['seqtk']['seq', '-r', f'{name_mate1}_original.fastq.gz'] > f'{name_mate1}_complemented.fastq')()
```
romunov: Think about adding output strings into variables and use those throughout the rest of the process.

jkokosar:
Please rebase to the latest master branch

jkokosar:
Please rebase to the latest master. The ROSE2 issue has been fixed.

PR #841, opened 2019-11-05
Add slamdunk process

AnzeLovse:
## Checklist

    before non-breaking changes.
    commit(s).

### Processes ###

  can be performed).
  and publication.

  tests).

### Tests ###


resolwe_bio/processes/expression/slamdunk.py
```diff
@@ -0,0 +1,105 @@
...
+from plumbum import TEE
+
+from resolwe.process import Process, Cmd, DataField, BooleanField, IntegerField, FileField, DirField
+
+
+class SlamdunkAllPaired(Process):
+    """Run slamdunk for paired-end reads with alignment to transcriptome."""
+
+    slug = 'slamdunk-all-paired'
+    process_type = 'data:reads:fastq:paired'
```
jkokosar: Change to `process_type = 'data:slamdunk:all'`

resolwe_bio/processes/expression/slamdunk.py
```diff
@@ -0,0 +1,105 @@
...
+            },
+        },
+        'resources': {
+            'cores': 8,
+            'memory': 16384,
+        },
+    }
+    data_name = '{{ reads|sample_name|default("?") }}'
+    version = '1.0.0'
+    process_type = 'data:expression:slamdunk'
```
jkokosar: Duplicated field. Also, `data:expression:..` type already exists, and this process does not match it with its defined output fields.

resolwe_bio/processes/expression/slamdunk.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+    class Input:
+        """Input fields for SlamdunkAllPaired."""
+
+        reads = DataField('reads:fastq:paired', label='Reads')
+        transcriptome = DataField('genome:fasta', label='FASTA file containig transcripts')
```
jkokosar: Rather use input type `data:seq:nucleotide`. Data type `genome:fasta` also works, but the process that prepares this type of Data object also computes indices for aligner tools (except for STAR aligner).

resolwe_bio/processes/expression/slamdunk.py
```diff
@@ -0,0 +1,105 @@
...
+        assert basename.endswith('fastq.gz')
+        name = basename[:-9]
+
+        reads_paths = [reads.path for reads in inputs.reads.fastq] + [reads.path for reads in inputs.reads.fastq2]
+        (Cmd['ln']['-s', f'{inputs.transcriptome.fasta.path}', 'transcriptome.fasta'])()
+        (Cmd['cat'][reads_paths] > f'{name}_concatenated.fastq.gz')()
+
+        args = [
+            '-r', 'transcriptome.fasta',
+            '-b', f'{inputs.regions.bed.path}',
```
jkokosar: I think `f'{inputs.regions.bed.path}'` can be just `inputs.regions.bed.path`

resolwe_bio/tests/processes/test_expression.py
```diff
@@ -743,3 +743,30 @@ def test_stringtie(self):
...
+                'src': os.path.join('slamseq', 'input', 'hs_transcript.fasta'),
+                'species': 'Homo sapiens',
+                'build': 'Gencode 32'
+            })
+            bedfile = self.run_process('upload-bed', {
+                'src': os.path.join('slamseq', 'input', 'hs_transcript.bed'),
+                'species': 'Homo sapiens',
+                'build': 'Gencode 32'
+            })
+            inputs = {
```
jkokosar: Use one indentation level less for the block of code from here until the end.

resolwe_bio/tests/processes/test_expression.py
```diff
@@ -743,3 +743,30 @@ def test_stringtie(self):
...
         self.assertFields(stringtie, 'build', 'ens_90')
         self.assertFields(stringtie, 'feature_type', 'gene')
+    
+    @with_resolwe_host
+    @tag_process('slamdunk-all-paired')
+    def test_slamdunk_paired(self):
+        with self.preparation_stage():
+            paired_reads = self.prepare_paired_reads(['hs_slamseq_R1_complemented.fastq.gz'],
+                                                     ['hs_slamseq_R2.fastq.gz'])
+            transcripts = self.run_process('upload-genome', {
```
jkokosar: `self.run_process('upload-fasta-nucl', ...)`

resolwe_bio/processes/expression/slamdunk.py
```diff
@@ -0,0 +1,105 @@
+import os
+
+from plumbum import TEE
+
+from resolwe.process import Process, Cmd, DataField, BooleanField, IntegerField, FileField, DirField
+
+
+class SlamdunkAllPaired(Process):
+    """Run slamdunk for paired-end reads with alignment to transcriptome."""
```
jkokosar: I would avoid the `transcriptome` notation here, since its possible to run the tool against any nucleotide sequence + bed regions file (as in the case of QUANT-Seq data).

resolwe_bio/processes/expression/slamdunk.py
```diff
@@ -0,0 +1,105 @@
...
+    version = '1.0.0'
+    process_type = 'data:expression:slamdunk'
+
+
+    class Input:
+        """Input fields for SlamdunkAllPaired."""
+
+        reads = DataField('reads:fastq:paired', label='Reads')
+        transcriptome = DataField('genome:fasta', label='FASTA file containig transcripts')
+        regions = DataField('bed', label='BED file with coordinates of regions of interes')
```
romunov: Missing t at the end.

resolwe_bio/processes/expression/slamdunk.py
```diff
@@ -0,0 +1,105 @@
...
+            'cores': 8,
+            'memory': 16384,
+        },
+    }
+    data_name = '{{ reads|sample_name|default("?") }}'
+    version = '1.0.0'
+    process_type = 'data:expression:slamdunk'
+
+
+    class Input:
```
romunov: Some descriptions end in a full stop, others do not. Make it uniform.

resolwe_bio/processes/expression/slamdunk.py
```diff
@@ -0,0 +1,105 @@
...
+
+        Cmd['zip']['-r', f'{name}_map.zip', os.path.join('slamdunk_all', 'map')]()
+        Cmd['zip']['-r', f'{name}_filter.zip', os.path.join('slamdunk_all', 'filter')]()
+        Cmd['zip']['-r', f'{name}_snp.zip', os.path.join('slamdunk_all', 'snp')]()
+        Cmd['zip']['-r', f'{name}_count.zip', os.path.join('slamdunk_all', 'count')]()
+
+        outputs.tcount = os.path.join(
+            'slamdunk_all',
+            'count',
+            f'{name}_concatenated.fastq_slamdunk_mapped_filtered_tcount.tsv'
```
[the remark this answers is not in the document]
AnzeLovse: I completely agree, this is due to the aligner step, that only removes the last suffix .gz in this case. One option is to decompress reads before running it or renaming the output files. Should I change that?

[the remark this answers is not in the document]
romunov: This will not likely improve the code in any significant way, just skip it (for now).

resolwe_bio/processes/expression/slamdunk.py
```diff
@@ -0,0 +1,103 @@
...
+        reads_paths = [reads.path for reads in inputs.reads.fastq] + [reads.path for reads in inputs.reads.fastq2]
+        (Cmd['ln']['-s', inputs.transcriptome.fasta.path, 'transcriptome.fasta'])()
+        concatenated_name = f'{name}_concatenated.fastq.gz'
+        (Cmd['cat'][reads_paths] > concatenated_name)()
+
+        args = [
+            '-r', 'transcriptome.fasta',
+            '-b', inputs.regions.bed.path,
+            '-o', 'slamdunk_all',
+            '-n', f'{inputs.max_alignments}',
```
jkokosar: `f'{inputs.max_alignments}` -> `inputs.max_alignments`
Fix also other occurances

PR #844, opened 2019-11-08
Append sample info to summary in filtering-chemut

jkokosar:
## Checklist

    before non-breaking changes.
    commit(s).

### Processes ###

  can be performed).
  and publication.

  tests).

### Tests ###

### DIFF
--- a/resolwe_bio/processes/variant_calling/filtering_chemut.yml
+++ b/resolwe_bio/processes/variant_calling/filtering_chemut.yml
@@ -9,14 +9,15 @@
     expression-engine: jinja
     executor:
       docker:
-        image: resolwebio/dnaseq:4.1.0
+        image: resolwebio/dnaseq:4.2.0
   data_name: "Filtered variants ({{ analysis_type }})"
-  version: 1.3.1
+  version: 1.4.0
   type: data:variants:vcf:filtering
   category: Other
   persistence: CACHED
   description: |
-    Filtering and annotation of Variant Calling data - Chemical mutagenesis in Dictyostelium discoideum
+    Filtering and annotation of Variant Calling data - Chemical
+    mutagenesis in _Dictyostelium discoideum_.
   input:
     - name: variants
       label: Variants file (VCF)
@@ -25,7 +26,11 @@
       label: Analysis type
       type: basic:string
       description: |
-        Choice of the analysis type. Use "SNV" or "INDEL" options for the analysis of haploid VCF files prepared by using GATK UnifiedGenotyper -glm option "SNP" or "INDEL", respectively. Choose options SNV_CHR2 or INDEL_CHR2 to run the GATK analysis only on the diploid portion of CHR2 (-ploidy 2 -L chr2:2263132-3015703).
+        Choice of the analysis type. Use "SNV" or "INDEL" options for
+        the analysis of haploid VCF files prepared by using
+        GATK UnifiedGenotyper -glm option "SNP" or "INDEL", respectively.
+        Choose options SNV_CHR2 or INDEL_CHR2 to run the GATK analysis
+        only on the diploid portion of CHR2 (-ploidy 2 -L chr2:2263132-3015703).
       choices:
         - label: SNV
           value: snv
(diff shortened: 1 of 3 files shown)


resolwe_bio/processes/variant_calling/filtering_chemut.yml
```diff
@@ -11,12 +11,13 @@
...
   data_name: "Filtered variants ({{ analysis_type }})"
-  version: 1.3.1
+  version: 1.4.0
   type: data:variants:vcf:filtering
   category: Other
   persistence: CACHED
   description: |
-    Filtering and annotation of Variant Calling data - Chemical mutagenesis in Dictyostelium discoideum
+    Filtering and annotation of Variant Calling data - Chemical
+    mutagenesis in Dictyostelium discoideum
```
romunov: Can we use markdown in here, i.e. _Dictyostelium discoideum_?

resolwe_bio/tools/append_sample_info.py
```diff
@@ -0,0 +1,33 @@
...
+
+parser = argparse.ArgumentParser(description=__doc__)
+parser.add_argument('vcf_file', help="VCF file (can be compressed using gzip/bgzip).")
+parser.add_argument('summary', help="Summary file to append to.")
+args = parser.parse_args()
+
+try:
+    vcf = VariantFile(args.vcf_file)
+except (OSError, ValueError):
+    print(error('Input VCF file does not exist or could not be correctly opened.'))
```
romunov: Is `print` necessary here?

resolwe_bio/tools/append_sample_info.py
```diff
@@ -0,0 +1,33 @@
...
+    exit(1)
```
romunov: Would this work in place of `exit(1)`?

```
error_msg = 'Input VCF file does not exist or could not be correctly opened.'
print(error_msg)
raise ValueError(error_msg)
```

resolwe_bio/tools/append_sample_info.py
```diff
@@ -0,0 +1,33 @@
...
+
+vcf_header = vcf.header
+header_records = {record.key: record.value for record in vcf_header.records}
+
+with open(args.summary, "a") as out_file:
+    try:
+        fasta_name = os.path.basename(header_records['reference'])
+    except KeyError:
+        fasta_name = ''
+        print(warning('Reference sequence (FASTA) name could not be recognized from the VCF header.'))
```
romunov: For warnings, I would use [warnings](https://docs.python.org/3/library/warnings.html#module-warnings).

```
import warnings
# ...
warn_msg = 'Reference sequence (FASTA) name could not be recognized from the VCF header.'
print(warning(warn_msg))
warnings.warn(warn_msg)
```

resolwe_bio/processes/variant_calling/filtering_chemut.yml
jkokosar: We can, so I will change this to `_Dictyostelium discoideum_`

resolwe_bio/tools/append_sample_info.py
jkokosar: We want to use the `resolwe_runtime_utils` process error/warning/info system for the resolwe process support scripts. As a standalone script outside of resolwe universe, the suggested approach would make sense.

resolwe_bio/tools/append_sample_info.py
jkokosar: Yes, we need the JSON formatted by `error()` function printed to stdout

resolwe_bio/tools/append_sample_info.py
jkokosar: I have refactored so that both the process error (set manually) and error msg from the parser itself are presented to the user.

resolwe_bio/processes/variant_calling/filtering_chemut.yml
```diff
@@ -9,14 +9,15 @@
...
   data_name: "Filtered variants ({{ analysis_type }})"
-  version: 1.3.1
+  version: 1.4.0
   type: data:variants:vcf:filtering
   category: Other
   persistence: CACHED
   description: |
-    Filtering and annotation of Variant Calling data - Chemical mutagenesis in Dictyostelium discoideum
+    Filtering and annotation of Variant Calling data - Chemical
+    mutagenesis in _Dictyostelium discoideum_
```
romunov: Missing full stop.

PR #851, opened 2019-11-29
Star samtools sort

HudoGriz:
## Checklist
    before non-breaking changes.
    commit(s).

### Description ###
When comparing sorting with STAR and sorting with samtools, we concluded that sorting with samtools is less memory intensive and faster on a HDD. The sorted alignments (outputs) were tested with deepTools for coverage, quantified using fetureCounts and viewed with IGV. In all test we did not observe any difference, between the alignements sorted with STAR or samtools.
I added an option for selecting the sorting process (samtools or STAR), with the default being samtools. I run tests once with samtools sorting and once with STAR sorting. Both tests produced no errors.

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -32,6 +32,8 @@ Changed
 - Freeze docutils package version to 0.15.2 because Sphinx has
   problems parsing development version numbers
 - Support Slamdunk/Alleyoop processes in MultiQC
+- Enable sorting of files in ``alignment-star`` process using samtools
+
 
 ===================
 24.0.0 - 2019-11-15
--- a/resolwe_bio/processes/alignment/star.yml
+++ b/resolwe_bio/processes/alignment/star.yml
@@ -14,7 +14,7 @@
       memory: 36864
       cores: 10
   data_name: "{{ reads|sample_name|default('?') }}"
-  version: 1.8.12
+  version: 1.9.0
   type: data:alignment:bam:star
   category: Align
   flow_collection: sample
@@ -305,6 +305,12 @@
           description: |
             Size of the bins, in bases, for the output of the bigwig. Only possible if 'Tool to calculate BigWig'
             is deepTools. If BigWig is calculated by UCSC BedGraphToBigWig then bin size is 1.
+    - name: star_sort
+      label: Sorting with STAR
+      type: basic:boolean
+      default: false
+      description: |
+        Set to false for sorting with samtools or to true for sorting with STAR which may be time and memory intensive.
   output:
     - name: bam
       label: Alignment file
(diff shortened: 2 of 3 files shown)


resolwe_bio/processes/alignment/star.yml
```diff
@@ -467,20 +473,39 @@
...
-        re-checkrc "Samtools sort command failed."
-        samtools index Aligned.sortedByCoord_temp.out.bam "${MATE1_NAME}.bam.bai"
-        re-checkrc "Samtools index command failed."
-        mv Aligned.sortedByCoord_temp.out.bam "${MATE1_NAME}.bam"
-      else
-        samtools index Aligned.sortedByCoord.out.bam "${MATE1_NAME}.bam.bai"
-        re-checkrc "Samtools index command failed."
-        mv Aligned.sortedByCoord.out.bam "${MATE1_NAME}.bam"
-      fi
+      {% if star_sort %}
```
romunov: `if [[ $(samtools view -c Aligned.sortedByCoord.out.bam) == 0 ]]` has been introduced due to a bug in STAR aligner. This has since been fixed. If we don't need this if/else anymore, perhaps we could simplify it to something along the lines of

```
OUT_NAME = "${MATE1_NAME}.bam"
OUT_NAME_BAI = "${MATE1_NAME}.bai"
OUT_NAME_TEMP = "${OUT_NAME}_temp"

{% if star_sort %}
    IN_NAME = "Aligned.sortedByCoord.out.bam"
{% else %}
    IN_NAME = "Aligned.out.bam"
{% endif %}

samtools sort "${IN_NAME}" > "${OUT_NAME_TEMP}
re-checkrc "Samtools sort command failed."
samtools index "${OUT_NAME_TEMP}" "${OUT_NAME_BAI}"
re-checkrc "Samtools index command failed."
mv "${OUT_NAME_TEMP}" "${OUT_NAME}"
```

docs/CHANGELOG.rst
```diff
@@ -14,6 +14,7 @@ Changed
 -------
 - Append sample and genome reference information to the summary output
   file in the ``filtering-chemut`` process
+- Sorting of files in ``alignment-star`` process is done via samtools
```
romunov: How about "Enable sorting of files in ``alignment-star`` process using samtools"? This way we convey the message that sorting can be performed using either tool

resolwe_bio/processes/alignment/star.yml
```diff
@@ -305,6 +305,12 @@
           description: |
             Size of the bins, in bases, for the output of the bigwig. Only possible if 'Tool to calculate BigWig'
             is deepTools. If BigWig is calculated by UCSC BedGraphToBigWig then bin size is 1.
+    - name: star_sort
+      label: Sorting with STAR
+      type: basic:boolean
+      default: false
+      description: |
+        Set this to true to force sorting with STAR, which might be slower and more memory demanding.
```
romunov: Please mention that sorting is done using samtools if false. Maybe something like

> Sort using STAR (true) which may be time and memory intensive or samtools (false)

PR #853, opened 2019-12-05
Make bigwig field not mandatory for bamclipper, bqsr and markduplicates

romunov:
Cases, where bigwig file was not calculated, stopped. This PR makes the output field of bigwig not required and should solve this problem.

Sphinx is having problems parsing docstrings so we're freezing the docutils to version 0.15.2.

## Checklist

    before non-breaking changes.
    commit(s).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,6 +55,7 @@ setuptools.setup(
             # work with our custom page template.
             'Sphinx~=1.5.6',
             'sphinx_rtd_theme',
+            'docutils==0.15.2',
         ],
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -55,6 +55,7 @@
             # work with our custom page template.
             'Sphinx~=1.5.6',
             'sphinx_rtd_theme',
+            'docutils==0.15.2',
```
dblenkus: Please add a comment why this is required.

PR #854, opened 2019-12-05
Explain why docutils has been frozen to 0.15.2

romunov:
Sphinx has problems parsing the dev version of `docutils` (0.16b.0). For the time being, we are freezing the version to 0.15.2.

    before non-breaking changes.
    commit(s).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,6 +55,8 @@ setuptools.setup(
             # work with our custom page template.
             'Sphinx~=1.5.6',
             'sphinx_rtd_theme',
+            # XXX: Temporarily freeze docutils to 0.15.2 since Sphinx has
+            # problems parsing dev version (0.16b).
             'docutils==0.15.2',
         ],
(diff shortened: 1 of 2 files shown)


romunov:
That is indeed a better way. Fixing.

setup.py
```diff
@@ -55,6 +55,8 @@
             # work with our custom page template.
             'Sphinx~=1.5.6',
             'sphinx_rtd_theme',
+            # XXX: Temporarily freeze docutils to 0.15.2 since Sphinx has
```
dblenkus: The main problem is that sphinx has [too open requirements](https://github.com/sphinx-doc/sphinx/blob/fb9adbf0be0d136cc61ad5cc7b45dae62f133aa9/setup.py#L26) and doesn't work with them. The second problem is that we install pre-release versions of all dependencies instead of just for Resolwe.

PR #858, opened 2019-12-16
Handle improper read_group tags in bqsr process

romunov:
A user might have incorrectly specify tag of the `read_group` argument. This PR checks that all inputs have the `-XX` structure.
Checks now goes from broad to specific, i.e. first it makes sure all tags exist/are correctly entered, there are no duplicates and finally if mandatory tags are used.

## Checklist
    before non-breaking changes.
    commit(s).

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -18,6 +18,8 @@ Changed
 Fixed
 -----
 - Fix typo in ``scheduling_class`` variable in several Python processes
+- Handle cases of improper tags passed to ``read_group`` argument of
+  the ``bqsr`` process
 
 
 ===================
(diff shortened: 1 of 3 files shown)


resolwe_bio/processes/reads_processing/bqsr.py
```diff
@@ -18,7 +18,7 @@ class BQSR(Process):
     slug = 'bqsr'
     name = 'BaseQualityScoreRecalibrator'
     process_type = 'data:alignment:bam:bqsr:'
-    version = '1.1.1'
+    version = '1.2.1'
```
jkokosar: Should be `version = '1.2.0'`

romunov: Fixed.

docs/CHANGELOG.rst
```diff
@@ -34,6 +34,8 @@ Changed
 - Support merging of multi-lane sequencing data into a single (pair) of
   FASTQ files in the ``upload-fastq-single``, ``upload-fastq-paired``,
   ``files-to-fastq-single`` and ``files-to-fastq-paired`` processes
+- Handle cases of improper tags passed to ``read_group`` argument of
```
jkokosar: Please move this to the unreleased section and rebase.

PR #862, opened 2019-12-23
Rewrite MultiQC process to Python

jkokosar:
## Checklist

    before non-breaking changes.
    commit(s).

### Processes ###

  can be performed).
  and publication.

  tests).

### Tests ###


resolwe_bio/processes/support_processors/multiqc.py
```diff
@@ -0,0 +1,242 @@
...
+
+def create_symlink(src, dst):
+    """Create a symbolic link."""
+    return Cmd['ln']('-s', '--backup=numbered', src, dst)
+
+
+def create_summary_table(samples, species, build):
+    """Prepare sample summary MultiQC table."""
+    sample_summary = {
+        "id": "sample_summary",
```
romunov: Single quotes?

resolwe_bio/processes/support_processors/multiqc.py
```diff
@@ -0,0 +1,242 @@
...
+
+    with open('sample_data_mqc.json', 'w') as out_file:
+        json.dump(sample_summary, out_file)
+
+
+class Multiqc(Process):
+    """Aggregate results from bioinformatics analyses across many samples into a single report.
+
+    [MultiQC](http://www.multiqc.info) searches a given directory for analysis logs and compiles a
+    HTML report. It's a general use tool, perfect for summarising the output from numerous
```
romunov: Maybe "general purpose tool"?

romunov: Numerous could also be "various". Can't say which one I'd use, though. :)

resolwe_bio/processes/support_processors/multiqc.py
```diff
@@ -0,0 +1,242 @@
...
+                label='--dirs',
+                default=True,
+                description='Prepend directory to sample names.',
+            )
+
+            dirs_depth = IntegerField(
+                label='--dirs-depth',
+                default=-1,
+                description='Prepend a specified number of directories to sample names. Enter a '
+                            'negative number to take from start of path.',
```
romunov: Maybe "negative number (default)... "?

resolwe_bio/processes/support_processors/multiqc.py
```diff
@@ -0,0 +1,242 @@
...
+            )
+
+            fullnames = BooleanField(
+                label='--fullnames',
+                default=False,
+                description='Do not clean the sample names (leave as full file name).',
```
romunov: I would write what this flag does (and that it's default is False) and not what it doesn't do.

resolwe_bio/processes/support_processors/multiqc.py
```diff
@@ -0,0 +1,242 @@
...
+        unsupported_data = []
+
+        for d in inputs.data:
+            sample_dir = d.entity_name
+
+            if sample_dir and not os.path.isdir(sample_dir):
+                os.mkdir(sample_dir)
+
+            try:
+                if d.entity_name and d.species and d.build:
```
romunov: `d.entity_name` could be `sample_dir`.

resolwe_bio/processes/support_processors/multiqc.py
```diff
@@ -0,0 +1,242 @@
...
+                    samples.append(d.entity_name)
```
romunov: Ibidem.

resolwe_bio/processes/support_processors/multiqc.py
```diff
@@ -0,0 +1,242 @@
...
+                    species.append(d.species)
+                    build.append(d.build)
+            except AttributeError:
+                pass
```
romunov: Perhaps issue a `self.warning`?

resolwe_bio/processes/support_processors/multiqc.py
```diff
@@ -0,0 +1,242 @@
...
+
+            if d.type.startswith('data:reads:fastq:single'):
+                for fq_report in d.fastqc_archive:
+                    name = os.path.basename(fq_report.path)
+                    create_symlink(fq_report.path, os.path.join(sample_dir, name))
```
romunov: How about naming arguments explicitly for arguably clarity? I.e.

```
create_symlink(src=fq_report.path, dst=os.path.join(sample_dir, name))
```

resolwe_bio/processes/support_processors/multiqc.py
```diff
@@ -0,0 +1,242 @@
...
+
+            elif d.type.startswith('data:reads:fastq:paired'):
+                for fq_report in d.fastqc_archive + d.fastqc_archive2:
+                    name = os.path.basename(fq_report.path)
+                    create_symlink(fq_report.path, os.path.join(sample_dir, name))
+
+            elif d.type == 'data:alignment:bam:star:':
```
romunov: This string ends with `:` - is this by design?

resolwe_bio/processes/support_processors/multiqc.py
```diff
@@ -0,0 +1,242 @@
...
+                stats_file = os.path.basename(d.stats.path)
+                assert stats_file.endswith('_stats.txt')
+                bam_name = stats_file[:-10]
+
+                if d.build == 'rRNA':
+                    rrna_report = f'{bam_name}.rRNA.Log.final.out'
```
romunov: Would underscores make filename any more clear?

```
f'{bam_name}.rRNA.Log.final.out'
f'{bam_name}_rRNA_Log_final.out'
```

Also consider if you want to introduce capital letters to the filename (i.e. `Log`). It makes sense to capitalize RNA, but I'm not sure about Log.
Also, is final relevant here? Are there non-final reports being outputted?

resolwe_bio/processes/support_processors/multiqc.py
```diff
@@ -0,0 +1,242 @@
...
+                    globin_report = f'{bam_name}.globin.Log.final.out'
+                    create_symlink(d.stats.path, os.path.join(sample_dir, globin_report))
+                else:
+                    report = f'{bam_name}.Log.final.out'
+                    create_symlink(d.stats.path, os.path.join(sample_dir, report))
+
+            elif d.type.startswith('data:alignment:bam'):
+                name = os.path.basename(d.stats.path)
+                create_symlink(d.stats.path, os.path.join(sample_dir, name))
+
```
romunov: Remove blank line like in above ifelse chunk.

resolwe_bio/processes/support_processors/multiqc.py
```diff
@@ -0,0 +1,242 @@
...
+            elif d.type == 'data:expression:featurecounts:':
```
romunov: When checking for type, some strings end with a colon while others do not. Is this by design?

resolwe_bio/processes/support_processors/multiqc.py
```diff
@@ -0,0 +1,242 @@
...
+                create_symlink(d.report.path, sample_dir)
+                # Alleyoop summary may contain PCA plot data
+                if os.path.isfile(d.plot_data.path):
+                    create_symlink(d.plot_data.path, sample_dir)
+
+            else:
+                unsupported_data.append(d.id)
+
+        if unsupported_data:
+            self.warning(f'Data with ID(s) {unsupported_data} is not supported by the MultiQC analysis.')
```
romunov: Couldn't this be moved to line 211 under the final `else`?

```
else:
    unsupported_data.append(d.id)
    self.warning(f'Data with ID(s) {unsupported_data} is not supported by the MultiQC analysis.')
```

I think you can remove "is" to read 

```
self.warning(f'Data with ID(s) {unsupported_data} not supported by the MultiQC analysis.')
```

resolwe_bio/processes/support_processors/multiqc.py
```diff
@@ -0,0 +1,242 @@
...
+
+        if inputs.advanced.cl_config:
+            args.extend(['--cl-config ', inputs.advanced.cl_config])
+
+        with Cmd.env(LC_ALL='C.UTF-8'):
+            return_code, _, _ = Cmd['multiqc']['.'][args] & TEE(retcode=None)
+            if return_code:
+                self.error('MultiQC analysis failed.')
+
+        if os.path.isdir('multiqc_data') and os.path.isfile('multiqc_report.html'):
```
romunov: I would make outputs output in any case, but check for existence of folders/files and fail if they do not exist.

```
if not os.path.isdir('multiqc_data') and not os.path.isfile('multiqc_report.html'):
    self.error('MultiQC finished without creating outputs.')

outputs.report = 'multiqc_report.html'
outputs.report_data = 'multiqc_data'
            
```

resolwe_bio/processes/support_processors/multiqc.py
```diff
@@ -0,0 +1,242 @@
...
+            sample_summary['data'][sample_name] = {
+                'Species': species,
+                'Genome Build': build,
+            }
+
+    with open('sample_data_mqc.json', 'w') as out_file:
+        json.dump(sample_summary, out_file)
+
+
+class Multiqc(Process):
```
romunov: Maybe `MultiQC`?

resolwe_bio/processes/support_processors/multiqc.py
jkokosar: Yes, since `d.type` always ends with `:` and using only `data:alignment:bam:star` would therefore evaluate to false, erroneously skipping the operation on this `Data`.

resolwe_bio/processes/support_processors/multiqc.py
jkokosar: MultiQC recognises the report files from various tools based on either content, or based on output file names. In this case, the `Log.final.out` is required for MultiQC to recognise and parse the input.

resolwe_bio/processes/support_processors/multiqc.py
jkokosar: It does not end with colon where we use `startswith()`, as the colon is not necessary there.

resolwe_bio/processes/support_processors/multiqc.py
jkokosar: Moving this to final `else` would issue as many warnings as there is unsupported Data. In practice this was unreadable in many cases, thus the change to issue just one warning containing a list of unsupported Data IDs.

resolwe_bio/processes/support_processors/multiqc.py
```diff
@@ -0,0 +1,299 @@
...
+    }
+    entity = {
+        'type': 'sample',
+    }
+    category = 'Other'
+    data_name = 'MultiQC report'
+    version = '1.4.0'
+
+    class Input:
+        """Input fields to process ReverseComplementSingle."""
```
JureZmrzlikar: Check description again :-)

resolwe_bio/processes/support_processors/multiqc.py
```diff
@@ -0,0 +1,299 @@
...
+        species = []
+        build = []
+        lib_type_samples = []
+        lib_type_reports = []
+        unsupported_data = []
+
+        for d in inputs.data:
+            sample_dir = d.entity_name
+
+            if sample_dir and not os.path.isdir(sample_dir):
```
JureZmrzlikar: Maybe
```
if sample_dir:
    os.makedirs(sample_dir, exist_ok=True)
```

resolwe_bio/processes/support_processors/multiqc.py
```diff
@@ -0,0 +1,299 @@
...
+
+    lib_strand_json = {
+        "id": "lib_strandedness",
+        "section_name": "Library Strandedness",
+        "plot_type": "table",
+        "file_format": "json",
+        "data": {}
+    }
+
+    for sample in zip(samples, reports):
```
JureZmrzlikar: It also works if you expand the variables directly in for loop:
```
for sample_name, report in zip(samples, reports):
    ....
```

resolwe_bio/processes/support_processors/multiqc.py
```diff
@@ -0,0 +1,299 @@
...
+    """Prepare sample summary MultiQC table."""
+    sample_summary = {
+        'id': 'sample_summary',
+        'section_name': 'Sample Summary',
+        'plot_type': 'table',
+        'file_format': 'json',
+        'data': {}
+    }
+
+    for sample in zip(samples, species, build):
```
JureZmrzlikar: It also works if you expand the variables directly in for loop (see comment below)

PR #863, opened 2020-01-07
Create Sample summary table in MultiQC reports

jkokosar:
A temporary solution until https://github.com/genialis/resolwe-bio/pull/862 is ready.

### DIFF
--- a/resolwe_bio/tools/multiqc_sample_info.py
+++ b/resolwe_bio/tools/multiqc_sample_info.py
@@ -1,0 +1,41 @@
+#!/usr/bin/env python3
+"""Format sample metadata table for MultiQC."""
+import argparse
+import json
+
+
+sample_summary_json = {
+    'id': 'sample_info',
+    'section_name': 'Sample Info',
+    'plot_type': 'table',
+    'file_format': 'json',
+    'data': {}
+}
+
+
+def parse_arguments():
+    """Parse command line arguments."""
+    parser = argparse.ArgumentParser(description=__doc__)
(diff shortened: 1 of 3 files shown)


resolwe_bio/tools/multiqc_sample_info.py
```diff
@@ -0,0 +1,42 @@
...
+    parser.add_argument('--species', nargs='+', help="Species.")
+    parser.add_argument('--build', nargs='+', help="Build.")
+    return parser.parse_args()
+
+
+def main():
+    """Invoke when run directly as a program."""
+    args = parse_arguments()
+
+    for sample in zip(args.samples, args.species, args.build):
```
romunov: This could be `for sample_name, species, build in zip():` which would make the next line redundant.

resolwe_bio/tools/multiqc_sample_info.py
```diff
@@ -0,0 +1,42 @@
...
+        sample_name, species, build = sample
+        if build not in ['rRNA', 'globin']:
+             sample_summary_json['data'][sample_name] = {
+                 'Species': species,
```
romunov: Is this meant to be "human readable" (space in genome build, capital letters)?

romunov:
Ah, I see #862 is already catering to this. :+1:

resolwe_bio/tools/multiqc_sample_info.py
jkokosar: Yes, those are labels in the final rendered table.

PR #867, opened 2020-01-15
Fix Mapping search for source_id / target_id

JureZmrzlikar:
If one did a Mappings query with multiple source_id/target_id as
query parameter the response would be empty or Error. There are
multiple options for one to get multiple Mappings from api. These
are the actual query parameters in request URL:

1. source_id="ID1,ID2"
2. source_id=["ID1", "ID2"]
3. source_id__in="ID1,ID2"
4. source_id__in=["ID1", "ID2"]

The main point is that value of query parameter must be casted to
list as some point. Since source_id is a custom filter the "__in"
snytax is not allowed (3. and 4. are not an option) but there is
no place in code where 1. and 2. would be casted to list. Therefore
the result of 1. or 2 was empty and 3. or 4. raised error.

This solution removes source_id (and targed_id) as custom filters
and places them as normal filtering parameters. Therefore "__in"
syntax is supported and the value of query parameter is splitted by
commma and casted to list. This is also sonsistent with how Feature
filter works for feature_id.

The regression of this solution is is the size of the list that can
be handled by ES in such cases. It reduces to 512 (not exact, but
1024 fails) by unknown reasons(ideas anyone?). However, in practice
this means that before one could filter by max 1 source_id and now
the limit is 512, which seems way better.

## Checklist

    before non-breaking changes.
    commit(s).

### Processes ###

  can be performed).
  and publication.

  tests).

### Tests ###

### DIFF
--- a/resolwe_bio/kb/tests/test_mapping.py
+++ b/resolwe_bio/kb/tests/test_mapping.py
@@ -56,7 +56,7 @@ class MappingTestCase(APITestCase, ElasticSearchTestCase):
         response = self.client.post(MAPPING_URL, {
             'source_db': 'SRC',
             'target_db': 'TGT',
-            'source_id': ['FT0', 'FT1', 'FT5']
+            'source_id__in': 'FT0,FT1,FT5'
         }, format='json')
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         self.assertEqual(len(response.data), 3)
@@ -68,7 +68,7 @@ class MappingTestCase(APITestCase, ElasticSearchTestCase):
         response = self.client.post(MAPPING_URL, {
             'source_db': 'SRC',
             'target_db': 'TGT',
-            'source_id': [str(x) for x in range(2048)]
+            'source_id__in': ','.join([str(x) for x in range(512)])
         }, format='json')
         self.assertEqual(response.status_code, status.HTTP_200_OK)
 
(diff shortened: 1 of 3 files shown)


resolwe_bio/kb/tests/test_mapping.py
```diff
@@ -68,7 +68,7 @@ def test_lookup(self):
         response = self.client.post(MAPPING_URL, {
             'source_db': 'SRC',
             'target_db': 'TGT',
-            'source_id': [str(x) for x in range(2048)]
+            'source_id__in': ','.join([str(x) for x in range(512)])
```
dblenkus: Why have you changed the number?

resolwe_bio/kb/views.py
```diff
@@ -156,35 +156,18 @@ class MappingSearchViewSet(ElasticSearchBaseViewSet):
     document_class = MappingSearchDocument
     serializer_class = MappingSerializer
 
-    filtering_fields = ('source_db', 'source_species', 'target_db', 'target_species', 'relation_type')
+    filtering_fields = (
+        'source_id',
```
dblenkus: Order them alphabetically.

JureZmrzlikar: Done.

resolwe_bio/kb/tests/test_mapping.py
JureZmrzlikar: It crashes with 2048 (or 1024) so i used 512 which seems to work. :-)

PR #876, opened 2020-01-29
Support MultiQC reports in ChIP-seq workflows

jkokosar:
## Checklist

    before non-breaking changes.
    commit(s).

### Processes ###

  can be performed).
  and publication.

  tests).

### Tests ###

### DIFF
--- a/resolwe_bio/processes/chip_seq/macs2_rose2_batch.yml
+++ b/resolwe_bio/processes/chip_seq/macs2_rose2_batch.yml
@@ -17,7 +17,7 @@
           - case
         required: True
   data_name: ChIP-Seq Workflow Batch
-  version: 1.0.3
+  version: 1.0.4
   type: data:chipseq:batch:macs2
   category: ChIP-Seq:Call Peaks
   persistence: CACHED
@@ -57,20 +57,19 @@
       type: list:data:alignment:bam
       description: |
         Select multiple treatment/background samples.
+    - name: promoter
+      label: Promoter regions BED file
+      type: data:bed
+      required: false
+      description: |
+        BED file containing promoter regions (TSS+-1000 bp for example). Needed to get the number
+        of peaks and reads mapped to promoter regions.
     - name: advanced
       label: Show advanced options
       type: basic:boolean
       default: false
       description: |
         Inspect and modify parameters.
-    - name: promoter
-      label: Promoter regions BED file
-      type: data:bed
-      required: false
-      hidden: "!advanced"
-      description: |
-        BED file containing promoter regions (TSS+-1000bp for example). Needed to get the number
-        of peaks and reads mapped to promoter regions.
     - name: tagalign
       label: Use tagAlign files
       type: basic:boolean
@@ -93,11 +92,12 @@
           type: basic:integer
           default: 15000000
         - name: tn5
-          label: TN5 shifting
+          label: Tn5 shifting
           type: basic:boolean
           default: false
           description: |
-            Tn5 transposon shifting. Shift reads on "+" strand by 4bp and reads on "-" strand by 5bp.
+            Tn5 transposon shifting. Shift reads on "+" strand by 4 bp
+            and reads on "-" strand by 5 bp.
         - name: shift
           label: User-defined cross-correlation peak strandshift
           type: basic:integer
@@ -208,7 +208,7 @@
           description: |
             Slocal and llocal parameters control which two levels of regions will be checked
             around the peak regions to calculate the maximum lambda as local lambda. By default,
-            MACS considers 1000bp for small local region (--slocal), and 10000bps for large local
+            MACS considers 1000 bp for small local region (--slocal), and 10000 bp for large local
             region (--llocal) which captures the bias from a long range effect like an open
             chromatin domain. You can tweak these according to your project. Remember that if the
             region is set too small, a sharp spike in the input data may kill the significant
@@ -220,7 +220,7 @@
           description: |
             Slocal and llocal parameters control which two levels of regions will be checked
             around the peak regions to calculate the maximum lambda as local lambda. By default,
-            MACS considers 1000bp for small local region (--slocal), and 10000bps for large local
+            MACS considers 1000 bp for small local region (--slocal), and 10000 bp for large local
             region (--llocal) which captures the bias from a long range effect like an open
             chromatin domain. You can tweak these according to your project. Remember that if the
             region is set too small, a sharp spike in the input data may kill the significant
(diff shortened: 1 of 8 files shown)


resolwe_bio/processes/chip_seq/macs2_rose2_batch.yml
```diff
@@ -17,7 +17,7 @@
           - case
         required: True
   data_name: ChIP-Seq Workflow Batch
-  version: 1.0.3
+  version: 1.1.0
```
romunov: For this file, only the order of option changes and `promoter` hidden status changes. Does this warrant a middle version bump?

resolwe_bio/processes/workflows/chip_seq.yml
```diff
@@ -357,3 +357,341 @@
...
+        - name: q_threshold
+          label: Quality filtering threshold
+          type: basic:integer
+          default: 30
+        - name: n_sub
+          label: Number of reads to subsample
+          type: basic:integer
+          default: 15000000
+        - name: tn5
+          label: TN5 shifting
```
romunov: Make N small perhaps, something along the lines of `Tn5` (also used in description and on wikipedia)?

resolwe_bio/processes/workflows/chip_seq.yml
```diff
@@ -357,3 +357,341 @@
...
+          choices:
+            - label: "1"
+              value: "1"
+            - label: "auto"
+              value: "auto"
+            - label: "all"
+              value: "all"
+          description: |
+            It controls the MACS behavior towards duplicate tags at the exact same location -- the
+            same coordination and the same strand. The 'auto' option makes MACS calculate the
```
romunov: coordination -> coordinate?

resolwe_bio/processes/workflows/chip_seq.yml
```diff
@@ -357,3 +357,341 @@
...
+          required: false
+          default: all
+          choices:
+            - label: "1"
+              value: "1"
+            - label: "auto"
+              value: "auto"
+            - label: "all"
+              value: "all"
+          description: |
```
romunov: Same choices and description as for the `duplicate` argument. Is this correct?

resolwe_bio/processes/workflows/chip_seq.yml
```diff
@@ -357,3 +357,341 @@
...
+            region is set too small, a sharp spike in the input data may kill the significant
+            peak.
+        - name: extsize
+          label: extsize
+          type: basic:integer
+          required: false
+          description: |
+            While '--nomodel' is set, MACS uses this parameter to extend reads in 5'->3' direction
+            to fix-sized fragments. For example, if the size of binding region for your
+            transcription factor is 200 bp, and you want to bypass the model building by MACS,
```
romunov: Sometimes bp is written without a space (i.e. 200bp) and with space (200 bp). Is this something you feel we need to standardize?

resolwe_bio/processes/workflows/chip_seq.yml
```diff
@@ -357,3 +357,341 @@
...
+            this parameter can be set as 200. This option is only valid when --nomodel is set or
+            when MACS fails to build model and --fix-bimodal is on.
+        - name: shift
+          label: Shift
```
romunov: Perhaps a bit more informative label? "Shift in bp"?

resolwe_bio/processes/workflows/chip_seq.yml
```diff
@@ -357,3 +357,341 @@
...
+          default: false
+          description: |
+            With this flag on, MACS will use the background lambda as local lambda. This means
+            MACS will not consider the local bias at peak candidate regions.
+        - name: fix_bimodal
+          label: Turn on the auto paired-peak model process
+          type: basic:boolean
+          default: false
+          description: |
+            Whether turn on the auto paired-peak model process. If it's set, when MACS failed
```
romunov: I think you can remove "Whether" -> "Turn on auto-peak model process.".

resolwe_bio/processes/workflows/chip_seq.yml
```diff
@@ -357,3 +357,341 @@
...
+            to build paired model, it will use the nomodel settings, the '--extsize' parameter
+            to extend each tags. If set, MACS will be terminated if paired-peak model is failed.
```
romunov: "tags" -> "tag"
"is failed" -> "has failed"

resolwe_bio/processes/workflows/chip_seq.yml
```diff
@@ -357,3 +357,341 @@
...
+        - name: nomodel
+          label: Bypass building the shifting model
+          type: basic:boolean
+          hidden: "tagalign"
+          default: false
+          description: |
+            While on, MACS will bypass building the shifting model.
+        - name: nomodel_prepeak
+          label: Bypass building the shifting model
```
romunov: Same label and description as for `nomodel`. Is this right?

resolwe_bio/processes/workflows/chip_seq.yml
```diff
@@ -357,3 +357,341 @@
...
+          default: true
+          hidden: "!tagalign"
+          description: |
+            While on, MACS will bypass building the shifting model.
+        - name: down_sample
+          label: Down-sample
+          type: basic:boolean
+          default: false
+          description: |
+            When set, random sampling method will scale down the bigger sample. By default, MACS
```
romunov: "When set to true...".

resolwe_bio/processes/workflows/chip_seq.yml
```diff
@@ -357,3 +357,341 @@
...
+            uses linear scaling. This option will make the results unstable and irreproducible
+            since each time, random reads would be selected, especially the numbers (pileup,
+            pvalue, qvalue) would change. Consider to use 'randsample' script before MACS2 runs
```
romunov: Is this script `randsample` even applicable in this workflow?

resolwe_bio/processes/workflows/chip_seq.yml
```diff
@@ -357,3 +357,341 @@
...
+          description: |
+            Cutoff for broad region. This option is not available unless --broad is set. If -p is
+            set, this is a p-value cutoff, otherwise, it's a q-value cutoff. DEFAULT = 0.1
+  run:
+    language: workflow
+    program:
+      - id: macs2
+        run: macs2-callpeak
+        input:
+          case: '{{input.case}}'
```
romunov: In some other workflows, we use the syntax `{{ input.case }}` (space between contents and brackets).

resolwe_bio/processes/workflows/chip_seq.yml
jkokosar: Yes, this is intentional. Depending on the choice of `prepeak` input, different UI elements are shown to the user, with different default values.

resolwe_bio/processes/workflows/chip_seq.yml
jkokosar: Using a space is the correct way of writing units. Fixed in the related processes.

resolwe_bio/processes/workflows/chip_seq.yml
jkokosar: Yes, the same situation as above.

resolwe_bio/processes/workflows/chip_seq.yml
jkokosar: I have left this the same as is in use in the Macs2 manual.

PR #880, opened 2020-02-12
Fix imports in differentialexpression-shrna process

JureZmrzlikar:


resolwe_bio/processes/differential_expression/shRNAde.py
```diff
@@ -1,8 +1,17 @@
...
-from resolwe.process import *
+from resolwe.process import (
+    Cmd,
+    DataField,
+    FileField,
+    ListField,
+    Process,
+    SchedulingClass,
+)
+
```
jkokosar: One line to many? Linters will tell.

jkokosar: HA! :)

JureZmrzlikar: Busted! :-)
And fixed...

PR #892, opened 2020-03-10
Handle empty input in BED file correction script

jkokosar:
For some samples, the MACS analysis fails to identify peaks and thus produces an empty .narrowpeaks file. In such cases, Pandas `EmptyDataError` occurs. With this fix, we can detect such events and fail the process with a more user-friendly error message.

### DIFF
--- a/resolwe_bio/tools/bed_file_corrections_genome_browsers.py
+++ b/resolwe_bio/tools/bed_file_corrections_genome_browsers.py
@@ -9,17 +9,27 @@ Correct bed file to be compatible with bedToBigBed tool.
 import argparse
 import pandas as pd
 
+from pandas.errors import EmptyDataError
+from resolwe_runtime_utils import error
+
 parser = argparse.ArgumentParser(description=__doc__)
 parser.add_argument('-f', '--bed_file', help="Bed file.")
 args = parser.parse_args()
 
-df = pd.read_csv(args.bed_file, delimiter='\t', header=None, dtype=str)
-df.iloc[:, 4] = pd.to_numeric(df.iloc[:, 4]).round().astype(int)
-df.iloc[:, 4] = df.iloc[:, 4].clip(upper=1000)
+try:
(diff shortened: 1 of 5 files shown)


resolwe_bio/tools/bed_file_corrections_genome_browsers.py
```diff
@@ -9,17 +9,27 @@
...
 
-df = pd.read_csv(args.bed_file, delimiter='\t', header=None, dtype=str)
-df.iloc[:, 4] = pd.to_numeric(df.iloc[:, 4]).round().astype(int)
-df.iloc[:, 4] = df.iloc[:, 4].clip(upper=1000)
+try:
+    df = pd.read_csv(args.bed_file, delimiter='\t', header=None, dtype=str)
+except EmptyDataError:
+    print(error(
+        f'The input BED file {args.bed_file} is empty. Your analysis might '
+        f'have failed to identify regions of interest (peaks, junctions, etc.)'
```
romunov: Missing full stop after ).

resolwe_bio/tools/bed_file_corrections_genome_browsers.py
```diff
@@ -9,17 +9,27 @@
...
+    ))
+else:
```
romunov: I don't think you need an else clause here. Just checking for an empty file (try/catch) should be enough to trip the script and exit or process if DataFrame contains any data.

jkokosar: It's not necessary to have the `else`, but it might be a better design choice. I did some reading on this topic, and found a good explanation here: https://stackoverflow.com/questions/3996329/else-clause-in-try-statement-what-is-it-good-for/3996378#3996378

romunov: I think these two cases are equivalent and is a matter of personal preference.

PR #897, opened 2020-03-17
Add trimming in wgbs workflows

AnzeLovse:
This PR adds Trimming to WGBS workflow. It is recommended to apply clipping of Illumina adapters and specific tails introduced in some library preparation kits for bisulfite sequencing. Prior tests show an increase from 30% mapped reads to more than 70% mapped reads which makes it a useful addition to the workflow.

## Checklist

    before non-breaking changes.
    commit(s).

### Processes ###

  can be performed).
  and publication.

  tests).

### Tests ###

### DIFF
--- a/resolwe_bio/processes/workflows/wgbs.yml
+++ b/resolwe_bio/processes/workflows/wgbs.yml
@@ -3,15 +3,16 @@
   data_name: "{{ reads|sample_name|default('?') }}"
   requirements:
     expression-engine: jinja
-  version: 1.4.0
+  version: 1.5.0
   type: data:workflow:wgbs
   category: Pipeline
   description: |
-    This WGBS pipeline is comprised of alignment, computation of
+    This WGBS pipeline is comprised of trimming, alignment, computation of
     methylation levels, identification of hypo-methylated regions (HMRs) and
     additional QC steps.
 
-    First, reads are aligned by __WALT__ aligner. [WALT (Wildcard ALignment
+    First, reads are trimmed to remove adapters or kit specific artifacts.
+    Reads are then aligned by __WALT__ aligner. [WALT (Wildcard ALignment
     Tool)](https://github.com/smithlabcode/walt) is fast and accurate read
     mapping for bisulfite sequencing. Then, methylation level at each genomic
     cytosine is calculated using __methcounts__. Finally, hypo-methylated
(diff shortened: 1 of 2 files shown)


resolwe_bio/processes/workflows/wgbs.yml
```diff
@@ -47,6 +48,64 @@
...
+    - name: trimming
+      label: Parameters for the trimming step
+      group:
+        - name: adapters
+          label: Adapter sequences
+          type: data:seq:nucleotide
+          required: false
+          description: |
+            Adapter sequence in FASTA format that will be removed from the read.
+            This field as well as 'Seed mismatches' and 'Simple clip threshold' parameters are needed to perform Illuminacliping.
```
romunov: Is this designation that the parameter is needed for illuminaclipping necessary? If we need this information, would having another group of parameters help?

If yes, consider clipping line width to something like 72 characters.

resolwe_bio/processes/workflows/wgbs.yml
```diff
@@ -364,6 +439,108 @@
...
         otherwise need to be decoded. Default is STRICT.
+    - name: trimming
+      label: Parameters for the trimming step
+      group:
+        - name: adapters
+          label: Adapter sequences
+          type: data:seq:nucleotide
+          required: false
+          description: |
+            Adapter sequence in FASTA format that will be removed from the read.
```
romunov: Same comment about line length as above.

resolwe_bio/processes/workflows/wgbs.yml
```diff
@@ -364,6 +439,108 @@
...
+        - name: simple_clip_threshold
+          label: Simple clip threshold
+          type: basic:integer
+          disabled: >
+            !trimming.adapters
+          required: false
+          description: |
+            Specifies how accurate the match between any adapter etc. sequence must be against a read.
+            This field as well as 'Adapter sequences' and 'Seed mismatches' parameter are needed to perform Illuminacliping.
+        - name: min_adapter_length
```
romunov: I'm just trying to understand the reasoning behind different parameters for trimming for paired and single-end workflows.

resolwe_bio/processes/workflows/wgbs.yml
AnzeLovse: It is necessary to explain that all of those fields are required in order to perform illuminaclipping as those are all set as optional. Separating the illuminaclip parameters is more in line with the original process. Do you think that in this case it is sufficient to explain which parameters are needed only in the description for adapters input?

resolwe_bio/processes/workflows/wgbs.yml
AnzeLovse: Each uses different trimmomatic process, one is intended for single-end reads and the other for paired-end reads and some of the parameters are not used in the first one.

resolwe_bio/processes/workflows/wgbs.yml
```diff
@@ -47,6 +48,75 @@
         performance when processing a BAM file in which
         variable-length data (read, qualities, tags) do not
         otherwise need to be decoded. Default is STRICT.
+    - name: illuminaclip
+      label: Illuminaclip
```
jkokosar: The label should `Adapter trimming`. `Illuminaclip` is a subcommand name of Trimmomatic, and it doesn't say much to the layperson.

AnzeLovse: Fixed. I have also modified the description to match this label.

PR #900, opened 2020-03-26
Decouple generation of genome indices

jkokosar:
## Checklist

    before non-breaking changes.
    commit(s).

### Processes ###

  can be performed).
  and publication.

  tests).

### Tests ###


resolwe_bio/tests/processes/test_junctions.py
```diff
@@ -76,28 +76,28 @@ def test_no_sj_inputs(self):
...
-            inputs = {
+            star_no_junctions = self.run_process('alignment-star', {
                 'genome': star_index_no_junctions.id,
                 'reads': reads_no_junctions.id,
-            }
-            star_no_junctions = self.run_process('alignment-star', inputs)
+            })
 
-            inputs = {
+            empty_bam_upload = self.run_process('upload-bam', {
```
JureZmrzlikar: It would be best to put such formatting in a separate commit, but IMO your time is better spent on other stuff than making this "right". Just for future reference. :-)

resolwe_bio/processes/WGBS/walt_index.py
```diff
@@ -0,0 +1,73 @@
...
+    class Input:
+        """Input fields for WaltIndex."""
+
+        ref_seq = DataField('seq:nucleotide', label='Reference sequence (nucleotide FASTA)')
+
+    class Output:
+        """Output fields to process WaltIndex."""
+
+        index = DirField(label='WALT index')
+        fastagz = FileField(label='FASTA file (compressed)')
```
JureZmrzlikar: Wy do we need fasta and fastagz outputs here? Isn't is ufficient to have them in the FASTA Data object? This duplicates the files and also disk usage...

jkokosar: I agree that this is not efficient in terms of storage. However:
- none of the currently supported aligners need .fasta for the alignment job. However, this might not always be the case. It's easier / less costly to have the input fasta next to the index now, than to handle this in some future migration.
- A typical sample will now have an index object on the input. Consider a scenario, where you want to see the resulting .bam in a genome browser. For this you also need the reference sequence. If the .fasta is not attached to the index object, one would need to fetch a reference sequence through additional parent/child relations.
- Existing workflows often take a current ``data:genome:fasta`` on the input. Often, both the index and the fasta are used in (multiple) workflow steps. By keeping the .fasta next to the index, we can retain only one "genome" input in most of the workflows.
- The number of actual index objects is relatively small - the storage problem is not that great in practice.

resolwe_bio/processes/import_data/nucleotide_seq.py
```diff
@@ -0,0 +1,133 @@
...
+    persistence = Persistence.RAW
+    requirements = {
+        'expression-engine': 'jinja',
+        'executor': {
+            'docker': {
+                'image': 'resolwebio/rnaseq:4.9.0'
+            }
+        },
+        'resources': {
+            'cores': 2,
```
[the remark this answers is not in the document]
jkokosar: I have retained what was in the now retired .yml process. Not sure about the exact reason. We do have a pigz step in the process, and at least that step is significantly faster if multiple cores are available.

jkokosar:
> Have you considered to use more than one core for index building in the processes that currently juts use one core?

Not all of the index building processes benefit from multiple threads. I have retained >1 core only where actually needed.

resolwe_bio/processes/WGBS/walt_index.py
JureZmrzlikar: OK, thanks for the explanation! Agree with you on all points, this makes sense.

docs/CHANGELOG.rst
```diff
@@ -9,10 +9,22 @@ This project adheres to `Semantic Versioning <http://semver.org/>`_.
...
+Added
+-----
+- Add processes: ``bowtie-index``, ``bowtie2-index``, ``bwa-index``,
+  ``hisat2-index``, ``subread-index`` and ``walt-index``.
+
 Changed
 -------
+- **BACKWARD INCOMPATIBLE:** Refactor ``upload-fasta-nucl`` process:
+  ``species`` and ``build`` input information on FASTA file upload  is
+  now mandatory, while ``source`` input is removed.
```
romunov: is -> has been

PR #924, opened 2020-06-02
Use SRA toolkit 2.10.0

jkokosar:
Requires https://github.com/genialis/resolwe-bio/pull/922

### DIFF
--- a/resolwe_bio/tests/processes/test_import.py
+++ b/resolwe_bio/tests/processes/test_import.py
@@ -38,12 +38,12 @@ class ImportProcessorTestCase(BioProcessTestCase):
             "fastqc_url",
             [
                 {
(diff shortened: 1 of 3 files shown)


resolwe_bio/tests/processes/test_import.py
```diff
@@ -37,14 +37,14 @@ def external_test_sra(self):
             import_sra,
             "fastqc_url",
             [
-                {
-                    "file": "fastqc/SRR1661332_fastqc/fastqc_report.html",
-                    "refs": ["fastqc/SRR1661332_fastqc"],
-                },
                 {
```
mzganec, line 40: Elements in `fastqc_url` should be listed in the order that corresponds to the order of SRR references in the input. If this order is not platform invariant, the relevant process needs to be corrected.

PR #925, opened 2020-06-03
Add downsampling + BWA Aln workflows

jkokosar:
Some typos for processes in the same file are also fixed.


resolwe_bio/processes/workflows/chip_seq.yml
```diff
@@ -798,3 +798,237 @@
...
+              label: Seed
+              type: basic:integer
+              default: 11
+            - name: fraction
+              label: Fraction
+              type: basic:decimal
+              required: false
+              range: [0, 1.0]
+              description: |
+                Use the fraction of reads [0 - 1.0] from the orignal input file instead
```
romunov, line 837: orignal -> original

resolwe_bio/processes/workflows/chip_seq.yml
```diff
@@ -798,3 +798,237 @@
...
+                of the absolute number of reads. If set, this will override the
+                "Number of reads" input parameter.
+            - name: two_pass
+              label: 2-pass mode
+              type: basic:boolean
+              default: true
+              description: |
+                Enable two-pass mode when down-sampling. Two-pass mode is twice
+                as slow but with much reduced memory.
```
romunov, line 846: I would go with ".. but much more memory efficient".

resolwe_bio/processes/workflows/chip_seq.yml
```diff
@@ -798,3 +798,237 @@
...
+    - name: alignment
+      label: alignment
+      group:
+        - name: q
+          label: Quality threshold
+          type: basic:integer
+          default: 5
+          range: [0, 60]
+          description: |
+            Parameter for dynamic read trimming.
```
romunov, line 856: Can you elaborate on what dynamic read trimming is?

resolwe_bio/processes/workflows/chip_seq.yml
```diff
@@ -798,3 +798,237 @@
...
+        - name: use_edit
+          label: Use maximum edit distance (excludes fraction of missing alignments)
```
romunov, line 858: I would move the text in the brackets into description and perhaps expand it to include more information on what this option is.

resolwe_bio/processes/workflows/chip_seq.yml
```diff
@@ -798,3 +798,237 @@
...
+                Use the fraction of reads [0 - 1.0] from the orignal input file instead
+                of the absolute number of reads. If set, this will override the
+                "Number of reads" input parameter.
+            - name: two_pass
+              label: 2-pass mode
+              type: basic:boolean
+              default: true
+              description: |
+                Enable two-pass mode when down-sampling. Two-pass mode is twice
+                as slow but with much reduced memory.
```
romunov, line 963: Same as my comment for this parameter in single-end workflow.

resolwe_bio/processes/workflows/chip_seq.yml
```diff
@@ -798,3 +798,237 @@
...
+              label: Seed
+              type: basic:integer
+              default: 11
+            - name: fraction
+              label: Fraction
+              type: basic:decimal
+              required: false
+              range: [0, 1.0]
+              description: |
+                Use the fraction of reads [0 - 1.0] from the orignal input file instead
```
romunov, line 954: orignal -> original

resolwe_bio/tests/processes/test_chipseq.py
```diff
@@ -625,3 +626,61 @@ def set_sample_name(data, sample_name):
         )
         self.assertFileExists(chipqc_macs14, "ccplot")
         self.assertFileExists(chipqc_macs14, "peak_profile")
+
+    @tag_process(
+        "workflow-subsample-bwa-aln-single", "workflow-subsample-bwa-aln-paired"
+    )
+    def test_subsample_bwa_align(self):
+        with self.preparation_stage():
+            input_folder = Path("test_subsample_align") / "input"
```
romunov, line 635: How about defining "base" input folder?

```
base_input = Path("test_subsample_align")
# and then
input_folder = base_input / "input"
output_folder = base_input / "output"
```

resolwe_bio/processes/workflows/chip_seq.yml
jkokosar: I would leave rewording of the descriptions for another time, since I think this is beyond the scope of this PR for several reasons:
1) All of those descriptions are copied from other (existing) processes/workflows and they would need to be updated as well.
2) in many cases, the parameter description is copied directly from the underlying tool documentation, which itself could often be much better.
3) Its often hard to parse the actual meaning of the parameters in use, even if you inspect the tool documentation directly. The "edit distance" above comes from [Levenshtein_distance](https://en.wikipedia.org/wiki/Levenshtein_distance) but it takes time to find this kind of proper descriptions and reword them, which again, is beyond the scope of this PR.

PR #961, opened 2020-09-03
Add upload-clinical-metadata process

AnzeLovse:
This adds a process for the upload of a table that uses orange formatting.

## Checklist

    before non-breaking changes.
    commit(s).

### Processes ###

  can be performed).
  and publication.

  tests).

### Tests ###


AnzeLovse:
@jkokosar please review this. I would also appreciate your feedback @Martin-Spendl .

resolwe_bio/processes/import_data/metadata.py
```diff
@@ -0,0 +1,136 @@
...
+        features = StringField(label="Number of features")
+        target = StringField(label="Target class description")
+        n_metas = IntegerField(label="Number of meta attributes")
+
+    def run(self, inputs, outputs):
+        """Run the analysis."""
+
+        table_path = inputs.src.import_file(imported_format="extracted")
+
+        extensions = [".csv", ".tab", ".tsv", ".xlsx", ".xls"]
```
jkokosar, line 113: A `.txt` would also be a common file suffix and need to be supported.

resolwe_bio/processes/import_data/metadata.py
```diff
@@ -0,0 +1,136 @@
...
+        if Path(table_path).suffix not in extensions:
```
jkokosar, line 114: Maybe use a `lower()` method to handle suffixes written as uppercase.

jkokosar (approved):
The main functionality looks good to me. There are only 2 small improvement recommendations.

Martin-Spendl:
Looks nice to me.
Just a question about how will we handle updating metadata? will the new upload create a new DataObject or rewrite the previous one? First probably, but adding just a few data points to the tsv at the time is quite a common occurrence nowadays.

resolwe_bio/processes/import_data/metadata.py
AnzeLovse: Orange requires lower case extension. Should we allow files with upper case suffixes and then replace the it?

jkokosar: Yes, I think this is not hard to do, but will make an upload process a bit more friendly.

AnzeLovse:
Good question. It will create a new DataObject each time. We could use only the latest metadata object when it is loaded in Orange or delete tables manually. Updating it each time could result in a corrupted table with no rollback (unless it is saved elsewhere).

resolwe_bio/processes/import_data/metadata.py
AnzeLovse: `.txt` is not supported in Orange so I left it out to avoid guessing the delimiter. It would be possible determine delimiter using `csv` module and then renaming the file, but this is not a really robust method.

resolwe_bio/processes/import_data/metadata.py
```diff
@@ -0,0 +1,148 @@
...
+        features = StringField(label="Number of features")
+        target = StringField(label="Target class description")
+        n_metas = IntegerField(label="Number of meta attributes")
+
+    def run(self, inputs, outputs):
+        """Run the analysis."""
+
+        table_path = inputs.src.import_file(imported_format="extracted")
+
+        extensions = [".csv", ".tab", ".tsv", ".xlsx", ".xls"]
```
mstajdohar, line 119: Can (should) you apply this constraint on the FileField?

resolwe_bio/tests/files/test_metadata_upload/inputs/sample_metatable.tsv
```diff
@@ -0,0 +1,11 @@
+function	gene	spo-early	spo-mid	heat 0	heat 10	heat 20
```
mstajdohar, line 1: Use the new Orange format for this use case.

mstajdohar (approved):

resolwe_bio/processes/import_data/metadata.py
AnzeLovse: I have added the supported extensions in the description on the FileField. Is it possible to define accepted file types?

resolwe_bio/tests/files/test_metadata_upload/inputs/sample_metatable.tsv
AnzeLovse: Thanks this is important to test as well.

resolwe_bio/processes/import_data/metadata.py
jkokosar: [validate_regex] (https://github.com/genialis/resolwe/blob/426b12241c69a4ba842e21b66e30347c5ea4161e/resolwe/flow/static/flow/fieldSchema.json#L101 ) is currently not supported in Python processes

resolwe_bio/processes/import_data/metadata.py
```diff
@@ -0,0 +1,153 @@
...
+def format_missing(missing):
+    """Format number of missing values as percentage."""
+    if missing:
+        return f"({missing:.1%} missing values)"
+    else:
+        return "(no missing values)"
+
+
+def get_features_description(data):
+    """Get formated features description."""
```
romunov, line 24: *formatted

resolwe_bio/processes/import_data/metadata.py
```diff
@@ -0,0 +1,153 @@
...
+    return new_path
+
+
+class ImportMetadata(Process):
+    """Upload clinical metadata table in Orange format.
+
+    Orange can read files in native tab-delimited format, or can load
+    data from any of the major standard spreadsheet file types, like CSV
+    and Excel. Native format starts with the names of attributes with
+    prefixes that define atribute type (continuous, discrete, time,
```
romunov, line 69: *attribute

resolwe_bio/tests/files/test_metadata_upload/inputs/malformed.tsv
```diff
@@ -0,0 +1,11 @@
+function	gene	spo-early	spo-mid	heat 0	heat 10	heat 20
```
romunov, line 1: One test file could have suffix in capital letters and the test would catch this info.

resolwe_bio/tests/processes/test_upload.py
```diff
@@ -722,3 +723,76 @@ def test_upload_bedpe(self):
         self.assertFile(bedpe, "bedpe", in_file)
         self.assertFields(bedpe, "species", species)
         self.assertFields(bedpe, "build", build)
+
+    @tag_process("upload-clinical-metadata")
+    def test_upload_metadata(self):
+        base = Path("test_metadata_upload")
```
romunov, line 729: IMO `metadata_upload` would be sufficiently long name of the folder. No need for the `test_`, we already know it's that given the location of this subfolder.

resolwe_bio/processes/import_data/metadata.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+def change_suffix(path):
+    """Change suffix of a file to lowercase."""
+    new_path = path.with_suffix(path.suffix.lower())
+    path.replace(new_path)
+    return new_path
+
+
+class ImportMetadata(Process):
```
romunov, line 63: I feel `Orange` should be somewhere in the name of the class, process and slug. Maybe `ImportOrangeMetadata`?

resolwe_bio/processes/import_data/metadata.py
```diff
@@ -0,0 +1,153 @@
...
+
+    For more information see Orange
+    [documentation](https://orange-visual-programming.readthedocs.io/loading-your-data/index.html#header-with-attribute-type-information).
+
+    An example of native tab-delimited format can be downloaded
+    [here](http://file.biolab.si/datasets/sample-head.xlsx).
+
+    """
+
+    slug = "upload-clinical-metadata"
```
romunov, line 87: `upload-orange-metadata` instead? This process focuses on importing some tabular data into orange, doesn't have to be clinical.

resolwe_bio/processes/import_data/metadata.py
```diff
@@ -0,0 +1,153 @@
...
+    name = "Clinical metadata table"
```
romunov, line 88: Maybe `Metadata table for Orange`?

romunov (approved):
Overall looks good but process name (and class name and slug) could be a bit more explicit about being related to Orange.

resolwe_bio/tests/files/test_metadata_upload/inputs/malformed.tsv
AnzeLovse: Good suggestion, I will add that.

resolwe_bio/processes/import_data/metadata.py
AnzeLovse: I agree that this entirely depends on Orange. The name and description have be updated accordingly.

PR #963, opened 2020-09-10
Use prefetch in SRA import processes

jkokosar:
## Checklist

    before non-breaking changes.
    commit(s).

### Processes ###

  can be performed).
  and publication.

  tests).

### Tests ###

### DIFF
--- a/resolwe_bio/processes/import_data/sra_file.py
+++ b/resolwe_bio/processes/import_data/sra_file.py
@@ -52,7 +52,7 @@ class ImportSra(Process):
     slug = "import-sra"
     name = "SRA data"
     process_type = "data:sra"
-    version = "1.2.0"
+    version = "1.3.0"
     category = "Import"
     scheduling_class = SchedulingClass.BATCH
     persistence = Persistence.TEMP
@@ -76,6 +76,12 @@ class ImportSra(Process):
         class Advanced:
             """Advanced options."""
 
+            prefetch = BooleanField(label="Prefetch SRA file", default=True)
+            max_size_prefetch = StringField(
+                label="Maximum file size to download in KB",
+                default="20G",
+                description="A unit prefix can be used instead of a value in KB (e.g. 1024M or 1G).",
+            )
             min_spot_id = IntegerField(label="Minimum spot ID", required=False)
             max_spot_id = IntegerField(label="Maximum spot ID", required=False)
             min_read_len = IntegerField(label="Minimum read length", required=False)
@@ -110,6 +116,8 @@ class ImportSra(Process):
         process_inputs = {
             "sra_accession": inputs.sra_accession,
             "advanced": {
+                "prefetch": inputs.advanced.prefetch,
+                "max_size_prefetch": inputs.advanced.max_size_prefetch,
                 "clip": inputs.advanced.clip,
                 "aligned": inputs.advanced.aligned,
                 "unaligned": inputs.advanced.unaligned,
@@ -139,7 +147,7 @@ class ImportSraSingle(Process):
     slug = "import-sra-single"
     name = "SRA data (single-end)"
     process_type = "data:reads:fastq:single"
-    version = "1.2.0"
+    version = "1.3.0"
     category = "Import"
     scheduling_class = SchedulingClass.BATCH
     persistence = Persistence.RAW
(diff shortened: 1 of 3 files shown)


resolwe_bio/processes/import_data/sra_file.py
```diff
@@ -361,6 +384,15 @@ def run(self, inputs, outputs):
         fastqc_url_2 = []
         fastqc_dir = Path("fastqc")
         for srr in inputs.sra_accession:
+            if inputs.advanced.prefetch:
+                # prefetch the SRR bundle
+                return_code, _, _ = Cmd["prefetch"][
+                    "-p", inputs.advanced.max_size_prefetch, srr
```
AnzeLovse, line 390: The same as in `import-sra`

resolwe_bio/processes/import_data/sra_file.py
```diff
@@ -194,6 +204,15 @@ def run(self, inputs, outputs):
         """Run the analysis."""
         fastq_gz = []
         for srr in inputs.sra_accession:
+            if inputs.advanced.prefetch:
+                # prefetch the SRR bundle
+                return_code, _, _ = Cmd["prefetch"][
+                    "-p", inputs.advanced.max_size_prefetch, srr
```
AnzeLovse, line 210: Should this option be ` -X` / `--max-size` instead of `-p`?

resolwe_bio/processes/import_data/sra_file.py
```diff
@@ -167,6 +173,10 @@ class Input:
         class Advanced:
             """Advanced options."""
 
+            prefetch = BooleanField(label="Prefetch SRA file", default=True)
+            max_size_prefetch = StringField(
+                label="Maximum file size to download in KB", default="20G"
```
AnzeLovse, line 178: It is not entirely clear that unit prefixes can be used. Perhaps it could be explained in the description.

AnzeLovse (commented):
Please consider my suggestions.

resolwe_bio/processes/import_data/sra_file.py
jkokosar: Fixed, `--max-size` parameter options was missing. `-p` prints out the download progress.

resolwe_bio/processes/import_data/sra_file.py
jkokosar: I have added a description as suggested.

resolwe_bio/processes/import_data/sra_file.py
```diff
@@ -76,6 +76,12 @@ class Input:
         class Advanced:
             """Advanced options."""
 
+            prefetch = BooleanField(label="Prefetch SRA file", default=True)
+            max_size_prefetch = StringField(
+                label="Maximum file size to download in KB",
```
romunov, line 81: Label should indicate that this is in relation to the `Prefetch` option. Ideally, we would hide this option when `prefetch = False`.

resolwe_bio/processes/import_data/sra_file.py
```diff
@@ -194,6 +208,15 @@ def run(self, inputs, outputs):
         """Run the analysis."""
         fastq_gz = []
         for srr in inputs.sra_accession:
+            if inputs.advanced.prefetch:
+                # prefetch the SRR bundle
+                return_code, _, _ = Cmd["prefetch"][
+                    "-p", "--max-size", inputs.advanced.max_size_prefetch, srr
```
romunov, line 214: Maybe list each option-value in a new row, if black permits?

resolwe_bio/tests/processes/test_import.py
```diff
@@ -10,11 +10,10 @@ class ImportProcessorTestCase(BioProcessTestCase):
     @tag_process("import-sra", "import-sra-single", "import-sra-paired")
     def external_test_sra(self):
         # single-end reads from Polyak RNA-seq demo dataset
+        # prefetch needs to be disabled in tests to avoid downloading the whole SRA file bundle
```
romunov, line 13: Would choosing a smaller file help? Perhaps a baby task for someone new to processes to get the feel of the workflow we have for working with processes/tests.

romunov (approved):
Other than what could be debatable wording of some labels, looks good.

resolwe_bio/processes/import_data/sra_file.py
jkokosar: It is auto-formatted by Black. One option per row is not possible.

resolwe_bio/tests/processes/test_import.py
jkokosar: It would, but one would need to find a really small sample (a few MB at most) in SRA for that.

PR #968, opened 2020-09-15
Add format parameter to macs2-callpeak

AnzeLovse:
This adds a format parameter that specifies file type and can be used when macs uses BAM files. 

## Checklist

    before non-breaking changes.
    commit(s).


resolwe_bio/processes/chip_seq/macs2.yml
```diff
@@ -421,6 +440,10 @@
         re-error "--broad can't be combined with --call-summits!"
       {% endif %}
 
+      {% if settings.shift and settings.format == "BAMPE" %}
+        re-error "Values other than 0 are not supported when the format is BAMPE."
```
jkokosar, line 444: State which input parameter values are affected in the error msg,

resolwe_bio/processes/chip_seq/macs2.yml
```diff
@@ -536,6 +559,9 @@
 
         # Single-end reads
         if [ $(samtools view -c -f 1 {{ alignment.bam.file }}) -eq 0 ]; then
+          {% if settings.format == "BAMPE" %}
```
jkokosar, line 562: Couldn't we decide between BAM and BAMPE automatically?

AnzeLovse: This is up for discussion. It is indeed possible, but it would significantly change the process. First, only the reads that are proper pairs would be considered and insert size would be use to build fragment pileup. Based on a few posts this option seems to be a preference rather than a standard choice and by choosing it manually makes the process more universal.

resolwe_bio/processes/chip_seq/macs2.yml
AnzeLovse: Yes, thanks

resolwe_bio/processes/chip_seq/macs2.yml
jkokosar: Ok, than it is best to leave it as is, being user configurable.

jkokosar (approved):

PR #974, opened 2020-09-24
Rewrite differential expression processes to Python

AnzeLovse:
Processes match older versions with the exception of `filter` field in Deseq2 which is renamed to `filter_options`. Before merging old files need to be removed.
## Checklist

    before non-breaking changes.
    commit(s).


AnzeLovse:
@jkokosar could I ask you for a review?

jkokosar (commented):
Nice work! Please remove the old files.

jkokosar (approved):

PR #977, opened 2020-10-02
Add Cut & Run workflow

romunov:
This workflow adds functionality of the Cut & Run workflow. This workflow trims primers using trimgalore, aligns with bowtie2, calls peaks using macs2 and sieves using alignmentSieve. A second, spike-in, organism is used to normalize bigwig files.

Test files come from a snippet of chromosome 17. Top bigwig (blue bars) show all reads, bottom bigwig (blue bars) shows filtered reads where `--maxFragmentLength=120`.

Screenshot from 2020-10-02 11-21-32

In order for the bigger BigWig files to process, an input field has been added to bowtie2 process.

## Checklist
    before non-breaking changes.
    commit(s).


resolwe_bio/processes/workflows/cut_and_run.yml
```diff
@@ -0,0 +1,624 @@
...
+              required: false
+              description: |
+                NextSeq/NovaSeq-specific quality trimming. Trims also dark cycles
+                appearing as high-quality G bases. This will set a specific quality
+                cutoff, but qualities of G bases are ignored. This can not be used
+                with Quality cutoff and will override it.
+            - name: phred
+              label: Phred score encoding
+              type: basic:string
+              required: false
```
jkokosar, line 46: Is there a default value for this input in the tool? If yes, I think setting the default instead of `required: false` is a better approach. Also, the workflow inputs would have those values set explicitly. This applies also to other fields in this input block.

resolwe_bio/processes/workflows/cut_and_run.yml
```diff
@@ -0,0 +1,624 @@
...
+      label: Alignment sieve options
+      group:
+        - name: min_frag_length
+          label: Minimum fragment length
+          type: basic:integer
+          description: |
+            The minimum fragment length needed for read/pair inclusion. This option
+            is primarily useful in ATACseq experiments, for filtering mono- or
+            di-nucleosome fragments. Default is 0.
+          required: false
```
jkokosar, line 483: Could this be set to `default: 0` instead of `required: false`?

resolwe_bio/processes/workflows/cut_and_run.yml
```diff
@@ -0,0 +1,624 @@
...
+        - name: max_frag_length
+          label: Maximum fragment length
+          type: basic:integer
+          description: |
+            The maximum fragment length needed for read/pair inclusion. A value of
+            0 indicates no limit. Default is 0.
+          required: false
```
jkokosar, line 490: See above

jkokosar (approved):
Looks good. If time permits, consider my remarks regarding the input field defaults.

resolwe_bio/processes/workflows/cut_and_run.yml
romunov: Setting the values implicitly seemed like a good idea at the moment. Thinking about it more, setting explicit value makes it more reproducible, even though it may incur some extra  cost when modifying called processes.

resolwe_bio/processes/workflows/cut_and_run.yml
romunov: All inputs have been set to their tool defaults when applicable.

resolwe_bio/processes/workflows/cut_and_run.yml
```diff
@@ -0,0 +1,624 @@
...
+    Cleavage Under Target and Release Using Nuclease. Workflow includes steps of
+    trimming the reads with trimgalore, aligning them using bowtie2 to target species
+    genome as well as a spike-in genome. Aligned reads are processed to produce
+    bigwig files to be viewed in a genome browser. Peaks called using MACS2.
+    Fragmenting of reads is performed using alignmentSieve from
+    deeptools package.
+  input:
+    - name: reads
+      label: Input reads
+      type: data:reads:fastq:paired
```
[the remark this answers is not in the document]
romunov: Some modification would be needed to the Trim galore process. This can be done at a later date.

PR #980, opened 2020-10-20
Add BigWig timeout and bin size parameters

romunov:
Primary reason for this PR is the addition of timeout parameter to `workflow-cutnrun` and corresponding processes. Due to the way the `bamtobigwig.sh` script is written, it makes sense to also add bin size.

    before non-breaking changes.
    commit(s).

### DIFF
--- a/resolwe_bio/processes/workflows/cut_and_run.yml
+++ b/resolwe_bio/processes/workflows/cut_and_run.yml
@@ -3,7 +3,7 @@
   data_name: "{{reads|sample_name|default('?')}}"
   requirements:
     expression-engine: jinja
-  version: 1.1.0
+  version: 1.2.0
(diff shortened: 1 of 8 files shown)


jkokosar (approved):
Fix an issue with linters. Looks good otherwise.

resolwe_bio/processes/workflows/cut_and_run.yml
```diff
@@ -605,6 +605,15 @@
           bigwig_opts:
             bigwig_binsize: "{{input.options_misc.bw_binsize}}"
             bigwig_timeout: "{{input.options_misc.bigwig_timeout}}"
+      - id: species_rmdup_pc
```
jkokosar, line 608: Bump the process version.

PR #981, opened 2020-10-30
Add id field to Feature and Mapping serializers

JureZmrzlikar:
Motivation for this PR: fix the issue with iterate() in ReSDK, that relies on object's id. But until now, filtering by id was not possible for features / mappings.

### DIFF
--- a/resolwe_bio/kb/tests/test_feature.py
+++ b/resolwe_bio/kb/tests/test_feature.py
@@ -192,6 +192,26 @@ class FeatureTestCase(TestCase, APITestCase):
         )
         self.assertEqual(len(response.data), 1)
 
+    def test_serialization(self):
+        FEATURE_SEARCH_URL = reverse("resolwebio-api:kb_feature_search")
+        kwargs = {"feature_id": "FT-0"}
+
+        feature = Feature.objects.get(**kwargs)
+        response = self.client.get(FEATURE_SEARCH_URL, kwargs, format="json")
+        self.assertEqual(len(response.data), 1)
+
+        serialized_feature = response.data[0]
+        self.assertEqual(serialized_feature["aliases"], feature.aliases)
+        self.assertEqual(serialized_feature["description"], feature.description)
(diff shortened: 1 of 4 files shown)


resolwe_bio/kb/tests/test_feature.py
```diff
@@ -192,6 +192,13 @@ def test_feature_search(self):
         )
         self.assertEqual(len(response.data), 1)
 
+    def test_serialization(self):
+        FEATURE_SEARCH_URL = reverse("resolwebio-api:kb_feature_search")
+
+        response = self.client.get(FEATURE_SEARCH_URL, {"limit": 1}, format="json")
+        self.assertEqual(len(response.data["results"]), 1)
+        self.assertTrue("id" in response.data["results"][0])
```
dblenkus, line 200: It would be great to check all fields, not only `id`.

JureZmrzlikar: Good point, thanks. Done.

resolwe_bio/kb/tests/test_feature.py
```diff
@@ -192,6 +193,14 @@ def test_feature_search(self):
         )
         self.assertEqual(len(response.data), 1)
 
+    def test_serialization(self):
+        FEATURE_SEARCH_URL = reverse("resolwebio-api:kb_feature_search")
+
+        response = self.client.get(FEATURE_SEARCH_URL, {"limit": 1}, format="json")
+        self.assertEqual(len(response.data["results"]), 1)
+        for field_name in FeatureSerializer.Meta.fields:
```
dblenkus, line 201: This only tests fundamentals of the serializer - that it includes fields specified in `Meta`. It would be better to check that values are set correctly.

I also prefer the WET approach instead of DRY for testing, so it would be better to type actual field names instead of using the `Meta` class.

JureZmrzlikar: @dblenkus I've modified my approach, please review again.

dblenkus (approved):

PR #1011, opened 2021-03-10
Fix quality recoding in SRA import processes

AnzeLovse:
This fixes recoding of Illumina 1.5 and 1.3 encoding. Encoding is determined by parsing FASTQC output file using `parse_encoding_type.py` tool which prints the encoding to stdout. By default newline is added at the end of the print statement. 

## Checklist

    before non-breaking changes.
    commit(s).

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -28,6 +28,8 @@ Fixed
 -----
 - Fix file import in processes ``upload-multiplexed-single`` and
   ``upload-multiplexed-paired``
+- Fix ``import-sra-paired`` and ``import-sra-paired`` to correctly
+  determine Illumina 1.5 and 1.3 encoding and run recoding.
 
 
 ===================
--- a/resolwe_bio/processes/import_data/sra_file.py
+++ b/resolwe_bio/processes/import_data/sra_file.py
@@ -149,7 +149,7 @@ class ImportSraSingle(Process):
     slug = "import-sra-single"
     name = "SRA data (single-end)"
     process_type = "data:reads:fastq:single"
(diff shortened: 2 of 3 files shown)


docs/CHANGELOG.rst
```diff
@@ -5,6 +5,15 @@ Change Log
 All notable changes to this project are documented in this file.
 This project adheres to `Semantic Versioning <http://semver.org/>`_.
 
+==========
+Unreleased
```
jkokosar, line 9: This section is duplicated. Move to the already existing Unreleased section.

resolwe_bio/processes/import_data/sra_file.py
```diff
@@ -283,16 +283,20 @@ def run(self, inputs, outputs):
             encoding_file = report_dir / "fastqc_data.txt"
             encoding = Cmd["parse_encoding_type.py"](str(encoding_file))
 
-            if encoding == "Illumina 1.5" or encoding == "Illumina 1.3":
+            if encoding == "Illumina 1.5\n" or encoding == "Illumina 1.3\n":
```
jkokosar, line 286: Bump the process version.

resolwe_bio/processes/import_data/sra_file.py
```diff
@@ -467,16 +471,20 @@ def run(self, inputs, outputs):
...
```
jkokosar, line 474: Bump the process version.

jkokosar (commented):
Please check if the same issue exists in the upload-fastq-single/paired process pair.

jkokosar (approved):

PR #1018, opened 2021-04-06
Add dnaio in resolwebio/sra-tools Docker image

AnzeLovse:
## Checklist

    before non-breaking changes.
    commit(s).

### DIFF
--- a/resolwe_bio/docker_images/sra-tools/Dockerfile
+++ b/resolwe_bio/docker_images/sra-tools/Dockerfile
@@ -1,4 +1,4 @@
-FROM docker.io/resolwebio/base:ubuntu-18.04
+FROM public.ecr.aws/genialis/resolwebio/base:ubuntu-20.04-03042021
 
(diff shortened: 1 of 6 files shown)


AnzeLovse:
@jkokosar could you please review this, thanks.

resolwe_bio/docker_images/sra-tools/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM docker.io/resolwebio/base:ubuntu-18.04
+FROM docker.io/resolwebio/base:ubuntu-20.04
```
jkokosar, line 1: Please use pinned version of the image from AWS ECR, similarly as here https://github.com/genialis/resolwe-bio/blob/master/resolwe_bio/docker_images/common/Dockerfile#L1

AnzeLovse: Thanks, I have updated the README as well.

jkokosar (approved):

PR #1034, opened 2021-05-26
Add per-sample whole genome sequencing pipeline (gVCF)

jkokosar:
## Checklist

    before non-breaking changes.
    commit(s).

### Processes ###

  can be performed).
  and publication.

  tests).

### Tests ###


resolwe_bio/processes/workflows/wgs_gvcf.py
```diff
@@ -0,0 +1,375 @@
...
+                "Seed mismatches, Simple clip threshold and "
+                "Palindrome clip threshold parameters are  needed to "
+                "perform adapter trimming. Minimum adaptel length and "
+                "Keep both reads are optional parameters.",
+            )
+
+            seed_mismatches = IntegerField(
+                label="Seed mismatches",
+                required=False,
+                disabled="!trimming.adapters",
```
AnzeLovse, line 97: Please check if fields are correctly disabled. This field seems to have the wrong group name `trimming` instead of `trimming_options`. Same goes for other fields.

resolwe_bio/processes/workflows/wgs_gvcf.py
```diff
@@ -0,0 +1,375 @@
...
+                choices=[
+                    ("null", "Unspecified"),
+                    ("FR", "FR"),
+                    ("RF", "RF"),
+                    ("TANDEM", "TANDEM"),
+                ],
+            )
+
+        class PicardWGSMetrics:
+            """AlignmentSummary parameters."""
```
AnzeLovse, line 183: `AlignmentSummary`-> `PicardWGSMetrics`

resolwe_bio/processes/workflows/wgs_gvcf.py
```diff
@@ -0,0 +1,375 @@
...
+                "completely ignore reads that accumulate beyond this value.",
+            )
+
+            sample_size = IntegerField(
+                label="Sample size used for Theoretical Het Sensitivity sampling",
+                default=10000,
+            )
+
+        class InsertSizeMetrics:
+            """AlignmentSummary parameters."""
```
AnzeLovse, line 223: `AlignmentSummary` -> `InsertSizeMetrics`

AnzeLovse (commented):
Few minor comments otherwise it looks good. And much simpler to run as well.

resolwe_bio/processes/workflows/wgs_gvcf.py
jkokosar: Thanks for pointing this out. Disabled fields were not not supported in Resolwe. Fixed in https://github.com/genialis/resolwe/pull/851

resolwe_bio/processes/variant_calling/wgs_preprocess.py
```diff
@@ -0,0 +1,251 @@
...
+            "--INPUT",
+            "aligned.bam",
+            "--OUTPUT",
+            "marked_duplicates.bam",
+            "--METRICS_FILE",
+            metrics_file,
+            "--VALIDATION_STRINGENCY",
+            "SILENT",
+            "--OPTICAL_DUPLICATE_PIXEL_DISTANCE",
+            "2500",
```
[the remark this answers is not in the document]
jkokosar: Added, thank you!

resolwe_bio/tests/files/wgs/input/TP53.bed
```diff
@@ -0,0 +1 @@
+17	7578063	7578234	TP53_31_59.chr17.7578190.7578536_tile_1
```
romunov, line 1: Some of these testing files are already present in `./bqsr` folder. What was the logic behind duplicating them?

resolwe_bio/tests/processes/test_wgs.py
```diff
@@ -67,3 +68,59 @@ def filter_startedon(line):
...
+        )
+
+        self.assertFile(
+            variants,
+            "vcf",
+            outputs / "variants.g.vcf.gz",
+            file_filter=filter_vcf_variable,
+            compression="gzip",
+        )
+        self.assertFileExists(
```
romunov, line 122: IMO this is tested by the framework as the field is required.

resolwe_bio/tests/processes/test_wgs.py
```diff
@@ -0,0 +1,69 @@
...
+            "wgs-preprocess",
+            {
+                "reads": reads.id,
+                "ref_seq": ref_seq.id,
+                "bwa_index": bwa_index.id,
+                "known_sites": [dbsnp.id],
+            },
+        )
+
+        self.assertFileExists(analysis_ready_bam, "bam")
```
romunov, line 57: IMO this is already tested by the framework since those fields are mandatory in the process.

resolwe_bio/processes/workflows/wgs_gvcf.py
```diff
@@ -0,0 +1,376 @@
...
+            intervals = DataField(
+                "bed",
+                label="Use intervals BED file to limit the analysis to the specified parts of the genome.",
+                required=False,
+            )
+
+            contamination = IntegerField(
+                label="Contamination fraction",
+                default=0,
+                description="Fraction of contamination in sequencing data (for all samples) to aggressively remove.",
```
romunov, line 76: Perhaps break a line to match the width of other lines of code (see for class Trimming).

resolwe_bio/processes/workflows/wgs_gvcf.py
```diff
@@ -0,0 +1,376 @@
...
+
+            adapters = DataField(
+                "seq:nucleotide",
+                label="Adapter sequences",
+                required=False,
+                description="Adapter sequence in FASTA format that will "
+                "be removed from the read. This field as well as "
+                "Seed mismatches, Simple clip threshold and "
+                "Palindrome clip threshold parameters are needed to "
+                "perform adapter trimming. Minimum adapter length and "
```
romunov, line 90: No need to reference other parameters in `adapters`, I think.

resolwe_bio/processes/workflows/wgs_gvcf.py
```diff
@@ -0,0 +1,376 @@
...
+                "below a threshold quality.",
+            )
+
+            minlen = IntegerField(
+                label="Minimum length",
+                required=False,
+                description="Drop the read if it is below a specified length.",
+            )
+
+        class AlignmentSummary:
```
romunov, line 157: Parameters in this class could use a description.

resolwe_bio/processes/variant_calling/genotype_gvcfs.py
```diff
@@ -0,0 +1,175 @@
...
+
+        gvcfs = ListField(
+            DataField("variants:gvcf"),
+            label="Input data (GVCF)",
+        )
+        ref_seq = DataField("seq:nucleotide", label="Reference sequence")
+
+        intervals = DataField(
+            "bed",
+            label="Intervals file (.bed).",
```
romunov, line 50: Labels don't need full stops at the end.

resolwe_bio/processes/variant_calling/genotype_gvcfs.py
```diff
@@ -0,0 +1,175 @@
...
+
+        species = inputs.gvcfs[0].output.species
+        if not all(gvcf.output.species == species for gvcf in inputs.gvcfs):
+            self.error("Not all of the input samples are of the same species.")
+
+        build = inputs.gvcfs[0].output.build
+        if not all(gvcf.output.build == build for gvcf in inputs.gvcfs):
+            self.error("Not all of the input samples have the same genome build.")
+
+        with open("sample_map.txt", "w") as sample_map:
```
romunov, line 118: `sample_map.txt` could be a variable.

resolwe_bio/tests/processes/test_wgs.py
```diff
@@ -124,3 +125,132 @@ def test_gatk_hc_gvcf(self):
...
+                    "build": "custom_build",
+                },
+            )
+
+            intervals = self.run_process(
+                "upload-bed",
+                {
+                    "src": inputs / "hg38.intervals.bed",
+                    "species": "Homo sapiens",
+                    "build": "hg19",
```
romunov, line 149: Maybe have build match others (`custom_build`).

resolwe_bio/tests/processes/test_wgs.py
```diff
@@ -124,3 +125,132 @@ def test_gatk_hc_gvcf(self):
...
+        )
+
+        self.assertFile(
+            joint_variants,
+            "vcf",
+            outputs / "cohort_variants.vcf.gz",
+            file_filter=filter_vcf_variable,
+            compression="gzip",
+        )
+        self.assertFileExists(
```
romunov, line 251: This test is probably not necessary as this is already tested by the framework.

romunov (approved):
Most of my comments are suggestions so I'm giving this a 👍🏼.

PR #1058, opened 2021-08-09
Add qin and ignorebadquality options to BBDuk single/paired process

JureZmrzlikar:

### DIFF
--- a/resolwe_bio/processes/reads_processing/bbduk.yml
+++ b/resolwe_bio/processes/reads_processing/bbduk.yml
@@ -16,7 +16,7 @@
       cores: 10
       memory: 8192
   data_name: '{{ reads|sample_name|default("?") }}'
-  version: 2.4.1
+  version: 2.5.0
   type: data:reads:fastq:single:bbduk
   category: Trim
   flow_collection: sample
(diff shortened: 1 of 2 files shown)


resolwe_bio/processes/reads_processing/bbduk.yml
```diff
@@ -479,6 +496,9 @@
           qtrim={{ operations.quality_trim }} \
         {% endif %} \
         trimq={{ operations.trim_quality }} \
+        {% if operations.ignore_bad_quality %} \
```
jkokosar, line 499: `qin={{ operations.quality_encoding_offset }} \` is missing for the ``bbduk-single`` process.

resolwe_bio/processes/reads_processing/bbduk.yml
```diff
@@ -16,7 +16,7 @@
       cores: 10
       memory: 8192
   data_name: '{{ reads|sample_name|default("?") }}'
-  version: 2.4.1
+  version: 2.5.1
```
jkokosar, line 19: Should be ``2.5.0``

resolwe_bio/processes/reads_processing/bbduk.yml
```diff
@@ -552,7 +572,7 @@
...
```
jkokosar, line 575: Should be ``2.5.0``

resolwe_bio/processes/reads_processing/bbduk.yml
JureZmrzlikar: Nice catch, thanks!

jkokosar (approved):

PR #1062, opened 2021-08-25
Add aligned reads (BAM) input option to wgs-preprocess tool

jkokosar:
## Checklist

    before non-breaking changes.
    commit(s).

### Processes ###

  can be performed).
  and publication.

  tests).

### Tests ###


resolwe_bio/processes/variant_calling/wgs_preprocess.py
```diff
@@ -90,49 +95,98 @@ class Output:
...
-            f"-t {self.requirements.resources.cores}",
-            "-Y",
-            f"{Path(inputs.bwa_index.output.index.path) / index_fasta_name}",
-            "input_reads_mate1.fastq.gz",
-            "input_reads_mate2.fastq.gz",
-        ]
-        (Cmd["bwa"]["mem"][bwa_inputs] > aligned_sam)()
-        self.progress(0.2)
+        if not inputs.reads and not inputs.aligned_reads:
+            self.error("Please provide the input data.")
```
romunov, line 108: Specify which input. Perhaps something like `Please provide FASTQ or BAM input files`.

resolwe_bio/processes/variant_calling/wgs_preprocess.py
```diff
@@ -90,49 +95,98 @@ class Output:
...
+                "input_reads_mate1.fastq.gz",
+                "input_reads_mate2.fastq.gz",
+            ]
+            (Cmd["bwa"]["mem"][bwa_inputs] > aligned_sam)()
+            self.progress(0.2)
+
+        else:
+            # Define output file names
+            name = inputs.aligned_reads.entity_name
+            if name == None or name == "":
```
romunov, line 150: Perhaps you could use `if not name:` instead?

```In [1]: name = ""                                                               

In [2]: if name: 
   ...:     print("hello") 
   ...:                                                                         

In [3]: name = None                                                             

In [4]: if name: 
   ...:     print("hello") 
   ...:                                                                         

In [5]: if not name: 
   ...:     print("hello") 
   ...:                                                                         
hello
```

romunov (approved):

resolwe_bio/processes/variant_calling/wgs_preprocess.py
jkokosar: Done, thanks for the suggestion!

resolwe_bio/processes/variant_calling/wgs_preprocess.py
```diff
@@ -90,49 +95,98 @@ class Output:
...
+            name = inputs.aligned_reads.entity_name
+            if not name:
+                bam_path = Path(inputs.aligned_reads.output.bam.path).name
+                assert bam_path.endswith(".bam")
+                name = bam_path[:-4]
+
+            collate_inputs = [
+                f"-@ {self.requirements.resources.cores}",
+                "-O",
+                inputs.aligned_reads.output.bam.path,
```
AnzeLovse, line 158: It would be nice to check if species and build are the same for aligned reads and index file.

AnzeLovse (approved):
Looks good to me.

PR #1068, opened 2021-09-02
Rewrite featureCounts process to Python

jkokosar:
This PR rewrites the featureCounts process from YAML to a Python process:
* the change is backwards incompatible due to the change to the input field structure
* removed is an option to produce RPKUM normalized counts, as the feature was never used
* feature mapping from IDs to symbols now utilises Resolwe's feature to access the KB directly


JureZmrzlikar (commented):
I first have a few more high level questions:
- Why did you decide for breaking-change of the inputs? Have you really considere the possible problems this might have?
- Did you do any validation on real data: e.g. comparing the results of yaml vs. Python processes? I think such comparison would make us feel way more comfortable accepting this.

Otherwise, this is a heroic effort. IMO, this process is and will be one of the most frequently used on the platform and as such deserves to be Pythonic all the way. Great work!

resolwe_bio/processes/expression/featureCounts.py
```diff
@@ -0,0 +1,888 @@
...
+
+def rename_columns_and_compress(infile, outfile):
+    """Rename columns and compress the expression files."""
+    exp = pd.read_csv(
+        infile,
+        sep="\t",
+        usecols=["FEATURE_ID", "EXPRESSION"],
+        index_col="FEATURE_ID",
+        squeeze=True,
+        float_precision="round_trip",
```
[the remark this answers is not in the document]
jkokosar: The default precision setting for reading the file containing the float values with ``read_csv`` will sacrifice the precision for speed, rounding up the float values in the process. To crate expression_set files, we are reading the expression files and merging them into a single data frame. To ensure that the values in in the merged data frame are the same as in the input files, the [float_precission="round_trip"](https://pandas.pydata.org/pandas-docs/stable/user_guide/io.html#specifying-method-for-floating-point-conversion) needs to be used.

resolwe_bio/processes/expression/featureCounts.py
```diff
@@ -0,0 +1,888 @@
...
+        # Cast FEATURE_ID column to string
+        df["FEATURE_ID"] = df["FEATURE_ID"].astype("str")
+
+    return df
+
+
+def prepare_expression_set(rc, tpm, cpm, feature_dict, outfile_name):
+    """Prepare expression set output data."""
+    rc_exp = exp_to_df(rc, "RAW_COUNT")
+    tmp_exp = exp_to_df(tpm, "TPM")
```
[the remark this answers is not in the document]
jkokosar: Fixed, thank you!

gregorjerse (approved):

resolwe_bio/processes/expression/featureCounts.py
```diff
@@ -0,0 +1,888 @@
...
+        return_code, _, _ = Cmd["featureCounts"][args][bam_file] & TEE(retcode=None)
+        if return_code:
+            self.error("Error while running featureCounts.")
+
+        # parse featureCounts output
+        get_gene_counts("featureCounts_rc.txt", "rc.txt", bam_file)
+        get_gene_lenghts("featureCounts_rc.txt", "gene_lengths.txt")
+
+        # Normalize counts
+        rnanorm_args = [
```
romunov, line 821: If you decide to go for "one parameter per row" as per my comment above, this would also be the place to change that.

resolwe_bio/processes/expression/featureCounts.py
```diff
@@ -0,0 +1,888 @@
...
+            "--cpm-output",
+            "cpm.txt",
+        ]
+        return_code, _, _ = Cmd["rnanorm"][rnanorm_args] & TEE(retcode=None)
+        if return_code:
+            self.error("Error while normalizing counts using rnanorm.")
+
+        # prepare the expression set outputs
+        feature_ids = pd.read_csv(
+            "rc.txt", sep="\t", index_col="FEATURE_ID"
```
romunov, line 836: Make `"rc.txt"` into a variable and use it through the script.

resolwe_bio/processes/expression/featureCounts.py
```diff
@@ -0,0 +1,888 @@
...
+            "species": inputs.aligned_reads.output.species,
+            "feature_id__in": feature_ids,
+        }
+
+        feature_ids_to_names = {
+            f.feature_id: f.name for f in self.feature.filter(**feature_filters)
+        }
+
+        prepare_expression_set(
+            rc="rc.txt",
```
romunov, line 850: See how nicely readable explicitly named arguments are? :)

resolwe_bio/processes/expression/featureCounts.py
```diff
@@ -0,0 +1,888 @@
...
+        get_gene_counts("featureCounts_rc.txt", "rc.txt", bam_file)
+        get_gene_lenghts("featureCounts_rc.txt", "gene_lengths.txt")
+
+        # Normalize counts
+        rnanorm_args = [
+            "rc.txt",
+            "--gene-lengths",
+            "gene_lengths.txt",
+            "--tpm-output",
+            "tpm.txt",
```
romunov, line 826: This could go into a variable and used throughout the script.

resolwe_bio/processes/expression/featureCounts.py
```diff
@@ -0,0 +1,888 @@
...
+            "--cpm-output",
+            "cpm.txt",
```
romunov, line 828: This could go into a variable and be used throughout the script. IMO this would make it more resilient for future changes.

resolwe_bio/processes/expression/featureCounts.py
```diff
@@ -0,0 +1,888 @@
...
+        )
+
+        # rename and compress the expression files
+        rename_columns_and_compress("rc.txt", f"{name}_rc.tab.gz")
+        rename_columns_and_compress("tpm.txt", f"{name}_tpm.tab.gz")
+        rename_columns_and_compress("cpm.txt", f"{name}_cpm.tab.gz")
+
+        # Compress the featureCounts output file
+        (
+            Cmd["gzip"]["-c", "featureCounts_rc.txt"]
```
romunov, line 864: Technically this could be rewritten into a more pythonic way, without resorting to `Cmd`.

```python
import gzip
import shutil

with open(filename="featureCounts_rc.txt", mode = "rb") as f_in:
    with gzip.open(filename=f"{name}_featureCounts_rc.txt.gz", mode="wb") as f_out:
        shutil.copyfileobj(f_in, f_out)
```

resolwe_bio/processes/expression/featureCounts.py
```diff
@@ -0,0 +1,888 @@
...
+            > f"{name}_featureCounts_rc.txt.gz"
+        )()
+
+        # Save the abundance estimates to JSON storage
+        Cmd["expression2storage.py"]("--output", "json.txt", exp_output)
```
romunov, line 869: Move this script into a function.  Eventually this will have to become a beast on its own that will have to reside within `/resolwe_bio/tools/` 

Perhaps something along the lines of

```python
import os
import utils
import json


def expression_to_storage(rc_input, rc_output):
    
    if not (rc_input and os.path.isfile(rc_input)):
        raise ValueError("Specify rc_input")

    def isfloat(value):
        """Check if value is float."""
        try:
            float(value)
            return True
        except ValueError:
            return False

    with utils.gzopen(rc_input) as f:
        # Split lines by tabs
        # Ignore lines without a number in second column
        # Build a dictionary of gene-expression pairs
        exp = {
            "genes": {
                gene_exp[0]: float(gene_exp[1])
                for gene_exp in (l.split("\t") for l in f)
                if len(gene_exp) == 2 and isfloat(gene_exp[1])
            }
        }

    if rc_output:
        with open(file=rc_output, mode="wt") as f:
            json.dump(exp, f)
    
    return rc_output
```

resolwe_bio/processes/workflows/mirna.yml
```diff
@@ -305,19 +305,16 @@
...
+          id_attribute: '{{ input.quant_options.id_attribute }}'
+          feature_class: '{{ input.quant_options.feature_class }}'
           normalization_type: '{{ input.quant_options.normalization_type }}'
-          advanced:
-            overlap:
-              allow_multi_overlap: '{{ input.quant_options.allow_multi_overlap}}'
-            multi_mapping_reads:
-              count_multi_mapping_reads: '{{ input.quant_options.count_multi_mapping_reads }}'
+          overlap:
+            allow_multi_overlap: '{{ input.quant_options.allow_multi_overlap}}'
```
romunov, line 315: Add space between `overlap` and `}}`.

romunov (changes requested):
Can you explain how the test files changed? I would expect that results to be identical to the yaml process version.

resolwe_bio/processes/expression/featureCounts.py
```diff
@@ -0,0 +1,888 @@
...
+        # Avoid reporting the full path to the alignment (.bam) file in the counts summary file.
+        # This is to prevent the FeatureCounts results to be reported as a separate sample in the MultiQC report
+        bam_file = Path(inputs.aligned_reads.output.bam.path).name
+        Path(bam_file).symlink_to(inputs.aligned_reads.output.bam.path)
+
+        # Set output file names
+        assert bam_file.endswith(".bam")
+        name = bam_file[:-4]
+
+        if inputs.normalization_type == "TPM":
```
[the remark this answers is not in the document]
jkokosar: Done, thanks for the suggestion.

resolwe_bio/processes/expression/featureCounts.py
```diff
@@ -0,0 +1,888 @@
...
+            # Run Salmon Quant
+            salmon_args = [
+                "-i",
+                inputs.cdna_index.output.index.path,
+                "-l",
+                "A",
+                reads_input if paired_end else ["-r", "reads.fastq"],
+                "-o",
+                "salmon_output",
+                "-p",
```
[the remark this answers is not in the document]
jkokosar: I agree that your suggestion improves the readability, however it is often not possible to use such a syntax due to the problems with Plumbum parameter passing to the installed tools (system calls). In the case of a call to Salmon tool as suggested above, Salmon would report that  `"-p 4"` is not a recognised argument.

resolwe_bio/processes/expression/featureCounts.py
```diff
@@ -0,0 +1,888 @@
...
+            args.append("-f")
+
+        if inputs.overlap.allow_multi_overlap:
+            args.append("-O")
+
+        if inputs.overlap.largest_overlap:
+            args.append("--largestOverlap")
+
+        if inputs.overlap.read_to_pos:
+            args.append("--read2pos")
```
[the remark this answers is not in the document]
jkokosar: Thanks for catching that. Fixed.

resolwe_bio/processes/expression/featureCounts.py
```diff
@@ -0,0 +1,888 @@
...
+            args.extend(["-d", inputs.paired_end.min_frag_length])
+
+        if paired_end and inputs.paired_end.check_frag_length:
+            args.extend(["-D", inputs.paired_end.max_frag_length])
+
+        if paired_end and inputs.paired_end.do_not_count_chimeric_fragments:
+            args.append("-C")
+
+        if paired_end and inputs.paired_end.do_not_sort:
+            args.append("--do_not_sort")
```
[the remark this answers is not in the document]
jkokosar: Good catch!

resolwe_bio/processes/expression/featureCounts.py
```diff
@@ -0,0 +1,888 @@
...
+
+        # Run featureCounts
+        return_code, _, _ = Cmd["featureCounts"][args][bam_file] & TEE(retcode=None)
+        if return_code:
+            self.error("Error while running featureCounts.")
+
+        # parse featureCounts output
+        get_gene_counts("featureCounts_rc.txt", "rc.txt", bam_file)
```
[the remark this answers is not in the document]
jkokosar: Done!

resolwe_bio/processes/expression/featureCounts.py
jkokosar: Done

resolwe_bio/processes/expression/featureCounts.py
jkokosar: Done, with slight simplifications.

resolwe_bio/processes/expression/featureCounts.py
jkokosar: > Can you explain how the test files changed? I would expect that results to be identical to the yaml process version.

In the YAML process, the float values were rounded at ~10th decimal place when importing tab delimited data to Pandas when creating "expression set" files. As explained in my comment to @JureZmrzlikar, we now use  `float_precission="round_trip"` so that the values don't change when reading and writing exiting float values using Pandas.

jkokosar:
> Can you explain how the test files changed? I would expect that results to be identical to the yaml process version.

In the YAML process, the float values were rounded at ~10th decimal place when importing tab delimited data to Pandas when creating "expression set" files. As explained in my comment to @JureZmrzlikar, we now use float_precission="round_trip" so that the values don't change when reading and writing exiting float values using Pandas.

romunov (approved):

PR #1070, opened 2021-09-08
Added process that uploads bwamem2 index files

OtonicarJan:
Added the process that uploads bwa-mem2 index.

## Checklist

* [x ] Update CHANGELOG.rst for each commit separately:
    before non-breaking changes.
    commit(s).


resolwe_bio/processes/import_data/bwamem2_index_files.py
```diff
@@ -0,0 +1,75 @@
...
+            "docker": {"image": "public.ecr.aws/genialis/resolwebio/dnaseq:6.1.0"},
+        },
+    }
+    data_name = '{{ ref_seq.fasta.file|basename|default("?") }}'
+    version = "1.0.0"
+
+    class Input:
+        """Input fields to process Import BWA-MEM2 index."""
+
+        ref_seq = DataField(
```
jkokosar, line 38: It's best if also the FASTA on input `ref_seq` is to be uploaded (`FileField`) in this process. With this, you'll also need to modify the `data_name` to be `'{{ ref_seq.file|default("?") }}'`

resolwe_bio/processes/import_data/bwamem2_index_files.py
```diff
@@ -0,0 +1,75 @@
...
+        name = basename[:-6]
+
+        index_dir = Path("BWAMEM2_index")
+        if not index_dir.exists():
+            index_dir.mkdir()
+
+        index_file = inputs.index_name.import_file(imported_format="compressed")
+        shutil.unpack_archive(index_file, index_dir)
+
+        shutil.copy(Path(inputs.ref_seq.output.fasta.path), Path.cwd())
```
jkokosar, line 66: When this is not the existing Data object, you need to handle the import differently. For FASTA import procedure, you can look at the existing code in the process `upload-fasta-nucl`

jkokosar (dismissed):

jkokosar (approved):

PR #1072, opened 2021-09-16
Update FASTQ file merging to use sample relations

AnzeLovse:
## Checklist

    before non-breaking changes.
    commit(s).

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -7,6 +7,17 @@ All notable changes to this project are documented in this file.
 This project adheres to `Semantic Versioning <http://semver.org/>`_.
 
 
+==========
+Unreleased
+==========
+
+Changed
+-------
+- **BACKWARD INCOMPATIBLE:** Update ``merge-fastq-single`` and
+  ``merge-fastq-paired`` processes to use sample relations for merging
+  FASTQ files
+
+
 ===================
 38.4.0 - 2021-09-14
 ===================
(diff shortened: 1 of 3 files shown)


AnzeLovse:
@jkokosar could you please review this PR?

resolwe_bio/processes/support_processors/merge_samples.py
```diff
@@ -1,218 +1,190 @@
...
+
+            self.run_process(
+                slug="upload-fastq-single",
+                inputs={
+                    "src": symlinks,
+                    "merge_lanes": True,
+                },
+            )
+
+            mereged_objects = Data.filter(entity__name=symlinks[0])
```
jkokosar, line 111: `mereged` -> `merged` here and below.

resolwe_bio/processes/support_processors/merge_samples.py
```diff
@@ -1,218 +1,190 @@
...
-
-        outputs.fastqc_url = fastqc_url
-        outputs.fastqc_archive = [fastqc]
+
+        # Check if user has selected multiple read objects from the same sample
+        data_by_sample = {}
+        for data in inputs.reads:
+            if data.entity_id in data_by_sample:
+                self.warning(
+                    "There are multiple read objects for "
```
jkokosar, line 85: It is common for the sample to have trimmed reads objects attached. For any subsequent joining of already analysed data, the original input (FASTQ) that originated the entity might make more sense. So, I propose to merge the first FASTQ on the sample, not the last.

docs/CHANGELOG.rst
```diff
@@ -22,6 +22,9 @@ Added
 
 Changed
 -------
+- **BACKWARD INCOMPATIBLE:** Update ``merge-fastq-single`` and
```
jkokosar, line 25: This should go to the unreleased section

resolwe_bio/processes/support_processors/merge_samples.py
AnzeLovse: Thanks that is a good suggestion.

PR #1073, opened 2021-09-17
Add GATK variant refinement process

OtonicarJan:
## Checklist

    before non-breaking changes.
    commit(s).


OtonicarJan:
@jkokosar please review this PR.

resolwe_bio/processes/variant_calling/gatk_variant_refinement.py
```diff
@@ -0,0 +1,126 @@
...
+            "cores": 2,
+            "memory": 16384,
+        },
+    }
+    data_name = '{{ vcf|sample_name|default("?") }}'
+
+    class Input:
+        """Input fields for GatkGenotypeRefinement."""
+
+        vcf_main = DataField(
```
jkokosar, line 43: Use just `vcf` for this input, so that the variant file input name in this process is the same as in the existing  VQSR process.

resolwe_bio/processes/variant_calling/gatk_variant_refinement.py
```diff
@@ -0,0 +1,126 @@
...
+        )
+        ref_seq = DataField("seq:nucleotide", label="Reference sequence")
+
+        vcf_pop = DataField(
+            "variants:vcf",
+            label="Population-level variant set VCF data",
+            required=False,
+        )
+
+        ped = FileField(label="PED file describing pedigree", required=False)
```
jkokosar, line 54: This design where you are mixing `DataField` inputs and `FileField` inputs in the same process is not optimal. Every time one would want to trigger this process, there will be a need to upload the input for the `ped` field, making unnecessary copies of this input in the database. It is better to have a separate "upload" process for `ped` files, and to use `DataField` in this process. In this way, one would upload only once, and use it as many times as needed.

resolwe_bio/processes/variant_calling/gatk_variant_refinement.py
```diff
@@ -0,0 +1,126 @@
...
+            "-O",
+            variants_cgp,
+            "--tmp-dir",
+            TMPDIR,
+        ]
+
+        if inputs.vcf_pop:
+            args_cgp.extend(["--supporting", inputs.vcf_pop.output.vcf.path])
+        elif inputs.ped:
+            args_cgp.extend(["-ped", inputs.ped.import_file])
```
jkokosar, line 88: Reorganize code to suport the ped input as an existing data object.

resolwe_bio/processes/variant_calling/gatk_variant_refinement.py
```diff
@@ -0,0 +1,126 @@
...
+
+        return_code, _, _ = Cmd["gatk"]["CalculateGenotypePosteriors"][args_cgp] & TEE(
+            retcode=None
+        )
+        if return_code:
+            self.error("GATK CalculateGenotypePosteriors failed.")
+
+        Cmd["tabix"]["-p", "vcf", "-f", variants_cgp]()
```
jkokosar, line 96: Check if the `CalculateGenotypePosteriors` already has build-in support to generate the index. If it does, the call to `tabix` tool would be redundant.

resolwe_bio/processes/variant_calling/gatk_variant_refinement.py
```diff
@@ -0,0 +1,126 @@
...
+            "GQ < 20",
+        ]
+
+        return_code, _, _ = Cmd["gatk"]["VariantFiltration"][args_filtration] & TEE(
+            retcode=None
+        )
+        if return_code:
+            self.error("GATK VariantFiltration failed.")
+
+        Path(variants_cgp).unlink(missing_ok=True)
```
jkokosar, line 117: The intermediate files that are not explicitly referenced to the matching output fields will be automatically deleted, so it is not needed to delete them manually. The manual deletion only makes sense if multiple additional steps of processing of this VCF would follow, and one would try to save the storage space needed for completion of this analysis.

resolwe_bio/processes/variant_calling/gatk_variant_refinement.py
```diff
@@ -0,0 +1,126 @@
...
+
+    class Output:
+        """Output fields for GatkGenotypeRefinement."""
+
+        vcf = FileField(label="Refined multi-sample vcf")
+        tbi = FileField(label="Tabix index")
+        species = StringField(label="Species")
+        build = StringField(label="Build")
+
+    def run(self, inputs, outputs):
```
jkokosar, line 64: Add an additional check to the body of the process which validates that the `species` and `build` info for the `ref_seq` and `vcf` inputs are the same. If they are not, fail the process with a warning. If the `vcf_pop` is given, check the same also for this input.

jkokosar (approved):

PR #1077, opened 2021-10-01
Update pull request template

AnzeLovse:
## Checklist

    before non-breaking changes.
    commit(s).

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -20,7 +20,7 @@ Mark an `[x]` for completed items.
   * **PATCH version (last number)**: changes/bug fixes that do not affect
     the api/interface. Examples: typo fix, change/add warning messages...
 * [ ] All inputs are used in process.
-* [ ] All output fields have their ``re-save`` calls.
+* [ ] All output fields have a value assigned to them.
 
 <!--
 Read the following guidelines and remove them before opening the pull request.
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -35,39 +35,40 @@ Read the following guidelines and remove them before opening the pull request.
...
   can be performed).
-* Description of tool should be self explanatory with links to tool's homepage
-  and publication.
+* Description of the tool should be self explanatory with links to the tool's
+  homepage and publication.
 * Write explanatory input labels and add short command line parameter names
   in square brackets (for example: Maximum length [``--maxlength``]).
 * Where possible provide default values for the process input fields.
-  If they are provided, then ``required: false`` is not needed.
+  If they are provided, then ``required = false`` is not needed.
```
JureZmrzlikar, line 43: Capitalize `false`? --> `False`

.github/pull_request_template.md
```diff
@@ -35,39 +35,40 @@ Read the following guidelines and remove them before opening the pull request.
...
+  must be manually tagged and pushed to Amazon ECR.
 * Mark incomplete PRs with the [WIP] (Work In Progress) tag.
 * Do not paste links to private repositories from the public one.
 
 ### Tests ###
 
-* Add meaningful names of test files with spaces:
-  * Genomes: ``genome species.fa.gz``
-  * Annotations: ``annotation species.gtf.gz``
+* Add meaningful names of test files with underscores:
```
JureZmrzlikar, line 61: Maybe this comment is not so relevant. In yaml processes, we had to be very careful with spaces in filenames, but in Python processes, this is not an issue. Maybe just "Add meaningful names of test files" would suffice.

JureZmrzlikar (approved):
Good work, thanks for the effort @AnzeLovse

.github/pull_request_template.md
AnzeLovse: That is a good point, I will change it to "Add meaningful names of test files". But I would keep the underscores in example file names as it is still a better practice to use underscores instead of spaces.

AnzeLovse:
Ping @romunov, @OtonicarJan, @jkokosar

jkokosar (approved):

.github/pull_request_template.md
```diff
@@ -20,7 +20,7 @@ Mark an `[x]` for completed items.
   * **PATCH version (last number)**: changes/bug fixes that do not affect
     the api/interface. Examples: typo fix, change/add warning messages...
 * [ ] All inputs are used in process.
-* [ ] All output fields have their ``re-save`` calls.
+* [ ] All output fields have a value assigned to them.
```
romunov, line 23: This is not 100 % exact because some output fields do not have mandatory "outputs". Perhaps something along the lines of 

> All required output fields have a value assigned to them.

On the other hand, this is something we would want to check with a test anyway, right?

romunov (approved):
This is great work, thank you for bringing the solution from start to the finish line.

PR #1078, opened 2021-10-04
Add VEP cache upload and ensembl-VEP processes

OtonicarJan:
## Checklist

    before non-breaking changes.
    commit(s).
* [x ] Bump the process version:
* [x ] All inputs are used in process.


OtonicarJan:
@jkokosar Could you please review this PR?

resolwe_bio/processes/variant_calling/ensembl_vep.py
```diff
@@ -0,0 +1,99 @@
...
+                "Species information for the input VCF file and cache don't match."
+            )
+        elif inputs.vcf.output.build != inputs.cache.output.build:
+            self.error(
+                "Build information for the input VCF file and cache don't match."
+            )
+
+        if inputs.cache.output.release != "104":
+            self.warning(
+                "The current version of ensembl-vep is 104. It is recommended that cache version is also 104"
```
jkokosar, line 72: This will go out of sync as soon as we update the VEP tool version in Docker image. If the tool can return its version (e.g. --version, or -v, or ...), you should store this as a variable and use it instead of hardcoded value.

resolwe_bio/processes/variant_calling/ensembl_vep.py
```diff
@@ -0,0 +1,99 @@
...
+
+        args = [
+            "--cache",
+            "--dir",
+            inputs.cache.output.cache.path,
+            "-v",
+            "--format",
+            "vcf",
+            "--everything",
+            "--force_overwrite",
```
jkokosar, line 83: Will this parameter ensure that if the file with the same name already exists, it will be overwritten?  There will be no multiple runs of the VEP tool within one Data objects, so the parameter is most probably redundant.

resolwe_bio/processes/variant_calling/ensembl_vep.py
```diff
@@ -0,0 +1,99 @@
...
+        return_code, _, _ = Cmd["vep"][args] & TEE(retcode=None)
+        if return_code:
+            self.error("Ensembl-vep failed.")
+
+        (Cmd["bgzip"]["-c", annotated_vcf] > annotated_vcf_gz)()
+        Cmd["tabix"]["-p", "vcf", annotated_vcf_gz]()
+
+        outputs.vcf = annotated_vcf_gz
+        outputs.tbi = annotated_vcf_index
+        outputs.summary = summary_html
```
jkokosar, line 99: Store also the species and build info.

jkokosar (approved):
Looks good! Nice work!

PR #1081, opened 2021-10-07
Add GATK VariantsToTable process

OtonicarJan:
## Checklist

    before non-breaking changes.
    commit(s).
* [x ] All output fields have a value assigned to them.


resolwe_bio/processes/variant_calling/variants_to_table.py
```diff
@@ -0,0 +1,123 @@
...
+    This tool extracts specified fields for each variant in a VCF file
+    to a tab-delimited table, which may be easier to work with than a VCF.
+    For additional information, please see
+    https://gatk.broadinstitute.org/hc/en-us/articles/360036711531-VariantsToTable
+    """
+
+    slug = "variants-to-table"
+    name = "GATK VariantsToTable"
+    category = "GATK"
+    process_type = "data:variants:vcf:variantstotable"
```
AnzeLovse, line 30: The output of the process does not include variants in the vcf file. It would be more appropriate to use a new data type. Something like `data:variantstable` or `variantstotable`.

resolwe_bio/processes/variant_calling/variants_to_table.py
```diff
@@ -0,0 +1,123 @@
...
+    data_name = "Variants in table"
+
+    class Input:
+        """Input fields for GATK VariantsToTable."""
+
+        vcf = DataField("variants:vcf", label="Input VCF file")
+
+        vcf_fields = ListField(
+            StringField(),
+            label="Fields from VCF file to include",
```
AnzeLovse, line 53: This could be simplified to Select VCF fields and then the description gives details about the format.

resolwe_bio/processes/variant_calling/variants_to_table.py
```diff
@@ -0,0 +1,123 @@
...
+        """Output fields for GATK VariantsToTable."""
+
+        txt = FileField(label="Tab-delimited file with variants")
+        species = StringField(label="Species")
+        build = StringField(label="Build")
+
+    def run(self, inputs, outputs):
+        """Run analysis."""
+
+        variants_table = "variants_table.txt"
```
AnzeLovse, line 100: As this is tab-delimited maybe replace the suffix with `.tsv`.

resolwe_bio/processes/variant_calling/variants_to_table.py
```diff
@@ -0,0 +1,123 @@
...
+        class AdvancedOptions:
+            """Advanced options."""
+
+            gf_fields = ListField(
+                StringField(),
+                label="Include FORMAT/sample-level fields",
+                default=["GT"],
+            )
+            split_alleles = BooleanField(
+                label="Split multi-allelic records",
```
AnzeLovse, line 79: IMO it would be nice to have this full description

Split multi-allelic records into multiple lines
By default, a variant record with multiple ALT alleles will be summarized in one line, with per alt-allele fields (e.g. allele depth) separated by commas. This may cause difficulty when the table is loaded by an R script, for example. Use this flag to write multi-allelic records on separate lines of output. Fields that are not allele-specific will be duplicated.

PR #1096, opened 2021-11-05
Rewrite processes Upload expression and Upload expression Cuffnorm to Python

OtonicarJan:
… Python

## Checklist

    before non-breaking changes.
    commit(s).
* [ x] All output fields have a value assigned to them.


docs/CHANGELOG.rst
```diff
@@ -37,6 +37,8 @@ Changed
   ``workflow-bbduk-star-fc-quant-paired``,
   ``workflow-cutadapt-star-fc-quant-single``,
   ``workflow-cutadapt-star-fc-quant-wo-depletion-single``
+- Rewrite ``upload-expression`` and ``upload-expression-cuffnorm``
```
jkokosar, line 40: Please move this CHANGELOG entry to the Unreleased section.

resolwe_bio/processes/import_data/expressions.py
```diff
@@ -0,0 +1,417 @@
...
+                ("ENSEMBL", "ENSEMBL"),
+                ("NCBI", "NCBI"),
+                ("UCSC", "UCSC"),
+            ],
+        )
+        species = StringField(
+            label="Species",
+            description="Species latin name.",
+            allow_custom_choice=True,
+            choices=[
```
jkokosar, line 178: Drop the last two entries in this choices menu.

resolwe_bio/processes/import_data/expressions.py
```diff
@@ -0,0 +1,417 @@
...
+            rc = inputs.rc.import_file(imported_format="compressed")
+            exp = inputs.rc.import_file(imported_format="compressed")
+            stem = Path(rc).stem
+            supported_extensions = (".txt", ".tab")
+            exp_type = "RAW_COUNT"
+
+        elif inputs.exp and inputs.exp_type and not inputs.rc:
+            exp = inputs.exp.import_file(imported_format="compressed")
+            stem = Path(exp).stem
+            supported_extensions = ".tab"
```
jkokosar, line 243: Shouldn't here also be `(".txt", ".tab")`?

resolwe_bio/processes/import_data/expressions.py
```diff
@@ -0,0 +1,417 @@
...
+        else:
+            prepare_expression_set(
+                exp=exp,
+                exp_type=exp_type,
+                feature_dict=feature_ids_to_names,
+                outfile_name=f"{name}_expressions",
+            )
+
+        if inputs.rc:
+            outputs.rc = rc
```
jkokosar, line 293: In the original process, the allowed file suffix on upload was either .txt or .tab, but the process always saved the files with the suffix .tab (except in the case of expression_set output field, where the suffix is .txt). For consistency, I think the suffix should be changed also in this process, so that if the file to be saved in exp or rc fields is uploaded as .txt, the suffix is converted to .tab before saving the file to the outputs.

resolwe_bio/tests/processes/test_upload.py
```diff
@@ -52,6 +52,9 @@ def test_bam_upload(self):
     @with_resolwe_host
     @tag_process("upload-expression")
     def test_upload_expression(self):
+        input_folder = Path("test_upload_expression") / "input"
```
jkokosar, line 55: Thank you for improving the test file folder structure organization!

jkokosar (commented):
This PR removes the ``upload-expression-star`` process. If this process will not be removed in the related PR (rewrite of STAR to Python), this change will need to go in the CHANGELOG as the braking change. FYI @AnzeLovse

resolwe_bio/processes/import_data/expressions.py
OtonicarJan: In the old process only `".tab"` was the supported extension for exp, however if the supported extensions for exp and for rc are indeed not the same then I must add additional check in case where both, rc and exp are used as inputs.

resolwe_bio/processes/import_data/expressions.py
```diff
@@ -0,0 +1,441 @@
...
+        exp_set_json = JsonField(label="Expressions (json)")
+        source = StringField(label="Gene ID source")
+        species = StringField(label="Species")
+        build = StringField(label="Build")
+        feature_type = StringField(label="Feature type")
+
+    def run(self, inputs, outputs):
+        """Run analysis."""
+
+        supported_extensions = (".txt", ".tab", ".tsv", "")
```
jkokosar, line 228: Why is `""` allowed here?

resolwe_bio/processes/import_data/expressions.py
```diff
@@ -0,0 +1,441 @@
...
+                    f"The supported extensions are {supported_extensions}."
+                )
+            exp_type = inputs.exp_type
+
+        if not stem.endswith(supported_extensions):
+            self.error(
+                f"The imported file has unsupported file name extension. "
+                f"The supported extensions are {supported_extensions}."
+            )
+        name = stem[:-4]
```
jkokosar, line 269: If in either the `rc` or `exp` field, a file without extension (e.g. `my_expressions`) is uploaded with `rc = inputs.rc.import_file(imported_format="compressed")`, then the `stem = Path(rc).stem` would return `my_expressions`, correct? Thus, `name = stem[:-4]` would unintentionally truncate the file name.

resolwe_bio/tests/processes/test_upload.py
```diff
@@ -52,6 +52,9 @@ def test_bam_upload(self):
     @with_resolwe_host
     @tag_process("upload-expression")
     def test_upload_expression(self):
```
jkokosar, line 54: Extend the test with additional uploads of files with alternative extentions (txt.gz, ..) to test if the renaming works as expected.

jkokosar:
@OtonicarJan please rebase. Add the **BACKWARD INCOMPATIBLE:** changelog entry before regular changelog entries.

jkokosar (approved):

PR #1098, opened 2021-11-09
Rewrite Seqtk sample (single-end and paired-end) processes to Python

OtonicarJan:
## Checklist

    before non-breaking changes.
    commit(s).
* [ x] All output fields have a value assigned to them.


resolwe_bio/processes/support_processors/seqtk.py
```diff
@@ -0,0 +1,324 @@
...
+        )
+        fastqc_archive = ListField(
+            FileField(),
+            label="Download FastQC archive",
+        )
+
+    def run(self, inputs, outputs):
+        """Run analysis."""
+
+        if inputs.advanced.fraction and not 0 < inputs.advanced.fraction < 1.0:
```
jkokosar, line 109: Please check if this should be `0 < inputs.advanced.fraction <= 1` instead

resolwe_bio/processes/support_processors/seqtk.py
```diff
@@ -0,0 +1,324 @@
...
+        )
+        fastqc_archive2 = ListField(
+            FileField(),
+            label="Download mate 2 FastQC archive",
+        )
+
+    def run(self, inputs, outputs):
+        """Run analysis."""
+
+        if inputs.advanced.fraction and not 0 < inputs.advanced.fraction < 1.0:
```
jkokosar, line 255: See above

jkokosar (approved):
A small comment, otherwise looks good!

docs/CHANGELOG.rst
```diff
@@ -38,6 +38,8 @@ Changed
   ``workflow-cutadapt-star-fc-quant-single``,
   ``workflow-cutadapt-star-fc-quant-wo-depletion-single``
 - Replace ANONYMOUS_USER_ID with ANONYMOUS_USER_NAME in settings.py
+- Rewrite processes ``seqtk-sample-single`` and ``seqtk-sample-paired``
```
jkokosar, line 41: Move the changelog entry to the unreleased section when rebasing

resolwe_bio/processes/support_processors/seqtk.py
AnzeLovse: FloatFields also support the range attribute that could replace this if sentence. 
```
fraction = FloatField(
    label="Fraction",
    required=False,
    range=[0.0, 1.0],
    description=...,
)
```

PR #1099, opened 2021-11-11
Add GATK MergeVCFs process

jkokosar:
## Checklist

    before non-breaking changes.
    commit(s).

### Processes ###

  can be performed).
  homepage and publication.

  tests).
  must be manually tagged and pushed to Amazon ECR.

### Tests ###


resolwe_bio/processes/variant_calling/gatk_merge_vcfs.py
```diff
@@ -0,0 +1,105 @@
...
+            )
+
+        with open(vcf_list, "w") as input_vcfs:
+            for vcf in inputs.vcfs:
+                input_vcfs.write(f"{vcf.output.vcf.path}\n")
+
+        args = [
+            f"I={vcf_list}",
+            f"O={combined_variants}",
+        ]
```
AnzeLovse, line 83: We should also support setting the garbage collection options and java heap size in advanced options. 
`--java-options -XX:ParallelGCThreads= and -Xmx`
Both can greatly impact performance and memory usage.

resolwe_bio/processes/variant_calling/gatk_merge_vcfs.py
```diff
@@ -0,0 +1,105 @@
...
+                    "sequence file does not match the species of the input VCFs."
+                )
+            if inputs.ref_seq.output.build != build:
+                self.error(
+                    "The genome build information of the provided reference "
+                    "sequence file does not match the build of the input VCFs."
+                )
+            args.append(f"D={inputs.ref_seq.output.fasta_dict.path}")
+
+        return_code, _, _ = Cmd["gatk"]["MergeVcfs"][args] & TEE(retcode=None)
```
AnzeLovse, line 98: We could assign `return_code, stdout, stderr` and then print `stdout, sdterr` in case the tool fails. This should provide more information about why the tool failed in the standard output file.

AnzeLovse (approved):
Looks good. Consider adding my suggestions for scalability and debugging.

AnzeLovse (commented):
Please also add a CHANGELOG entry.

resolwe_bio/processes/variant_calling/gatk_merge_vcfs.py
```diff
@@ -0,0 +1,140 @@
...
+            java_gc_threds = IntegerField(
+                label="Java ParallelGCThreads",
+                default=2,
+                description="Sets the number of threads used during parallel phases of the garbage collectors.",
+            )
+
+            max_heap_size = IntegerField(
+                label="Java maximum heap size (Xmx)",
+                default=12,
+                description="Set the maximum Java heap size.",
```
AnzeLovse, line 71: Add that this is in GB to avoid allocating too much memory.

AnzeLovse (approved):

PR #1100, opened 2021-11-11
Add java memory settings in GenomicsDBImport

AnzeLovse:
## Checklist

    before non-breaking changes.
    commit(s).


resolwe_bio/processes/variant_calling/create_genomicsdb.py
```diff
@@ -169,6 +182,17 @@ def run(self, inputs, outputs):
         else:
             self.error("Intervals file is required for creating a new database.")
 
+        temp_dir = Path("temp/")
```
jkokosar, line 185: Better approach is to use the temp dir as defined by env variables: `temp_dir = os.environ.get("TMPDIR")`

AnzeLovse: Thanks, I will replace it.

jkokosar (approved):

PR #1110, opened 2021-12-06
Rewrite GO Enrichment analysis process to Python

OtonicarJan:
## Checklist

    before non-breaking changes.
    commit(s).


resolwe_bio/processes/go_enrichment/go.py
```diff
@@ -0,0 +1,143 @@
...
+            description="Specify species name. This field is required if gene subset is set.",
+            choices=[
+                ("Homo sapiens", "Homo sapiens"),
+                ("Mus musculus", "Mus musculus"),
+                ("Rattus norvegicus", "Rattus norvegicus"),
+                ("Dictyostelium discoideum", "Dictyostelium discoideum"),
+            ],
+        )
+        pval_threshold = FloatField(
+            label="P-value threshold", default=0.1, required=False
```
jkokosar, line 59: Please remove `required=False` as it is redundant if the default value is set.

resolwe_bio/processes/go_enrichment/go.py
```diff
@@ -0,0 +1,143 @@
...
+        )
+        min_genes = IntegerField(
+            label="Minimum number of genes",
+            default=1,
+            required=False,
```
jkokosar, line 64: The same as above.

resolwe_bio/processes/go_enrichment/go.py
```diff
@@ -0,0 +1,143 @@
...
+                        mappings[m.source_id] = m.target_id
+                    else:
+                        self.warning(f"Mapping {m} returned multiple times.")
+
+            if len(inputs.genes) > len(mappings):
+                self.warning("Not all features could be mapped.")
+
+            target_ids = mappings.values()
+
+        with tempfile.NamedTemporaryFile() as input_genes:
```
jkokosar, line 121: This way of calling the `processor` tool from Python was necessary before the support for `plumbum` was introduced to Resolwe. I think you should be able to replace this part of the code with a standard `Cmd["processor"][...]()` call. If the tool emits to stdout, you can use something like `(Cmd["processor"][...] > "terms.json")()`

jkokosar (approved):

PR #1113, opened 2021-12-13
Small fixes

jkokosar:
- fix VEP process Data name
- remove unused output in STAR aligner process


resolwe_bio/processes/variant_calling/ensembl_vep.py
```diff
@@ -42,7 +42,7 @@ class EnsemblVep(Process):
             "storage": 200,
         },
     }
-    data_name = '{{vcf.file|default("?") }}'
+    data_name = "Annotated variants (VEP)"
```
AnzeLovse, line 45: Does this require an entry in the CHANGELOG?

AnzeLovse (approved):

PR #1116, opened 2021-12-15
Unify feature endpoints and add an action for pasted genes

dblenkus:


gregorjerse (approved):
Looks nice and clean.

PR #1119, opened 2021-12-20
Change Ensembl-VEP version check in Ensembl-VEP process

OtonicarJan:
## Checklist

    before non-breaking changes.
    commit(s).

### DIFF
--- a/resolwe_bio/processes/variant_calling/ensembl_vep.py
+++ b/resolwe_bio/processes/variant_calling/ensembl_vep.py
@@ -14,6 +14,35 @@ from resolwe.process import (
 )
 
 
+def check_version(stdout):
+    """Check version of Ensembl-VEP.
+
+    Example of the first part of an output from the command `vep --help`:
+    #----------------------------------#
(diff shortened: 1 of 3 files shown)


resolwe_bio/processes/variant_calling/ensembl_vep.py
```diff
@@ -14,6 +14,36 @@
...
+    Help: dev@ensembl.org , helpdesk@ensembl.org
+    Twitter: @ensembl
+    """
+    with open(vep_output, "w") as f:
+        f.write(stdout + "\n")
+    vep_version = int(
+        float(
+            [line for line in open(vep_output).readlines() if "ensembl-vep" in line][
+                0
+            ].split()[2]
```
AnzeLovse, line 41: Writing to file might be redundant because it is not used anywhere else. Would something like `next((line for line in stdout.split("\n") if "ensembl-vep" in line)).split()[2]` be ok?

AnzeLovse (approved):

PR #1124, opened 2022-01-05
Rewrite General RNA-Seq pipeline to Python

OtonicarJan:
## Checklist

    before non-breaking changes.
    commit(s).


resolwe_bio/processes/workflows/bbduk_star_featurecounts_qc.py
```diff
@@ -0,0 +1,802 @@
...
+    process_type = "data:workflow:rnaseq:featurecounts:qc"
+    category = "Pipeline"
+
+    class Input:
+        """Input fields."""
+
+        class Preprocessing:
+            """Preprocessing with BBDuk."""
+
+            reads = DataField(
```
jkokosar, line 56: I suggest to exclude the required inputs (reads, genome, etc) from the corresponding parent groups and leave them on the top level. List the groups of non-required inputs after the required inputs.  Boolean field `show_advanced` can then be omitted.

resolwe_bio/processes/workflows/bbduk_star_featurecounts_qc.py
```diff
@@ -0,0 +1,802 @@
...
+                    "do not soft-clip. Extend5pOfRead1: fully extend only the 5p of the read1, all "
+                    "other ends: local alignment. Extend5pOfReads12: fully extend only the 5' of the "
+                    "both read1 and read2, all other ends use local alignment.",
+                    default="Local",
+                )
+
+            class TwoPassOptions:
+                """Two-pass mapping options."""
+
+                two_pass_mode = BooleanField(
```
jkokosar, line 320: This is defined but not used anywhere. STAR two-pass is most probably not relevant for this pipeline.

resolwe_bio/processes/workflows/bbduk_star_featurecounts_qc.py
```diff
@@ -0,0 +1,802 @@
...
+            annotation = DataField(
+                "annotation",
+                label="Annotation",
+                description="GTF and GFF3 annotation formats are supported",
+            )
+            show_advanced = BooleanField(
+                label="Advanced options",
+                default=False,
+            )
+            assay_type = StringField(
```
jkokosar, line 400: I'm in favour of exposing `assay_type` and related `cdna_index` inputs on the top-level, as this is a critical parameter that must not be selected in the wrong way (it must match the library type of the dataset being analysed). If the setting is hidden in the advanced section of the inputs, it is easy to get this wrong.

resolwe_bio/processes/workflows/bbduk_star_featurecounts_qc.py
```diff
@@ -0,0 +1,802 @@
...
+            )
+
+        class QualityControl:
+            """Quality control."""
+
+            rrna_reference = DataField(
+                "index:star",
+                label="Indexed rRNA reference sequence",
+                description="Reference sequence index prepared by STAR aligner indexing tool.",
+                required=False,
```
jkokosar, line 522: This is a required input.

resolwe_bio/processes/workflows/bbduk_star_featurecounts_qc.py
```diff
@@ -0,0 +1,802 @@
...
+            )
+            globin_reference = DataField(
+                "index:star",
+                label="Indexed Globin reference sequence",
+                description="Reference sequence index prepared by STAR aligner indexing tool.",
+                required=False,
```
jkokosar, line 528: This is a required input.

resolwe_bio/processes/workflows/bbduk_star_featurecounts_qc.py
jkokosar: It should also be on the top level of the inputs.

resolwe_bio/processes/workflows/bbduk_star_featurecounts_qc.py
```diff
@@ -0,0 +1,802 @@
...
+                else -1,
+                "quality_trim": "r",
+                "trim_quality": inputs.preprocessing.trim_quality,
+                "quality_encoding_offset": inputs.preprocessing.quality_encoding_offset,
+                "ignore_bad_quality": inputs.preprocessing.ignore_bad_quality,
+                "maxns": inputs.preprocessing.maxns,
+            },
+        }
+
+        if inputs.preprocessing.reads.type == "data:reads:fastq:single:":
```
jkokosar, line 587: Use `startswith()` for this check, as the pipeline should fit all subtypes of the reads objects, e.g. "data:reads:fastq:single:bbduk:" or "data:reads:fastq:paired:bbduk:"

resolwe_bio/processes/workflows/bbduk_star_featurecounts_qc.py
```diff
@@ -0,0 +1,802 @@
...
+            preprocessing = Data.create(
+                process=BioProcess.get_latest(slug="bbduk-single"),
+                input=input_bbduk,
+            )
+        elif inputs.preprocessing.reads.type == "data:reads:fastq:paired:":
```
jkokosar, line 592: Add also the `else` statement, where the workflow is terminated if the expected input type is not provided.

resolwe_bio/processes/workflows/bbduk_star_featurecounts_qc.py
```diff
@@ -0,0 +1,802 @@
...
+                "quality_trim": "r",
+                "trim_quality": inputs.preprocessing.trim_quality,
+                "quality_encoding_offset": inputs.preprocessing.quality_encoding_offset,
+                "ignore_bad_quality": inputs.preprocessing.ignore_bad_quality,
+                "maxns": inputs.preprocessing.maxns,
+            },
+        }
+
+        if inputs.preprocessing.reads.type == "data:reads:fastq:single:":
+            preprocessing = Data.create(
```
jkokosar, line 588: if you assign just the process slug to a variable here (`process_slug = "bbduk-single"`), you can use just one `Data.create(slug = process_slug, input = input_bbduk)` statement outside the `if elif else` block.

resolwe_bio/processes/workflows/bbduk_star_featurecounts_qc.py
```diff
@@ -0,0 +1,802 @@
...
+            "aligned_reads": alignment,
+            "n_reads": inputs.quantification.n_reads,
+            "assay_type": inputs.quantification.assay_type,
+            "annotation": inputs.quantification.annotation,
+            "feature_class": inputs.quantification.feature_class,
+            "feature_type": inputs.quantification.feature_type,
+            "id_attribute": inputs.quantification.id_attribute,
+        }
+
+        if inputs.quantification.cdna_index:
```
jkokosar, line 734: Add an additional check here for the `cdna_index` input. If the `cdna_index` is not given when the `assay_type` is set to auto, fail the workflow. Add this check to the beginning of the workflow, so no individual processes are triggered if this state is detected.

resolwe_bio/processes/workflows/bbduk_star_featurecounts_qc.py
```diff
@@ -0,0 +1,802 @@
...
+            "reads": preprocessing,
+            "n_reads": inputs.downsampling.n_reads,
+            "advanced": {
+                "seed": inputs.downsampling.advanced.seed,
+                "fraction": inputs.downsampling.advanced.fraction,
+                "two_pass": inputs.downsampling.advanced.two_pass,
+            },
+        }
+
+        if inputs.preprocessing.reads.type == "data:reads:fastq:single:":
```
jkokosar, line 753: Use `startswith()`

resolwe_bio/processes/workflows/bbduk_star_featurecounts_qc.py
```diff
@@ -0,0 +1,708 @@
...
+            preprocessing,
+            alignment,
+            downsampling,
+            quantification,
+            alignment_qc_rrna,
+            alignment_qc_globin,
+            idxstats,
+        ]
+
+        inputs_multiqc = {
```
romunov, line 703: Above chunk `multiqc_inputs` could be incorporated into this dict without an intermediate step as `multiqc_inputs` is not used anywhere else.

Renaming `inputs_multiqc` to `input_multiqc` would make it more in line with the rest of the input variable names used in this process.

romunov (approved):
Yuge work, nicely done!

docs/CHANGELOG.rst
```diff
@@ -33,6 +33,9 @@ Changed
   ``workflow-bbduk-star-htseq``).
 - **BACKWARD INCOMPATIBLE:** Unify Feature autocomplete and search
   endpoints into a single endpoint
+- **BACKWARD INCOMPATIBLE:** Rewrite processes
```
jkokosar, line 36: Move to the unreleased section

PR #1126, opened 2022-01-05
Rewrite Alignment-free RNA-seq workflow to Python

OtonicarJan:
## Checklist

    before non-breaking changes.
    commit(s).


resolwe_bio/processes/workflows/bbduk_salmon_qc.py
```diff
@@ -0,0 +1,400 @@
...
+        }
+
+        if inputs.reads.type.startswith("data:reads:fastq:single:"):
+            bbduk_slug = "bbduk-single"
+        elif inputs.reads.type == "data:reads:fastq:paired:":
+            input_bbduk["operations"]["trim_pairs_evenly"] = True
+            input_bbduk["operations"]["trim_by_overlap"] = True
+            bbduk_slug = "bbduk-paired"
+        else:
+            self.error("Wrong input type.")
```
romunov, line 289: Perhaps the error message could be a bit more specific regarding the input type.

resolwe_bio/processes/workflows/bbduk_salmon_qc.py
```diff
@@ -0,0 +1,400 @@
...
+                "two_pass": inputs.downsampling.two_pass,
+            },
+        }
+
+        if inputs.reads.type.startswith("data:reads:fastq:single:"):
+            seqtk_slug = "seqtk-sample-single"
+        elif inputs.reads.type.startswith("data:reads:fastq:paired:"):
+            seqtk_slug = "seqtk-sample-paired"
+        else:
+            self.error("Wrong input type.")
```
romunov, line 336: Perhaps the error message could be more specific about the input type.

resolwe_bio/processes/workflows/bbduk_salmon_qc.py
```diff
@@ -0,0 +1,400 @@
...
+            quantification,
+            downsampling,
+            alignment_qc,
+            alignment_qc_rrna,
+            alignment_qc_globin,
+            qorts,
+            idxstats,
+        ]
+
+        inputs_multiqc = {"data": multiqc_inputs}
```
romunov, line 398: Use `input_multiqc` to be consistent with the above input dicts. You can probably incorporate the above chunk (`multiqc_inputs`) directly into this dict without an intermediate list.

romunov (approved):
More or less done!

resolwe_bio/processes/workflows/bbduk_salmon_qc.py
```diff
@@ -0,0 +1,396 @@
...
+
+            seq_bias = BooleanField(
+                label="Perform sequence-specific bias correction",
+                default=True,
+                description="Perform sequence-specific bias correction.",
+            )
+            gc_bias = BooleanField(
+                label="Perform fragment GC bias correction",
+                default=False,
+                description="Perform fragment GC bias correction.",
```
jkokosar, line 154: This input is hardcoded as `True` for the paired-end inputs. We need to expand the description to explain the behaviour of this parameter for different input types. Since one can control this flag (disable the gc bias correction) only in the case of single-end inputs, maybe a better approach is not to set the default value to begin with. @romunov any preference?

resolwe_bio/processes/workflows/bbduk_salmon_qc.py
```diff
@@ -0,0 +1,398 @@
...
+
+        class Quantification:
+            """Quantification (Salmon)."""
+
+            seq_bias = BooleanField(
+                label="Perform sequence-specific bias correction",
+                default=True,
+                description="Perform sequence-specific bias correction.",
+            )
+            gc_bias = BooleanField(
```
jkokosar, line 151: This is now a field without a default value. Please mark it as a non-required input. Add additional logic to expand the input parameter dict with `gc_bias` input if it is provided by the user.

jkokosar (commented):
A small change is still needed, otherwise looks good! Please rebase.

jkokosar (approved):

PR #1138, opened 2022-02-02
Pin R package versions and bump GATK version in resolwebio/dnaseq:6.3.0 Docker image

OtonicarJan:
Pin R package versions 
Bump GATK version to 4.2.4.1

### DIFF
--- a/resolwe_bio/docker_images/dnaseq/packages-manual/r-packages.sh
+++ b/resolwe_bio/docker_images/dnaseq/packages-manual/r-packages.sh
@@ -2,16 +2,14 @@
 
 . /var/cache/build/packages-manual/common.sh
 
-# These steps need to be performed because certain packages produce a "database corrupt error". The solution is to
-# restart R, but this is not feasible during the docker built or building of dependencies. This chunk also installs
-# packages from bioconductor. This entire step should be considered temporary until we can get packrat to work for
-# packages from all universes (CRAN, Bioconductor).
-Rscript --slave --no-save --no-restore-history -e "install.packages('assertthat')"
+# Install R packages.
+# Install renv which is a pre-requisite to installing all other packages in next chunk.
+Rscript --slave --no-save --no-restore-history -e "install.packages('renv', repos = 'https://cloud.r-project.org')"
+Rscript --slave --no-save --no-restore-history -e "renv::settings\$bioconductor.version('3.13')"
 
-Rscript --slave --no-save --no-restore-history -e \
-"source('https://bioconductor.org/biocLite.R'); \
-BiocInstaller::biocLite(c('IRanges', 'GenomicRanges', 'Rsamtools'))\
-"
+sed -e 's/^#.*$//g' -e '/^$/d' /var/cache/build/packages-r.txt | \
+    Rscript --slave --no-save --no-restore-history \
+        -e "renv::install(readLines('stdin'), repos='https://cloud.r-project.org')"
 
 # XXX: This is unverifiable and thus may compromise the whole image.
 # XXX: Use notary (https://github.com/ropenscilabs/notary) when ready.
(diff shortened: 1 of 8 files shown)


resolwe_bio/docker_images/dnaseq/packages-r.txt
```diff
@@ -0,0 +1,2 @@
+bioc::IRanges@2.28.0
```
romunov, line 1: It would probably suffice to install existing versions. In what I believe to be the latest dnaseq image (6.2.0) I see the following versions:

```
chemut_1.0.1
Rsamtools_2.8.0
GenomicRanges_1.44.0
IRanges_2.26.0
```

These should also match the Bioconductor release of 3.13 (and not 3.14, which is the latest).

romunov (changes requested):
Pin R to version R 4.1.1 and I think we're good to go.

jkokosar (commented):
The image builds OK for me. Please rebase.

resolwe_bio/docker_images/dnaseq/packages-manual/r-packages.sh
```diff
@@ -2,18 +2,21 @@
...
 . /var/cache/build/packages-manual/common.sh
 
-# These steps need to be performed because certain packages produce a "database corrupt error". The solution is to
-# restart R, but this is not feasible during the docker built or building of dependencies. This chunk also installs
-# packages from bioconductor. This entire step should be considered temporary until we can get packrat to work for
-# packages from all universes (CRAN, Bioconductor).
-Rscript --slave --no-save --no-restore-history -e "install.packages('assertthat')"
+# Pin BiocManager version 3.13
+Rscript --slave --no-save --no-restore-history -e "install.packages('BiocManager')"
+Rscript --slave --no-save --no-restore-history -e "BiocManager::install(version = '3.13', ask = FALSE)"
```
romunov, line 7: Pin bioconductor version using https://rstudio.github.io/renv/reference/settings.html#settings and forego the installation of the whole bioconductor universe.

resolwe_bio/docker_images/dnaseq/packages-manual/r-packages.sh
```diff
@@ -2,18 +2,21 @@
 
 . /var/cache/build/packages-manual/common.sh
 
-# These steps need to be performed because certain packages produce a "database corrupt error". The solution is to
-# restart R, but this is not feasible during the docker built or building of dependencies. This chunk also installs
-# packages from bioconductor. This entire step should be considered temporary until we can get packrat to work for
-# packages from all universes (CRAN, Bioconductor).
-Rscript --slave --no-save --no-restore-history -e "install.packages('assertthat')"
+# Pin BiocManager version 3.13
+Rscript --slave --no-save --no-restore-history -e "install.packages('BiocManager')"
```
romunov, line 6: `BiocManager` is installed by `renv` when you refer to a package using `bioc::<package>` on installation. >ou can omit this since we don't care what version of `BiocManager` gets installed.

resolwe_bio/docker_images/dnaseq/packages-r.txt
```diff
@@ -0,0 +1,4 @@
+argparse@2.0.3
```
romunov, line 1: Install Rsamtools to the version compatible with BioC 3.13 ([2.8.0](https://www.bioconductor.org/packages/3.13/bioc/html/Rsamtools.html)).

We should bump only package versions for those packages that are deeclared (in DESCRIPTION of the package) to not work with R 4.1.1.
`argparse` has no R requirements and should be left as it is. `devtools` works with R > 3.0.2 and should be left as it is, too. IMO, this is fine, other than the missing Rsamtools mentioned above.

romunov (changes requested):
This is nearly done. A few more changes and we're good to go.

PR #1144, opened 2022-02-13
Add filters for species, tissue, treatment and outcome on BioCollection

plojyon:
## Checklist

    before non-breaking changes.
    commit(s).

### Processes ###

  can be performed).
  homepage and publication.

  tests).
  must be manually tagged and pushed to Amazon ECR.

### Tests ###


resolwe_bio/filters.py
```diff
@@ -32,6 +34,56 @@ class BioCollectionFilter(CollectionFilter):
...
+
+    def _filter_by_outcome(self, queryset, name, value):
+        locations = [
+            "descriptor__response_and_survival_analysis__clinical_benefit",
+            "descriptor__response_and_survival_analysis__confirmed_bor",
+            "descriptor__response_and_survival_analysis__unconfirmed_bor",
+            "descriptor__response_and_survival_analysis__pfs",
+            "descriptor__response_and_survival_analysis__os",
+            "descriptor__response_and_survival_analysis__dfs",
+            "descriptor__response_and_survival_analysis__ttp",
```
gregorjerse, line 84: I believe you should also add corresponding indices in the database, otherwise, these queries will be slow.

See https://github.com/genialis/resolwe-bio/blob/master/resolwe_bio/migrations/resolwe_indexes_2.sql for example how to add missing indices to JSON fields. 

Check also that the added indices are really applied to the query.

tests/urls.py
```diff
@@ -3,15 +3,16 @@
...
+from resolwe.flow.views import CollectionViewSet, EntityViewSet
+from resolwe_bio.filters import BioCollectionFilter, BioEntityFilter
 from resolwe_bio.kb.views import (
     FeatureViewSet, MappingSearchViewSet
 )
 
 from .routers import SearchRouter
 
 
+CollectionViewSet.filter_class = BioCollectionFilter
```
gregorjerse, line 15: Maybe instead of patching URLs, one could use the `Composer` class. Otherwise, the URLs in the genialis-base repository will also have to be patched.

See https://github.com/genialis/resolwe/blob/90e4eb1160ac3fe0c2f0ac1b6bd18700cd82da55/resolwe/composer.py#L8 

There is some guidance on how to use the composer in

https://github.com/genialis/resolwe/blob/90e4eb1160ac3fe0c2f0ac1b6bd18700cd82da55/resolwe/flow/filters.py#L189

I believe you have to extend the filter used in the `CollectionViewSet` class, which is `resolwe.flow.filters.CollectionFilter`. 

Please try it out and remove line above if it is working.

gregorjerse (changes requested):

resolwe_bio/extensions.py
```diff
@@ -1,9 +1,65 @@
...
+from django.db.models import Q
 from django_filters import rest_framework as filters
 
 from resolwe.composer import composer
 
 
+class ExtendedCollectionFilter:
+    """Collection filter extension."""
+
+    descriptor__general__species = filters.CharFilter(
```
gregorjerse, line 11: You have to return collections that contain at least one sample (instance of Entity class) that has a descriptor with the given species.

So something like this:

```
descriptor__general__species = filters.CharFilter(
        field_name="entity__descriptor__general__species", lookup_expr="contains", distinct=True
    )
```

The distinct is necessary or collection with multiple samples matching given criteria will be returned multiple times.


All other queries also have to be modified.

resolwe_bio/migrations/resolwe_indexes_3.sql
```diff
@@ -0,0 +1,50 @@
+-- Index for Django filter descriptor__general__species__icontains
+CREATE INDEX idx_collection_descriptor__general__species ON flow_collection USING gin (UPPER(descriptor #>> '{general,species}'::text[]) gin_trgm_ops );
```
gregorjerse, line 2: All indices have to be applied to the `flow_entity` table. Fix index names accordingly.

resolwe_bio/tests/unit/test_api.py
```diff
@@ -30,6 +31,87 @@ def _check_filter(
...
+    @classmethod
+    def setUpTestData(cls):
+        cls.contributor = get_user_model().objects.create_user(
+            username="joemiller",
+            email="contributor@test.com",
+            first_name="Joe",
+            last_name="Miller",
+        )
+
+        cls.descriptor_schema = DescriptorSchema.objects.create(
```
gregorjerse, line 53: Fix tests: multiple collections containing multiple samples ...

gregorjerse (changes requested):

docs/CHANGELOG.rst
```diff
@@ -64,6 +64,7 @@ Fixed
 Added
 -----
 - Add bcftools version 1.14 to ``resolwebio/common`` Docker image
+- Add filters for species, tissue, treatment and outcome on BioCollection
```
gregorjerse, line 67: I would add something more descriptive and avoid using non-existing term BioCollection. Maybe something like 'Add ExtendedCollectionFilter filter to allow filtering collections by samples containing given species, tissue ...'.

gregorjerse (changes requested):
Code now looks great. 
Please fix the commit message, rebase and then it can be merged.

gregorjerse (approved):

PR #1150, opened 2022-02-22
Add workflow RNA-seq Variant Calling

OtonicarJan:
    before non-breaking changes.
    commit(s).


resolwe_bio/processes/workflows/rnaseq_variant_calling.py
```diff
@@ -0,0 +1,400 @@
...
+
+            Data.create(
+                process=BioProcess.get_latest(slug="gatk-select-variants"),
+                input=input_select,
+            )
+
+        multiqc_tools = [inputs.reads, alignment, mark_duplicates, bqsr]
+
+        if inputs.preprocessing:
+            multiqc_tools.insert(1, preprocessing)
```
jkokosar, line 393: Nothing wrong, just FYI: MultiQC process/tool settings define the module order in the MultiQC report. There is no need to pass the objects in any order here.

resolwe_bio/processes/workflows/rnaseq_variant_calling.py
```diff
@@ -0,0 +1,400 @@
...
+            },
+            "output_options": {
+                "out_unmapped": inputs.alignment.out_unmapped,
+            },
+        }
+
+        if inputs.preprocessing:
+            input_star["reads"] = preprocessing
+        else:
+            input_star["reads"] = inputs.reads
```
AnzeLovse, line 293: This could be simplified and added to the dictionary above as `"reads": preprocessing if inputs.preprocessing else inputs.reads`.

resolwe_bio/processes/workflows/rnaseq_variant_calling.py
```diff
@@ -0,0 +1,400 @@
...
+            label="Perform reads processing with BBDuk",
+            default=True,
+            description="If your reads have not been processed, set this to True.",
+        )
+        dbsnp = DataField(
+            data_type="variants:vcf",
+            label="dbSNP file",
+            description="File with known variants.",
+        )
+        indels = DataField(
```
AnzeLovse, line 71: Broad uses multiple known sites for INDELs. It would be good to support this in our pipeline as well by changing this to `ListField`. We could even join `dbsnp` with `indels` field and just extend the description with what is expected.

resolwe_bio/processes/workflows/rnaseq_variant_calling.py
```diff
@@ -0,0 +1,400 @@
...
+            "advanced": {
+                "use_original_qualities": True,
+            },
+        }
+
+        if inputs.intervals_bed:
+            input_bqsr["intervals"] = inputs.intervals_bed
+
+        if inputs.indels:
+            input_bqsr["known_sites"].append(inputs.indels)
```
AnzeLovse, line 327: This should be modified if `ListField` is used for `inputs.indels`.

resolwe_bio/processes/workflows/rnaseq_variant_calling.py
```diff
@@ -0,0 +1,400 @@
...
+        bqsr = Data.create(
+            process=BioProcess.get_latest(slug="bqsr"),
+            input=input_bqsr,
+        )
+
+        input_hc = {
+            "alignment": bqsr,
+            "genome": inputs.ref_seq,
+            "dbsnp": inputs.dbsnp,
+            "mbq": 10,
```
AnzeLovse, line 338: Should this be replaced with stand_call_conf. It seems that best practices use `--standard-min-confidence-threshold-for-calling` flag.

resolwe_bio/processes/workflows/rnaseq_variant_calling.py
```diff
@@ -0,0 +1,400 @@
...
+        )
+
+        input_bqsr = {
+            "bam": split_ncigar,
+            "reference": inputs.ref_seq,
+            "known_sites": [inputs.dbsnp],
+            "read_group": inputs.read_group,
+            "advanced": {
+                "use_original_qualities": True,
+            },
```
AnzeLovse, line 320: `java_gc_threads` and `max_heap_size` could be added here.

AnzeLovse (commented):
This mostly looks good. I appreciate the extended descriptions of less obvious fields. The tests clearly show that the workflow works. Let's plan a proper validation once this is merged.

resolwe_bio/tests/workflows/test_wgs.py
```diff
@@ -159,3 +159,95 @@ def test_gatk_hc_gvcf(self):
             file_filter=filter_vcf_variable,
             compression="gzip",
         )
+
```
AnzeLovse, line 162: It might be more fitting to add tests under rnaseq or its own category.

resolwe_bio/processes/workflows/rnaseq_variant_calling.py
OtonicarJan: I will add `--standard-min-confidence-threshold-for-calling` flag. The option `mbq` is added here, so that it is set to the [default](https://gatk.broadinstitute.org/hc/en-us/articles/360037225632-HaplotypeCaller#--min-base-quality-score) value of the actual tool. The default value in the process `vc-gatk4-hc` is different.

resolwe_bio/processes/workflows/rnaseq_variant_calling.py
```diff
@@ -0,0 +1,400 @@
...
+        },
+    }
+    data_name = 'RNA-seq Variants ({{ reads|sample_name|default("?") }})'
+    version = "1.0.0"
+    process_type = "data:workflow:rnaseq:variants"
+    category = "Pipeline"
+    entity = {
+        "type": "sample",
+    }
```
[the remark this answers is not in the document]
jkokosar: We'll unify this to include workflow Data objects to the Sample entities where relevant.

jkokosar (approved):

PR #1155, opened 2022-03-01
Rework Orange metadata table upload processes

jkokosar:
- Deprecate existing `upload-orange-metadata` process and replace it with `upload-metadata-unique` and
  `upload-metadata` that support  1-to-1 and 1-to-n relations between Collection samples and metadata content, respectively. 
- The metadata file upload will fail, if the upload is not within a working Collection
- The metadata file is expected to contain a sample row with allowed header names

TODO: 
 - replace resolwebio/orange Docker image with the resolwebio/common image when additional Python package is installed in the common image.
 - deprecate resolwebio/orange Docker image


JureZmrzlikar (commented):
Could you explain a bit what is the use case for one-tomany metadata table? I don't get it.

resolwe_bio/processes/import_data/metadata.py
```diff
@@ -60,8 +62,28 @@ def change_suffix(path):
...
+    elif any(path.endswith(ext) for ext in ["tab", "tsv"]):
+        df = pd.read_csv(path, sep="\t")
+    elif path.endswith("csv"):
+        df = pd.read_csv(path)
+    else:
+        df = pd.DataFrame()
+    return df
+
+
+class UploadOrangeMetadataOneToOne(Process):
```
JureZmrzlikar, line 80: Is it at this point even necessary to use Orange keyword in any place in this process?

resolwe_bio/processes/import_data/metadata.py
```diff
@@ -60,8 +62,28 @@ def change_suffix(path):
...
+    """Upload one-to-one metadata table in Orange format.
```
JureZmrzlikar, line 81: How about: "Upload metadata file where each row corresponds to a sample"?

resolwe_bio/processes/import_data/metadata.py
```diff
@@ -60,8 +62,28 @@ def change_suffix(path):
...
+
+    The uploaded metadata table represents one-to-one (1:1) relation to
+    samples in the working collection. Metadata table must contain a column
+    with one of the following headers: "mS#Sample ID", "mS#Sample name" or
```
JureZmrzlikar, line 85: As we discussed this, we should remove the `mS#` part. Just use "Sample ID", "Sample name" or "Sample slug". We should handle legacy files in resdk.

resolwe_bio/processes/import_data/metadata.py
```diff
@@ -60,8 +62,28 @@ def change_suffix(path):
...
+    "mS#Sample slug".
 
     Orange can read files in native tab-delimited format, or can load
```
JureZmrzlikar, line 88: If we decide to remove the `mS#` prefixes, this description is no longer necessary.

resolwe_bio/processes/import_data/metadata.py
```diff
@@ -148,6 +173,189 @@ def run(self, inputs, outputs):
...
+                f"Missing sample column. The uploaded Orange table should "
+                f"contain one of the following columns: {sorted(sample_columns)}."
+            )
+
+        samples = Entity.filter(collection_id=collections[0].id)
+
+        if "mS#Sample ID" in df_data.columns:
+            col_samples = set([sample.id for sample in samples])
+            sample_header = "mS#Sample ID"
+            intersection = col_samples.intersection(set(df_data[sample_header].values))
```
JureZmrzlikar, line 190: No need to repeat this line 3 times. Just write once after this if-elif clause is done.

resolwe_bio/processes/import_data/metadata.py
```diff
@@ -148,6 +173,189 @@ def run(self, inputs, outputs):
...
+            )
+
+        outputs.table = str(path)
+        outputs.n_samples = n_samples
+        outputs.features = get_features_description(data)
+        outputs.target = get_target_description(data)
+        outputs.n_metas = len(data.domain.metas)
+
+
+class UploadOrangeMetadataOneToMany(Process):
```
JureZmrzlikar, line 223: I am just curious... Could we shorten this class by just subclassing the `UploadOrangeMetadataOneToOne`? And than overwrite the diffs. This would make the diffs stand out more clearly + removed the unnecessary duplication.
Maybe @gregorjerse would know if there are any potential problems with this approach.

JureZmrzlikar (commented):
After some discussion with @mstajdohar i think this is mature enough to polish and merge. But I would also like to see what other reviewers say. :-)

resolwe_bio/processes/import_data/metadata.py
```diff
@@ -60,8 +62,28 @@ def change_suffix(path):
     return new_path
 
 
-class UploadOrangeMetadata(Process):
-    """Upload metadata table in Orange format.
+def convert_to_data_frame(path):
+    """Convert the uploaded file to Pandas data frame."""
+    if path.endswith("xls"):
+        df = pd.read_excel(path, engine="xlrd")
```
AnzeLovse, line 68: No need to create `df` it can be immediately returned.

AnzeLovse (commented):
I think this is mostly good to go. Have we reached any conclusion regarding column name prefixes?

resolwe_bio/processes/import_data/metadata.py
AnzeLovse: This would be an ideal solution to avoid duplicated code. Another alternative would also be wrapping the majority of processing code in functions.

resolwe_bio/processes/import_data/metadata.py
```diff
@@ -60,8 +62,168 @@ def change_suffix(path):
...
+        table = FileField(label="Uploaded table")
+        n_samples = IntegerField(label="Number of samples")
+        features = StringField(label="Number of features")
+        target = StringField(label="Target class description")
+        n_metas = IntegerField(label="Number of meta attributes")
+
+    def run(self, inputs, outputs):
+        """Run the analysis."""
+
+        collections = Collection.filter(data__id=self.data.id)
```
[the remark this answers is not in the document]
mzganec: I vote in favor of `collections`. The plural form is linked to the return type, not the number of returned items in the iterable.

resolwe_bio/processes/import_data/metadata.py
mzganec: > Could we shorten this class by just subclassing the UploadOrangeMetadataOneToOne?

Probably the other way around — `UploadOrangeMetadataOneToOne` is stricter than `UploadOrangeMetadataOneToMany` in handling duplicate samples.

resolwe_bio/processes/import_data/metadata.py
```diff
@@ -114,42 +184,65 @@ class Output:
...
+        intersection = col_samples.intersection(df_samples)
+
+        if not intersection:
+            self.warning(
+                "None of the samples listed in the uploaded Sample metadata table "
+                "match the Samples in the working Collection."
+            )
+
         outputs.table = str(path)
         outputs.n_samples = n_samples
```
JureZmrzlikar, line 248: This is not true since there could be duplicates. Better use somthing like:

`outputs.n_samples = len(df_data[sample_headers[0]].unique())`

resolwe_bio/processes/import_data/metadata.py
```diff
@@ -1,93 +1,161 @@
...
-        target_description = (
-            "Multi-target; "
-            f"{len(data.domain.class_vars)} target variables "
-            f"{missing_in_class}"
+    class Input:
+        """Input field to process UploadMetadataUnique."""
+
+        src = FileField(
+            label="Table with metadata",
+            description="The table should be in Orange format and use "
```
JureZmrzlikar, line 65: No need for the table to be in Orange format. It just needs to be tabular end contain one of the required columns.

resolwe_bio/processes/import_data/metadata.py
```diff
@@ -1,93 +1,161 @@
...
-    attribute type and the third header line contains role information
-    to identify dependent features (class), irrelevant features (ignore)
-    or meta features (meta).
+        col_samples = {
+            getattr(sample, sample_columns[sample_headers[0]]) for sample in samples
+        }
 
-    For more information see Orange
-    [documentation](https://orange-visual-programming.readthedocs.io/loading-your-data/index.html#header-with-attribute-type-information).
+        df_samples = df_data[sample_headers[0]].values
```
JureZmrzlikar, line 124: If you define `df_samples = df_data[sample_headers[0]]` and use them one line lower as 
`intersection = col_samples.intersection(df_samples.values)` you can tham omit creating the `sample_s` variable. This would simplify code and also remove one of the sample-something variable that makes code a bit hairder to digest.

resolwe_bio/processes/import_data/metadata.py
```diff
@@ -1,93 +1,161 @@
...
+        path = Path(inputs.src.import_file(imported_format="extracted"))
 
-    Orange can read files in native tab-delimited format, or can load
-    data from any of the major standard spreadsheet file types, like CSV
-    and Excel. Native format starts with the names of attributes with
-    prefixes that define attribute type (continuous, discrete, time,
-    string) and role (class, meta, ignore, instance weights). Prefixes
-    are separated from the attribute name with a hash sign (“#”).
+        # change the file suffix if it is either upper or mixed case
+        if not path.suffix.islower():
```
JureZmrzlikar, line 89: You could pack all the logic here in function `change_suffix` (I would rather name it `lower_suffix`, more descriptive). This function would just return a path if it was lowercase, but otherwise do all the path fixing and info reporting.

Since the logic is duplicated in both processes this would also shorten the code.

resolwe_bio/processes/import_data/metadata.py
```diff
@@ -1,93 +1,161 @@
...
+            path = change_suffix(path)
+            self.info(
+                "File extension of the table was replaced with a lower case version."
+            )
+
+        try:
+            df_data = read_tabular_data(path, error=self.error)
```
JureZmrzlikar, line 96: This try-catch approach could/should be put in `read_tabular_data`, especially since some sort of error handling is already happening there.

resolwe_bio/processes/import_data/metadata.py
```diff
@@ -1,93 +1,161 @@
...
+        except Exception as err:
+            self.error(f"It was not possible to read the provided data table. {err}")
+
+        sample_columns = {
```
JureZmrzlikar, line 100: Since this is a constant, used in both processes it could be set as a constant at the start of the file.

resolwe_bio/processes/import_data/metadata.py
```diff
@@ -1,93 +1,161 @@
...
+            "Sample ID": "id",
+            "Sample slug": "slug",
+            "Sample name": "name",
+        }
+
+        sample_headers = df_data.columns.intersection(sample_columns)
+        if len(sample_headers) != 1:
```
JureZmrzlikar, line 107: This check could also be moved to `read_tabular_data`. Then, when you get the `df_data` you are sure it contains just one header and just get it as `sample_header = df_data.columns.intersection(sample_columns)[0]`. This simplifies some of the below code.

resolwe_bio/processes/import_data/metadata.py
```diff
@@ -1,93 +1,161 @@
...
+            self.error(
+                f"The uploaded metadata table needs to contain "
+                f"exactly one of the following columns: "
+                f"{sorted(sample_columns.keys())}."
+            )
+
+        n_samples = len(df_data[sample_headers[0]].unique())
```
JureZmrzlikar, line 114: The check for "n_samples < 1" could also be moved to `read_tabular_data()`

JureZmrzlikar (approved):
The code looks OK, but I (and possibly future viewers / editors) would appreciate some simplification. Feel free to follow some of the suggestions.

PR #1197, opened 2022-06-02
Generalize bigWig calculation

AnzeLovse:
This updates the existing process `scale-bigwig` to be generally applicable (BIOINF-94). 

## Checklist

    before non-breaking changes.
    commit(s).


resolwe_bio/processes/support_processors/bam_conversion.py
```diff
@@ -99,41 +111,53 @@ class Input:
...
         species = StringField(label="Species")
         build = StringField(label="Build")
 
     def run(self, inputs, outputs):
         """Run the analysis."""
-        path = inputs.alignment.output.bam.path
-        basename = os.path.basename(path)
+        bam_path = Path(inputs.alignment.output.bam.path)
+        basename = bam_path.name
         assert basename.endswith(".bam")
```
jkokosar, line 144: This could be
```
assert bam_path.name.endswith(".bam")
name = bam_path.stem
```

resolwe_bio/processes/support_processors/bam_conversion.py
```diff
@@ -99,41 +111,53 @@ class Input:
...
-        out_index = f"{name}.bai"
 
-        with open(inputs.bedpe.output.bedpe.path) as f:
-            spike_count = f.readlines()
-            spike_count = len(spike_count)
-        scale_factor = inputs.scale / spike_count
+        if inputs.bedpe:
+            with open(inputs.bedpe.output.bedpe.path, "rb") as f:
+                spike_count = sum(1 for _ in f)
+            scale_factor = inputs.scale / spike_count
```
jkokosar, line 150: This could fail if `bedpe` is an empty file.

jkokosar (approved):
Please consider the suggested improvements, otherwise looks good!

resolwe_bio/processes/support_processors/bam_conversion.py
AnzeLovse: Good point, thank you.

PR #1204, opened 2022-07-20
Add optional calculation of variance in abundance estimates in Salmon Quant

OtonicarJan:
This PR adds an optional output of variance in abundance estimates of inferential replicates. Output file with variance currently looks something like this: 

```
FEATURE_ID    Variance
ENSG00000279493    2
```

```
tximport will also load in information about inferential replicates -- a list of matrices of the Gibbs samples from the posterior,
 or bootstrap replicates, per sample -- if these data are available in the expected locations relative to the files . The 
inferential replicates, stored in infReps in the output list, are on estimated counts, and therefore follow counts in the output 
list. By setting varReduce=TRUE , the inferential replicate matrices will be replaced by a single matrix with the sample 
variance per transcript/gene and per sample.

```


resolwe_bio/processes/expression/salmon.py
```diff
@@ -293,6 +293,37 @@ class Options:
                 "must be assigned to the transcriptome "
                 "for quantification to proceed.",
             )
+            num_bootstraps = IntegerField(
```
jkokosar, line 296: Bump the process version

jkokosar:
@romunov Please review when you can. Thank you.

OtonicarJan:
FYI @mzganec

resolwe_bio/tools/tximport_summarize.R
```diff
@@ -25,7 +26,18 @@ if (args$ignoreTxVersion) {
 write.table(tx2gene, file=args$gene_mapping, sep="\t", quote=FALSE, row.names=FALSE)
 
 # Run tximport
-txi_tpm = tximport(args$infile, type="salmon", tx2gene=tx2gene, ignoreTxVersion=args$ignoreTxVersion)
+if (args$variance) {
+    txi_tpm = tximport(args$infile, type="salmon", tx2gene=tx2gene, varReduce = TRUE,
+                    ignoreTxVersion=args$ignoreTxVersion)
+
+    tpm_variance = data.frame(Variance = txi_tpm$variance)
+    tpm_variance = cbind(FEATURE_ID = rownames(tpm_variance), tpm_variance)
```
romunov, line 34: Lines 33 and 34 could also be written as

```r
tpm_variance <- data.frame(FEATURE_ID = rownames(txi_tpm$variance), Variance = txi_tpm$variance)
```
but I see that the rest of the code is written in the original manner, so consider this just as a hint for future coding. :)

resolwe_bio/tools/tximport_summarize.R
```diff
@@ -25,7 +26,18 @@ if (args$ignoreTxVersion) {
...
+    rownames(tpm_variance) = NULL
```
romunov, line 35: Since you're using `row.names = FALSE` to write to a file, removing row names may not be necessary.

romunov (approved):
The R part looks good. If possible I would run the R script through `styler` package, but that's just a personal preference.

PR #1213, opened 2022-09-27
Flatten General descriptor schema

OtonicarJan:
## Checklist

    before non-breaking changes.
    commit(s).

### Processes ###

  can be performed).
  homepage and publication.

  tests).

### Tests ###

### DIFF
--- a/resolwe_bio/descriptors/unified-descriptor.yml
+++ b/resolwe_bio/descriptors/unified-descriptor.yml
@@ -1,7 +1,7 @@
 - slug: general
   name: General descriptor schema
   description: General descriptor schema.
-  version: 1.0.0
+  version: 2.0.0
   schema:
     - name: general_information
       label: General information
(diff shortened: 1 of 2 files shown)


romunov (approved):

OtonicarJan:
@uksurd88 @AnzeLovse @jkokosar

resolwe_bio/descriptors/unified-descriptor.yml
```diff
@@ -60,1404 +60,1412 @@
...
+            - label: 50%<
+              value: more_than_50
+            - label: N/A
+              value: na
+          description: |
+            PDL1 tumor proportion score category.
+          required: false
+        - name: biomarkers__chip
+          label: Biomarkers/CHIP
+          type: basic:integer
```
AnzeLovse, line 1251: Based on the data this looks like a binary field with values 0/1. IMO it would make sense to have this as a string field with two choices.
```
choices:
    - label: "Yes"
      value: "yes"
    - label: "No"
      value: "no"
```

resolwe_bio/descriptors/unified-descriptor.yml
```diff
@@ -60,1404 +60,1412 @@
...
+            - label: White
+              value: white
+            - label: Black or African American
+              value: black or african american
+            - label: American Indian or Alaska Native
+              value: american indian or alaska native
+            - label: Asian
+              value: asian
+            - label: Native Hawaian or Pacific Islander
+              value: native hawaian or pacific islander
```
AnzeLovse, line 159: Should be Hawaiian with double i.

resolwe_bio/descriptors/unified-descriptor.yml
```diff
@@ -60,1404 +60,1412 @@
...
+          choices:
+            - label: Positive
+              value: positive
+            - label: Negative
+              value: negative
+        - name: biomarkers__ebv
+          label: Biomarkers/EBV
+          type: basic:string
+          description: |
+            Epstein–Barr virus status.
```
AnzeLovse, line 1042: The character U+2013 "–" could be confused with the character U+002d "-", which is more common in source code.

resolwe_bio/descriptors/unified-descriptor.yml
```diff
@@ -60,1404 +60,1412 @@
...
+          choices:
+            - label: Positive
+              value: positive
+            - label: Negative
+              value: negative
+        - name: biomarkers__ebv_by_wes
+          label: Biomarkers/EBV by WES
+          type: basic:string
+          description: |
+            Epstein–Barr virus status by WES.
```
AnzeLovse, line 1053: Same as above.

resolwe_bio/descriptors/unified-descriptor.yml
```diff
@@ -60,1404 +60,1412 @@
...
+          label: Signatures/Stromal score
+          type: basic:decimal
+          description: |
+            Stromal classification of tumors by quantification of stromal cells in TME.
+          required: false
+        - name: signatures__emt_score
+          label: Signatures/EMT score
+          type: basic:decimal
+          description: |
+            Epithelial–mesenchymal transition score.
```
AnzeLovse, line 1339: The character U+2013 "–" could be confused with the character U+002d "-", which is more common in source code.

AnzeLovse (commented):
Great effort! Do we have any character length limit for dropdown choices in visualisations?

resolwe_bio/descriptors/unified-descriptor.yml
```diff
@@ -60,1404 +60,1412 @@
...
+          label: Biomarkers/PDL1 TPS category
+          type: basic:string
+          choices:
+            - label: <1%
+              value: less_than_1
+            - label: 1-49%
+              value: from_1_to_49
+            - label: 50%<
+              value: more_than_50
+            - label: N/A
```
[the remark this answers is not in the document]
uksurd88: @AnzeLovse I'd propose keeping them fixed thus its becomes a category rather than a spectrum of integers. Harvard data has these categories, so for their UC, we still need to map those categories to something else then.

resolwe_bio/descriptors/unified-descriptor.yml
```diff
@@ -60,1404 +60,1412 @@
...
+          type: basic:integer
+          description: |
+            Tumor mutation burden - the total number of genomic alterations
+            found in the cells of a cancer.
+            See http://purl.obolibrary.org/obo/NCIT_C150128.
+          required: false
+        - name: biomarkers__tmb_classification
+          label: Biomarkers/TMB assessment
+          type: basic:string
+          choices:
```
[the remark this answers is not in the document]
uksurd88: I think no, we are picking this one up from http://purl.obolibrary.org/obo/NCIT_C181335 with label as Tumor Mutation Burden Assessment. Essentially, it allows both quantitate and qualitative values, meanwhile our choices are _high, medium and low_ and they make sense

resolwe_bio/descriptors/unified-descriptor.yml
uksurd88: @OtonicarJan @AnzeLovse agree

resolwe_bio/descriptors/unified-descriptor.yml
AnzeLovse: Different sources also use different spelling ([see](https://ontobee.org/ontology/NCIT?iri=http://purl.obolibrary.org/obo/NCIT_C17049#:~:text=Unknown%3B%20Multiracial%3B%20White%3B-,Native%20Hawaiian%20or%20Other%20Pacific%20Islander,-%3B%20Black%20or%20African)).

PR #1224, opened 2022-11-03
Change processes of the General RNA-seq pipeline to support per-lane QC

OtonicarJan:
## Checklist

    before non-breaking changes.
    commit(s).

### Processes ###

  can be performed).
  homepage and publication.

  tests).

### Tests ###


resolwe_bio/processes/expression/featureCounts_beta.py
```diff
@@ -0,0 +1,980 @@
...
+    if len(filter_col) > 1:
+        exp["sum_count"] = exp[filter_col].sum(axis=1)
+        exp = exp.astype({"sum_count": int})
+
+        return exp[["sum_count"]].to_csv(
+            outfile,
+            index_label="FEATURE_ID",
+            header=["EXPRESSION"],
+            sep="\t",
+        )
```
AnzeLovse, line 51: The insides of if statment could just sum counts and specify the return column. Then the return can be the same in both cases.
```
    if len(filter_col) > 1:
        return_column = "sum_count"
        exp[return_column] = exp[filter_col].sum(axis=1)
    else:
        return_column = sample_name


    exp = exp.astype({return_column: int})
    return exp[[return_column]].to_csv(
            outfile,
            index_label="FEATURE_ID",
            header=["EXPRESSION"],
            sep="\t",
    )
```

resolwe_bio/processes/expression/featureCounts_beta.py
```diff
@@ -0,0 +1,980 @@
...
+
+        if inputs.exon_exon_junctions.junc_counts:
+            args.append("-J")
+
+        if inputs.exon_exon_junctions.junc_counts and inputs.exon_exon_junctions.genome:
+            args.extend(["-G", inputs.exon_exon_junctions.genome.path])
+
+        if inputs.general.by_read_group:
+            # Check if @RG is in header of the BAM file
+            (Cmd["samtools"]["view", "-H"][bam_file] > "header.sam")()
```
AnzeLovse, line 851: `-o` option could be used to avoid piping and to improve error handling. If the output is a `.txt` file the check can be done in Python.

resolwe_bio/processes/reads_processing/bbduk_beta.py
```diff
@@ -0,0 +1,1360 @@
...
+            f"xmin={inputs.header_parsing.x_min}",
+            f"ymin={inputs.header_parsing.y_min}",
+            f"xmax={inputs.header_parsing.x_max}",
+            f"ymax={inputs.header_parsing.y_max}",
+            f"entropy={inputs.complexity.entropy}",
+            f"entropywindow={inputs.complexity.entropy_window}",
+            f"entropyk={inputs.complexity.entropy_k}",
+            f"minbasefrequency={inputs.complexity.min_base_frequency}",
+            f"entropymask={inputs.complexity.entropy_mask}",
+            f"-Xmx{int(0.85*(self.requirements.resources.memory/num_of_lanes))}m",
```
AnzeLovse, line 601: Let's make sure we thoroughly test resource handling on real data.

resolwe_bio/tests/processes/test_alignment.py
```diff
@@ -310,6 +310,182 @@ def filter_star_report(line):
...
+            reads = self.prepare_reads(
+                [str(input_folder / "hs_single bbduk_star_htseq_reads_single.fastq.gz")]
+            )
+            single_lanes = self.prepare_reads(
+                [
+                    str(
+                        input_folder
+                        / "hs_single bbduk_star_htseq_reads_single.fastq.gz"
+                    ),
+                    str("hs sim_reads_single.fastq.gz"),
```
AnzeLovse, line 328: No need for `str`.

resolwe_bio/tests/processes/test_alignment.py
```diff
@@ -310,6 +310,182 @@ def filter_star_report(line):
...
+                        input_folder / "hs_paired_R2 workflow_bbduk_star_htseq.fastq.gz"
+                    )
+                ],
+            )
+            paired_lanes = self.prepare_paired_reads(
+                mate1=[
+                    str(
+                        input_folder / "hs_paired_R1 workflow_bbduk_star_htseq.fastq.gz"
+                    ),
+                    str("hs sim_reads1.fastq.gz"),
```
AnzeLovse, line 348: Same as above.

resolwe_bio/tests/processes/test_alignment.py
```diff
@@ -310,6 +310,182 @@ def filter_star_report(line):
...
+                ],
+                mate2=[
+                    str(
+                        input_folder / "hs_paired_R2 workflow_bbduk_star_htseq.fastq.gz"
+                    ),
+                    str("hs sim_reads2.fastq.gz"),
```
AnzeLovse, line 354: Same as above.

resolwe_bio/tests/processes/test_expression.py
```diff
@@ -494,6 +494,197 @@ def test_feature_counts(self):
...
+        )
+
+        expression_lanes = self.run_process(
+            "feature_counts-beta",
+            {
+                "aligned_reads": paired_lanes.id,
+                "annotation": annotation.id,
+            },
+        )
+        self.assertFile(
```
AnzeLovse, line 681: If the process modified the `feature_counts_output` the tests would have to cover that as well.

resolwe_bio/processes/reads_processing/bbduk_beta.py
```diff
@@ -0,0 +1,1360 @@
...
+        num_of_lanes = len(inputs.reads.output.fastq)
+        input_reads = []
+        for lane in range(num_of_lanes):
+            reads_basename = Path(inputs.reads.output.fastq[lane].path)
+            shutil.copy(reads_basename, Path.cwd())
+
+            new_path = Path(reads_basename.name.replace(" ", "_"))
+            Path(reads_basename.name).rename(new_path)
+            assert new_path.name.endswith(".fastq.gz")
+            name = new_path.name[:-9]
```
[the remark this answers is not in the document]
OtonicarJan: Files have to be renamed, since bbduk does not handle spaces in the file names. Should we rename files in the previous step (upload of FASTQs, SRA)?

resolwe_bio/processes/reads_processing/bbduk_beta.py
```diff
@@ -0,0 +1,1360 @@
...
+
+
+def prepare_fastqc(fastqgz, error):
+    """Prepare FastQC data for output."""
+    fastqc = []
+    fastqc_url = []
+    for fq in fastqgz:
+        reads_name = Path(fq).name.replace(".fastq.gz", "")
+        report_dir = Path("fastqc") / Path(f"{reads_name}_fastqc")
+        if not report_dir.is_dir():
```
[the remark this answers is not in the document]
OtonicarJan: This is a check in Upload fastq reads processes, however it is not needed here, since the step of preparing fastqc outputs is a bit different.

resolwe_bio/processes/support_processors/multiqc.py
```diff
@@ -499,9 +521,13 @@ def run(self, inputs, outputs):
 
             elif d.process.type == "data:expression:featurecounts:":
                 name = os.path.basename(d.output.counts_summary.path)
-                create_symlink(
-                    d.output.counts_summary.path, os.path.join(sample_dir, name)
+                sum_featurecounts_columns(
+                    summary_file=d.output.counts_summary.path, out_file=f"summed_{name}"
```
jkokosar, line 525: Is the out file `f"summed_{name}"` placed in the sample dir?

OtonicarJan: Fixed in the latest push.

resolwe_bio/processes/support_processors/multiqc.py
```diff
@@ -144,6 +144,28 @@ def create_lib_strand_table(samples, reports):
...
+        dtype={
+            "Status": str,
+        },
+    )
+    if len(exp.columns) > 1:
+        sum_column = exp.columns[0].split(":")[0]
+        exp[sum_column] = exp.sum(axis=1)
+
+        exp = exp.astype({sum_column: int})
+        return exp[[sum_column]].to_csv(
```
AnzeLovse, line 162: No need to return as `to_csv` returns None.

resolwe_bio/processes/reads_processing/bbduk_beta.py
```diff
@@ -0,0 +1,1380 @@
...
+            f"entropywindow={inputs.complexity.entropy_window}",
+            f"entropyk={inputs.complexity.entropy_k}",
+            f"minbasefrequency={inputs.complexity.min_base_frequency}",
+            f"entropymask={inputs.complexity.entropy_mask}",
+            f"-Xmx{int(0.85*(self.requirements.resources.memory/num_of_lanes))}m",
+        ]
+
+        if self.requirements.resources.cores >= num_of_lanes:
+            args.append(
+                f"threads={int(self.requirements.resources.cores//num_of_lanes)}"
```
AnzeLovse, line 646: If there are more lanes than cores this would be 0. Raising an informative error in such case should suffice.

jkokosar (approved):

AnzeLovse (approved):

PR #1237, opened 2023-01-26
Add mutation_status fields to general sample descriptor schema

jkokosar:
## Checklist

    before non-breaking changes.
    commit(s).

### Processes ###

  can be performed).
  homepage and publication.

  tests).

### Tests ###

### DIFF
--- a/resolwe_bio/descriptors/unified-descriptor.yml
+++ b/resolwe_bio/descriptors/unified-descriptor.yml
@@ -1,7 +1,7 @@
 - slug: general
   name: General descriptor schema
   description: General descriptor schema.
-  version: 2.1.0
+  version: 2.2.0
   schema:
     - name: general
       label: General information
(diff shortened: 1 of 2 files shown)


resolwe_bio/descriptors/unified-descriptor.yml
```diff
@@ -814,6 +814,326 @@
           label: Pharmacokinetics and pharmacodynamics/Description
           type: basic:string
           required: false
+        - name: mutation_status_kras
+          label: Mutation status/KRAS
+          type: basic:string
+          required: false
+          choices:
+            - label: KRAS G12D
```
OtonicarJan, line 822: Are these mutations sorted by the frequency of affected cases? I would sort them alphabetically and/or by amino acid position.

resolwe_bio/descriptors/unified-descriptor.yml
```diff
@@ -814,6 +814,326 @@
...
+            - label: NRAS
+              value: nras
+        - name: mutation_status_atm
+          label: Mutation status/ATM
+          type: basic:string
+          required: false
+          choices:
+            - label: ATM R337C
+              value: atm_r337c
+            - label: 'ATM K2811Sfs*46'
```
OtonicarJan, line 868: Frameshifts in SnpEff are written like this `Lys2811fs`. Three letter symbols can be changed to one letter symbol. What about the rest of frameshift annotation, should this be synchronized?

resolwe_bio/descriptors/unified-descriptor.yml
```diff
@@ -814,6 +814,326 @@
...
+            - label: CDKN2A H83Y
+              value: cdkn2a_h83y
+            - label: CDKN2A
+              value: cdkn2a
+        - name: mutation_status_rbm10
+          label: Mutation status/RBM10
+          type: basic:string
+          required: false
+          choices:
+            - label: 'RBM10 F913='
```
OtonicarJan, line 943: Synonymous variants are written like this in SnpEff `Phe913Phe`. Should this be synchronized?

resolwe_bio/descriptors/unified-descriptor.yml
```diff
@@ -814,6 +814,326 @@
...
+              value: 'rbm10_r97*'
+            - label: RBM10 Q481L
+              value: 'rbm10_q481l'
+            - label: 'RBM10 R295*'
+              value: 'rbm10_r295*'
+            - label: 'RBM10 E732*'
+              value: 'rbm10_e732*'
+            - label: RBM10
+              value: rbm10
+        - name: mutation_status_ptprd
```
OtonicarJan, line 959: Gene PTPRD has really low expression across many cancers, so it is hard to detect variants using RNA-seq VC in it.

resolwe_bio/descriptors/unified-descriptor.yml
```diff
@@ -814,6 +814,326 @@
...
+          label: Mutation status/PTPRD
+          type: basic:string
+          required: false
+          choices:
+            - label: PTPRD S431L
+              value: ptprd_s431l
+            - label: PTPRD
+              value: ptprd
+        - name: mutation_status_u2af1
```
OtonicarJan, line 968: Gene U2AF1 has really low expression across many cancers, so it is hard to detect variants using RNA-seq VC in it.

resolwe_bio/descriptors/unified-descriptor.yml
```diff
@@ -814,6 +814,326 @@
...
+          label: Mutation status/U2AF1
+          type: basic:string
+          required: false
+          choices:
+            - label: U2AF1
+              value: u2af1
+        - name: mutation_status_ntrk3
```
OtonicarJan, line 975: Gene NTRK3 has really low expression across many cancers, so it is hard to detect variants using RNA-seq VC in it.

resolwe_bio/descriptors/unified-descriptor.yml
```diff
@@ -814,6 +814,326 @@
...
+          type: basic:string
+          required: false
+          choices:
+            - label: POLE P286R
+              value: pole_p286r
+            - label: POLE V411L
+              value: pole_v411l
+            - label: POLE
+              value: pole
+        - name: mutation_status_lrp1b
```
OtonicarJan, line 1008: Gene LRP1B has really low expression across many cancers, so it is hard to detect variants using RNA-seq VC in it.

PR #1249, opened 2023-02-22
Add STAR-based RNA-seq workflow

OtonicarJan:
## Checklist

    before non-breaking changes.
    commit(s).

### Processes ###

  can be performed).
  homepage and publication.

  tests).

### Tests ###


resolwe_bio/processes/workflows/bbduk_star.py
```diff
@@ -0,0 +1,713 @@
...
+            },
+        }
+        if inputs.alignment.filtering_options.out_multimap_max:
+            input_star["filtering"][
+                "out_multimap_max"
+            ] = inputs.alignment.filtering_options.out_multimap_max
+
+        if inputs.alignment.filtering_options.out_mismatch_max:
+            input_star["filtering"][
+                "out_missmatch_max"
```
romunov, line 561: Typo, should be `out_mismatch_max`.

resolwe_bio/processes/workflows/bbduk_star.py
```diff
@@ -0,0 +1,713 @@
...
+            ] = inputs.alignment.filtering_options.out_mismatch_max
+
+        if inputs.alignment.filtering_options.out_mismatch_nl_max:
+            input_star["filtering"][
+                "out_missmatch_nl_max"
```
romunov, line 566: Typo, should be `out_mismatch_nl_max`.

romunov (changes requested):
A few typos, two critical but trivial to fix.

resolwe_bio/processes/workflows/bbduk_star.py
```diff
@@ -0,0 +1,713 @@
...
+                "(i.e. containing only non-canonical unannotated junctions) will be "
+                "suppressed. If you have stranded RNA-seq data, you do not need to "
+                "use any specific STAR options. Instead, you need to run Cufflinks with "
+                "the library option --library-type options. For example, "
+                "cufflinks --library-type fr-firststrand should be used for the standard "
+                "dUTP protocol, including Illumina's stranded Tru-Seq. "
+                "This option has to be used only for Cufflinks runs and not for STAR runs.",
+            )
+            noncannonical = BooleanField(
+                label="Remove non-cannonical junctions (Cufflinks compatibility)",
```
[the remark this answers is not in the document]
OtonicarJan: Changed to non-canonical, since STAR manual uses it.

resolwe_bio/processes/support_processors/multiqc.py
```diff
@@ -496,6 +498,12 @@ def run(self, inputs, outputs):
                     create_symlink(
                         d.output.stats.path, os.path.join(sample_dir, report)
                     )
+                    if d.output.gene_counts:
```
jkokosar, line 501: Bump the process version

resolwe_bio/processes/workflows/bbduk_star.py
```diff
@@ -0,0 +1,713 @@
...
+    """STAR-based RNA-seq pipeline.
+
+    First, reads are preprocessed by __BBDuk__ which removes adapters, trims
+    reads for quality from the 3'-end, and discards reads that are too short
+    after trimming. Compared to similar tools, BBDuk is regarded for its
+    computational efficiency.  Next, preprocessed reads are aligned by __STAR__
+    aligner. At the time of implementation, STAR is considered a
+    state-of-the-art tool that consistently produces accurate results from
+    diverse sets of reads, and performs well even with default settings. Aligned
+    reads are then summarized to genes by STAR. For more information see
```
jkokosar, line 27: > Aligned reads are then summarized to genes by STAR

STAR aligner counts and reports the number of aligned reads per gene while mapping.

jkokosar (approved):

resolwe_bio/processes/workflows/bbduk_star.py
```diff
@@ -0,0 +1,714 @@
...
+from resolwe.process.models import Process as BioProcess
+
+
+class WorkflowSTAR(Process):
+    """STAR-based RNA-seq pipeline.
+
+    First, reads are preprocessed by __BBDuk__ which removes adapters, trims
+    reads for quality from the 3'-end, and discards reads that are too short
+    after trimming. Compared to similar tools, BBDuk is regarded for its
+    computational efficiency.  Next, preprocessed reads are aligned by __STAR__
```
romunov, line 23: There is double space in front of "Next".

resolwe_bio/processes/workflows/bbduk_star.py
```diff
@@ -0,0 +1,714 @@
...
+                label="Custom adapter sequences",
+                required=False,
+                default=[],
+                description="Custom adapter sequences can be specified by inputting them "
+                "one by one and pressing Enter after each sequence.",
+            )
+            kmer_length = IntegerField(
+                label="K-mer length [k=]",
+                default=23,
+                description="Kmer length used for finding contaminants. "
```
romunov, line 127: Sometimes we use `kmer` and other times `k-mer`. Make it uniform throughout the input fields.

resolwe_bio/processes/workflows/bbduk_star.py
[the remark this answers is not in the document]
romunov: Thanks. Variable name could be modified, too.

romunov (approved):
My comments were more or less addressed, thanks.

PR #1256, opened 2023-03-27
Add an option to filter by genotype fields in GATK VariantFiltration

OtonicarJan:
## Checklist

    before non-breaking changes.
    commit(s).

### Processes ###

  can be performed).
  homepage and publication.

  tests).

### Tests ###


resolwe_bio/processes/variant_calling/gatk_variant_filtration.py
```diff
@@ -83,6 +83,29 @@ class Input:
...
+            "To filter by alternative allele depth, use the expression: "
+            "'vc.getGenotype('sample_name').getAD().1 < n",
+            required=False,
+        )
+        genotype_filter_name = ListField(
+            StringField(),
+            label="Names to use for the list of genotype filters",
+            description="Similar to the INFO field based expressions, but used on the FORMAT "
+            "(genotype) fields instead. Warning: filter names should be in the same order as "
+            "filter expressions. ",
```
AnzeLovse, line 106: Remove space after period.

resolwe_bio/processes/variant_calling/gatk_variant_filtration.py
```diff
@@ -83,6 +83,29 @@ class Input:
...
+            description="Similar to the INFO field based expressions, but used on the FORMAT "
+            "(genotype) fields instead. VariantFiltration will add the sample-level FT tag to "
+            "the FORMAT field of filtered samples (this does not affect the record's FILTER tag). "
+            "One can filter normally based on most fields (e.g. 'GQ < 5.0'), but the GT "
+            "(genotype) field is an exception. We have put in convenience methods so that "
+            "one can now filter out hets ('isHet == 1'), refs ('isHomRef == 1'), or homs "
+            "('isHomVar == 1'). Also available are expressions isCalled, isNoCall, isMixed, "
+            "and isAvailable, in accordance with the methods of the Genotype object."
+            "To filter by alternative allele depth, use the expression: "
+            "'vc.getGenotype('sample_name').getAD().1 < n",
```
AnzeLovse, line 98: Perhaps add what `sample_name` and `n` stand for `vc.getGenotype('sample_name').getAD().1 < n`. 

Also the example should end with `'` and the description with a period.

resolwe_bio/processes/variant_calling/gatk_variant_filtration.py
```diff
@@ -246,6 +289,29 @@ class Input:
...
```
AnzeLovse, line 304: Same as above.

resolwe_bio/processes/variant_calling/gatk_variant_filtration.py
```diff
@@ -246,6 +289,29 @@ class Input:
...
+            required=False,
+        )
+        genotype_filter_name = ListField(
+            StringField(),
+            label="Names to use for the list of genotype filters",
+            description="Similar to the INFO field based expressions, but used on the FORMAT "
+            "(genotype) fields instead. Warning: filter names should be in the same order as "
+            "filter expressions. ",
```
AnzeLovse, line 312: Same as above.

resolwe_bio/processes/variant_calling/gatk_variant_filtration.py
```diff
@@ -83,6 +83,29 @@ class Input:
...
+            StringField(),
+            label="Expressions used with FORMAT field to filter",
+            description="Similar to the INFO field based expressions, but used on the FORMAT "
+            "(genotype) fields instead. VariantFiltration will add the sample-level FT tag to "
+            "the FORMAT field of filtered samples (this does not affect the record's FILTER tag). "
+            "One can filter normally based on most fields (e.g. 'GQ < 5.0'), but the GT "
+            "(genotype) field is an exception. We have put in convenience methods so that "
+            "one can now filter out hets ('isHet == 1'), refs ('isHomRef == 1'), or homs "
+            "('isHomVar == 1'). Also available are expressions isCalled, isNoCall, isMixed, "
+            "and isAvailable, in accordance with the methods of the Genotype object."
```
AnzeLovse, line 96: Missing whitespace after period at the end.

resolwe_bio/processes/variant_calling/gatk_variant_filtration.py
```diff
@@ -246,6 +289,29 @@ class Input:
...
```
AnzeLovse, line 302: Same as above.

resolwe_bio/tests/processes/test_variant_calling.py
```diff
@@ -604,6 +623,23 @@ def test_variant_filtration(self):
             compression="gzip",
         )
 
+        filtering_multi_sample = self.run_process(
+            "gatk-variant-filtration",
+            {
+                "vcf": vcf_multi_sample.id,
+                "ref_seq": ref_seq.id,
+                "genotype_filter_expressions": ["AD.1 < 5"],
```
AnzeLovse, line 631: Is this an alternative to `vc.getGenotype('sample_name').getAD().1 < n`?

AnzeLovse: Thank you, `AD.1 < 5` seems more general.

jkokosar (approved):

PR #1272, opened 2023-05-17
Add listener command to query Feature objects

gregorjerse:
and use Redis to cache them

EXP-611

### DIFF
--- a/resolwe_bio/kb/listener_plugin.py
+++ b/resolwe_bio/kb/listener_plugin.py
@@ -1,8 +1,19 @@
 """Register knowledgebase tables in listener."""
+from typing import Tuple
+
 from django.contrib.auth import get_user_model
 from django.db.models import QuerySet
 
-from resolwe.flow.managers.listener.python_process_plugin import ExposeObjectPlugin
+from resolwe.flow.executors.socket_utils import Message, Response
+from resolwe.flow.managers.listener.listener import Processor
+from resolwe.flow.managers.listener.permission_plugin import ExposeObjectPlugin
+from resolwe.flow.managers.listener.plugin import (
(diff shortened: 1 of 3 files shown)


resolwe_bio/kb/listener_plugin.py
```diff
@@ -25,3 +31,75 @@ class ExposeMapping(ExposeObjectPlugin):
...
+
+    def handle_filter_features(
+        self,
+        data_id: int,
+        message: Message[str, str, list[str]],
+        manager: "Processor",
+    ) -> Response[list[int]]:
+        """Get the feature data based on the given identifiers."""
+        source, species, feature_ids, requested_fields = message.message_data
+        # Warning: the order of the fields must be the same as in the FeatureCache
```
dblenkus, line 66: Use that list to make sure this is kept in sync

resolwe_bio/kb/listener_plugin.py
```diff
@@ -25,3 +31,75 @@ class ExposeMapping(ExposeObjectPlugin):
...
+    name = "KnowledgeBase plugin"
+
+    def handle_filter_features(
+        self,
+        data_id: int,
+        message: Message[str, str, list[str]],
+        manager: "Processor",
+    ) -> Response[list[int]]:
+        """Get the feature data based on the given identifiers."""
+        source, species, feature_ids, requested_fields = message.message_data
```
dblenkus, line 65: What if there is more that one combination of source and species?

gregorjerse: Then user will have to perform more than one request. We have discussed this with Janez and he agreed on it.

dblenkus (approved):

PR #1276, opened 2023-06-20
Changed RNA-SeQC helper functions and tests

marcellevstek:
This pull request is intended to fix the conversion of UCSC type annotations to a suitable format for `collapse_annotation.py` script. The annotation should have a suitable gene>transcript>exon hierarchy.

    before non-breaking changes.
    commit(s).
* [X] All output fields have a value assigned to them.


resolwe_bio/processes/support_processors/rnaseqc_qc.py
```diff
@@ -102,13 +103,23 @@ def add_attributes(x):
         + df["gene_id"].astype(str)
     )
 
+    chromosome_order = df["chromosome"].unique().tolist()
```
romunov, line 106: Is this assuming the chromosome order in the annotation is in the correct (final) order, i.e. chr1, chr2... chrX, MT? What would happen, and do we care, if chromosomes are in some other order, i.e. chr1, chr10, chr11, chr2...?

docs/CHANGELOG.rst
```diff
@@ -16,10 +16,15 @@ Added
...
 -------
+- Change test files and modify test function for ``rnaseqc-qc`` process
+- Bump ``rnaseqc-qc`` process version
+- Rename ``filter_sense_rate`` to ``filter_rnaseqc_metrics`` in ``filter.py``
+  and add additional fields for filtering
 
 Fixed
 -----
-
+- Fix ``format_ucsc`` function in ``rnaseqc-qc`` process
```
romunov, line 26: The change is substantial, not just a bug fix. This also reflects in the version where the middle number was changed. I would moved this to `Changed` section. The message could be "Changed the ordering of exons in `format_ucsc` function of `rnaseq-qc` process".

romunov (changes requested):
The description of the PR is empty. Perhaps give a short description of why the change was necessary. Also, the PR title and commit message mention MultiQC, but I don't find any changes to this process, nor is there a mention of the change in the changelog.

Commit message seems a bit too vague. RNA-SeQC has already been added in the previous PR. This PR addresses the testing files and how UCSC annotations are handled. Consider making the commit message reflect that.

romunov (approved):

PR #1283, opened 2023-07-18
Add Euclidean distance metric in clustering-hierarchical-etc and find-similar processes

marcellevstek:
## Checklist

    before non-breaking changes.
    commit(s).


resolwe_bio/processes/clustering/hierarchical_clustering_etc.py
```diff
@@ -160,6 +160,7 @@ class Input:
         distance = StringField(
             label="Distance metric",
             choices=[
+                ("euclidean", "Euclidean"),
```
romunov, line 163: `get_distance_metric` function should also handle this distance.

Ideally there would also be a small test for this.

romunov (changes requested):

resolwe_bio/processes/clustering/hierarchical_clustering_etc.py
marcellevstek: Input field `metric` of `scipy.cluster.hierarchy.linkage` takes the argument `'euclidean'`.

romunov (approved):

romunov (approved):
Autosquash and LGTM.

PR #1287, opened 2023-08-28
Report gene coverage and assigned reads

marcellevstek:
for all samples in MultiQC

    before non-breaking changes.
    commit(s).

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -16,11 +16,11 @@ Added
 
 Changed
 -------
+- Report gene body coverage by ``rnaseqc-qc`` and assigned reads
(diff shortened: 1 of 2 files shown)


docs/CHANGELOG.rst
```diff
@@ -19,6 +19,8 @@ Changed
 
 Fixed
 -----
+- Report gene body coverage by ``rnaseqc-qc`` and assigned reads
```
romunov, line 22: With this kind of wording, this could also be considered a change, not a fix. If you move it to Changed, this should also be reflected in the version number (1.20.1 -> 1.21.0).

romunov (approved):

PR #1299, opened 2023-11-17
Use downsampled alignment for RNA-SeQC in workflows

marcellevstek:
    before non-breaking changes.
    commit(s).

### DIFF
--- a/resolwe_bio/processes/workflows/bbduk_star.py
+++ b/resolwe_bio/processes/workflows/bbduk_star.py
@@ -41,7 +41,7 @@ class WorkflowSTAR(Process):
         "expression-engine": "jinja",
     }
     data_name = "{{ reads|name|default('?') }}"
-    version = "1.3.1"
+    version = "1.4.0"
     entity = {
         "type": "sample",
     }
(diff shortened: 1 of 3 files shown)


resolwe_bio/processes/workflows/bbduk_star.py
```diff
@@ -41,7 +41,7 @@ class WorkflowSTAR(Process):
         "expression-engine": "jinja",
     }
     data_name = "{{ reads|name|default('?') }}"
-    version = "1.3.1"
+    version = "1.4.1"
```
jkokosar, line 44: This should be `version = "1.4.0"`

resolwe_bio/processes/workflows/bbduk_star_featurecounts_qc.py
```diff
@@ -46,7 +46,7 @@ class WorkflowBBDukStarFcQC(Process):
     entity = {
         "type": "sample",
     }
-    version = "6.1.1"
+    version = "6.2.1"
```
jkokosar, line 49: `version = "6.2.0"`

jkokosar (approved):

PR #1300, opened 2023-11-21
Add Samtools bedcov as a process

marcellevstek:
    before non-breaking changes.
    commit(s).


resolwe_bio/processes/samtools/samtools_bedcov.py
```diff
@@ -163,51 +133,52 @@ def run(self, inputs, outputs):
         name = f"{Path(inputs.bedfile.output.bed.path).stem}"
         coverage_report = f"{name}_coverage.txt"
```
marcellevstek, line 134: Maybe we should save this as a TSV file, as it is indeed tab delimited.

resolwe_bio/processes/samtools/samtools_bedcov.py
```diff
@@ -0,0 +1,184 @@
...
+    data_name = "{{ bam|name|default('?') }}"
+    version = "1.0.0"
+    entity = {"type": "sample"}
+    scheduling_class = SchedulingClass.BATCH
+
+    class Input:
+        """Input fields for SamtoolsBedcov."""
+
+        bams = ListField(
+            DataField(data_type="alignment:bam"), label="Input BAM files", required=True
```
jkokosar, line 87: Omit `required=True` as it is redundant.

resolwe_bio/processes/samtools/samtools_bedcov.py
```diff
@@ -0,0 +1,184 @@
...
+    category = "Samtools"
+    data_name = "{{ bam|name|default('?') }}"
+    version = "1.0.0"
+    entity = {"type": "sample"}
+    scheduling_class = SchedulingClass.BATCH
+
+    class Input:
+        """Input fields for SamtoolsBedcov."""
+
+        bams = ListField(
```
jkokosar, line 86: Each BAM from `bams` input is one sample. Defining entity = {"type": "sample"} will result in a new sample being crated, as the inputs are already part of individual/separate samples. 
If the goal is to attach this data to the existing samples, then each BAM should be processed separately (remove ListField). If there is a need for multisample bedcov process, this should be a separate analysis.

resolwe_bio/processes/samtools/samtools_bedcov.py
```diff
@@ -0,0 +1,184 @@
...
+
+    """
+
+    slug = "samtools-bedcov"
+    process_type = "data:alignment:bam:samtools"
+    name = "Samtools bedcov"
+    requirements = {
+        "expression-engine": "jinja",
+        "executor": {
+            "docker": {"image": "public.ecr.aws/genialis/resolwebio/rnaseq:6.0.0"},
```
romunov, line 70: Common image should be smaller (faster pull), and it also includes samtools.

resolwe_bio/processes/samtools/samtools_bedcov.py
```diff
@@ -0,0 +1,184 @@
...
+        """Input fields for SamtoolsBedcov."""
+
+        bams = ListField(
+            DataField(data_type="alignment:bam"), label="Input BAM files", required=True
+        )
+        bedfile = DataField(
+            data_type="bed",
+            label="Target BED file",
+            description="Target BED file with regions to extract.",
+            required=True,
```
jkokosar, line 93: As above, you can omit this.

resolwe_bio/processes/samtools/samtools_bedcov.py
jkokosar: I vote to retain .txt

marcellevstek:
What is your opinion on processes handling multiple samples in one analysis? Should we use a separate processes for single and multiple sample, or keep the option to output multiple samples in one file as in this code? Matjaz said that having multiple samples in one output file is not optimal, as most processes in Expressions work on individual samples instead of collections.

resolwe_bio/processes/samtools/samtools_bedcov.py
marcellevstek: Ok, we will then use it per sample.

resolwe_bio/processes/samtools/samtools_bedcov.py
```diff
@@ -0,0 +1,167 @@
...
+    Input columns : [chr, start, stop, coverage sum]
+    Output columns : [chr, start, stop, coverage mean]
+    """
+    coverage_df = pd.read_csv(
+        StringIO(output),
+        sep="\t",
+        names=["chrom", "start", "end", "coverage"],
+    )
+
+    coverage_df["coverage"] = coverage_df["coverage"] / (
```
jkokosar, line 37: You could round the mean coverage to 2 decimal places

resolwe_bio/processes/samtools/samtools_bedcov.py
```diff
@@ -0,0 +1,167 @@
...
+    process_type = "data:alignment:bam:samtools"
+    name = "Samtools bedcov"
+    requirements = {
+        "expression-engine": "jinja",
+        "executor": {
+            "docker": {"image": "public.ecr.aws/s4q6j6e8/resolwebio/common:3.1.0"},
+        },
+        "resources": {
+            "cores": 2,
+            "memory": 16384,
```
jkokosar, line 70: I suspect the process to be fairly efficient. We might get away with less resources. I suggest we drop this to 8GB RAM.

resolwe_bio/tests/processes/samtool_bedcov_temp.py
```diff
@@ -0,0 +1,26 @@
+skip_reads = StringField(
```
jkokosar, line 1: This is here probably by mistake?

jkokosar (commented):
Almost ready!

resolwe_bio/tests/processes/samtool_bedcov_temp.py
marcellevstek: Yes, got included by mistake.

jkokosar (approved):
Looks good! Please rebase

PR #1304, opened 2023-12-12
Change ``samtools-bedcov`` process

marcellevstek:
    before non-breaking changes.
    commit(s).

Rework output and labels in samtools-bedcov process. Coverage report is now compressed and labels shorted
to better render in frontend fields.


marcellevstek:
Output file is now compressed and labels changed so they are better formatted on the frontend.

romunov (commented):
PRs description could be improved, as well as the commit message. Perhaps mention that the report is now compressed. A commit message could be something along the lines of

> Rework output and labels in samtools-bedcov process
>
> Coverage report is now compressed and labels shorted
> to better render in frontend fields.

I also miss the changelog entry.

jkokosar:
Squash the two commits together. @romunov Changelog entry is not needed for this commit, as the feature was not released yet.

marcellevstek:
The function for calculating mean coverage was changed so it doesn't assume that there are only four initial columns. This is something that was OK at the start, but I wanted to make the code prettier and thus ran into problems.

jkokosar (approved):

PR #1307, opened 2024-01-08
Feature: mutations

gregorjerse:
## Checklist

    before non-breaking changes.
    commit(s).

### Processes ###

  can be performed).
  homepage and publication.

  tests).

### Tests ###

REID-674 REID-675


resolwe_bio/variants/models.py
```diff
@@ -0,0 +1,151 @@
...
+    variant = models.OneToOneField(
+        Variant, on_delete=models.CASCADE, related_name="annotation"
+    )
+    type = models.CharField(max_length=TYPE_MAX_LENGTH, blank=True, null=True)
+    annotation = models.CharField(max_length=ANNOTATION_MAX_LENGTH)
+    annotation_impact = models.CharField(max_length=ANNOTATION_IMPACT_MAX_LENGTH)
+    gene = models.CharField(max_length=GENE_MAX_LENGTH)
+    protein_impact = models.CharField(max_length=PROTEIN_IMPACT_MAX_LENGTH)
+    feature_id = ArrayField(
+        models.CharField(max_length=FEATURE_ID_MAX_LENGTH), default=list
```
jkokosar, line 83: As a single variant can have an impact on multiple (overlapping) transcripts, we want to retain all available information (list) fields as well as isolate the effect associated with the canonical transcript. Thus, we should have two instances of related fields (one as a list, one as a single value):

The following fields are to be renamed and should be optional and provided as a ``list`` of ``CharField`` fields:
* ``annotations`` 
* ``annotation_impacts`` 
* ``genes``
* ``protein_impacts``
* ``transcript_ids`` 

Additionally, add the following new fields which should be a single element ``CharField`` and also not required. 
* ``canonical_annotation``
* ``canonical_annotation_impact``
* ``canonical_gene``
* ``canonical_protein_impact``
* ``canonical_transcript_id``

jkokosar: fyi @AnzeLovse

resolwe_bio/variants/models.py
```diff
@@ -0,0 +1,151 @@
...
+GENOME_ASSEMBLY_MAX_LENGTH = 20
+CHROMOSOME_MAX_LENGTH = 20
+REFERENCE_MAX_LENGTH = 100
+ALTERNATIVE_MAX_LENGTH = 100
+
+# Metadata
+VARIANT_DATA_SOURCE_MAX_LENGTH = 100
+VARIANT_ANNOTATION_SOURCE_MAX_LENGTH = 100
+ANNOTATION_MAX_LENGTH = 200
+GENOTYPE_MAX_LENGTH = 100
```
marcellevstek, line 27: If this refers to genotypes in format "0/1" or "1/1", it could also be shortened.

resolwe_bio/variants/models.py
```diff
@@ -0,0 +1,151 @@
...
+
+    experiment = models.ForeignKey(
+        VariantExperiment,
+        on_delete=models.CASCADE,
+        related_name="variant_calls",
+        null=True,
+        blank=True,
+    )
+
+    # QC data.
```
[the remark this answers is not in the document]
marcellevstek: @jkokosar @mzganec FYI. In the MSI model, we also use variants that have values other than PASS in the filter column.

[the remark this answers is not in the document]
jkokosar: @marcellevstek updated comment above.

resolwe_bio/variants/models.py
jkokosar: Good point. Currently, the mutations_table output for GT field is more verbose, for example G/T, T/T, TAA/T. This then gets converted to 0, 1 or 2 format in [VariantsTable](https://github.com/genialis/resolwe-bio-py/blob/master/src/resdk/tables/variant.py#L350). We should do this encoding when writing to DB, and keep the length of this field at max 3 (so that either 0/1 or 1 format is supported).

resolwe_bio/tests/unit/test_api.py
```diff
@@ -2,7 +2,7 @@
 from rest_framework import status
 from rest_framework.test import APIRequestFactory, force_authenticate
 
-from resolwe.flow.models import Collection, Data, DescriptorSchema, Entity, Process
```
dblenkus, line 5: Removing the descriptor should go into a separate commit.

resolwe_bio/variants/views.py
```diff
@@ -0,0 +1,234 @@
...
+    OrderingFilter,
+)
+from resolwe.rest.serializers import SelectiveFieldMixin
+
+from .models import Variant, VariantAnnotation, VariantCall, VariantExperiment
+
+logger = logging.getLogger(__name__)
+
+
+class VariantSerializer(SelectiveFieldMixin, serializers.ModelSerializer):
```
dblenkus, line 28: Put serializers into a separate file.

resolwe_bio/variants/views.py
```diff
@@ -0,0 +1,234 @@
...
+            "genome_assembly",
+            "chromosome",
+            "position",
+            "reference",
+            "alternative",
+            "annotation",
+        ]
+
+
+class VariantFilter(CheckQueryParamsMixin, filters.FilterSet):
```
dblenkus, line 47: And also filters.

resolwe_bio/variants/models.py
```diff
@@ -0,0 +1,151 @@
...
+    )
+    clinical_significance = models.CharField(
+        max_length=CLINICAL_SIGNIFICANCE_MAX_LENGTH, blank=True, null=True
+    )
+    # Optional references to the external databases.
+    dbsnp_id = models.CharField(max_length=DBSNP_ID_MAX_LENGTH, blank=True, null=True)
+    clinical_var_id = models.CharField(
+        max_length=CLINICAL_VAR_ID_MAX_LENGTH, blank=True, null=True
+    )
+    data = models.ForeignKey(
```
dblenkus, line 96: Is it ok to delete these data when the Data object is deleted, but the sample may still exist?

resolwe_bio/variants/views.py
```diff
@@ -0,0 +1,234 @@
...
+            "clinical_significance": TEXT_LOOKUPS,
+            "dbsnp_id": TEXT_LOOKUPS,
+            "clinical_var_id": TEXT_LOOKUPS,
+        }
+
+
+class VariantAnnotationViewSet(mixins.ListModelMixin, viewsets.GenericViewSet):
+    """VariantAnnotation endpoint."""
+
+    queryset = VariantAnnotation.objects.all()
```
dblenkus, line 133: Do we need some permissions system to access this?

resolwe_bio/variants/views.py
```diff
@@ -0,0 +1,234 @@
...
+            "experiment__id": NUMBER_LOOKUPS,
+            "quality": NUMBER_LOOKUPS,
+            "depth": NUMBER_LOOKUPS,
+        }
+
+
+class VariantCallViewSet(mixins.ListModelMixin, viewsets.GenericViewSet):
+    """VariantCall endpoint."""
+
+    queryset = VariantCall.objects.all()
```
dblenkus, line 191: Permissions?

resolwe_bio/variants/views.py
```diff
@@ -0,0 +1,234 @@
...
+            "contributor__id": NUMBER_LOOKUPS,
+            "date": DATE_LOOKUPS,
+            "variant_data_source": TEXT_LOOKUPS,
+        }
+
+
+class VariantExperimentViewSet(mixins.ListModelMixin, viewsets.GenericViewSet):
+    """VariantExperiment endpoint."""
+
+    queryset = VariantExperiment.objects.all()
```
dblenkus, line 229: Permissions?

tests/urls.py
```diff
@@ -1,30 +1,38 @@
...
+from resolwe_bio.variants.views import VariantAnnotationViewSet, VariantViewSet
 
 from .routers import SearchRouter
 
-
 EntityViewSet.filterset_class = BioEntityFilter
 
 api_router = routers.DefaultRouter(trailing_slash=False)
-api_router.register(r'sample', EntityViewSet)
+api_router.register(r"sample", EntityViewSet)
```
dblenkus, line 16: Can you move reformatting to a separate commit and keep only relevant changes here?

dblenkus (changes requested):

pyproject.toml
```diff
@@ -7,3 +7,10 @@ build-backend = "setuptools.build_meta"
 
 [tool.black]
 target-version = ["py36", "py37", "py38", "py39", "py310", "py311"]
+
+[tool.mypy]
+plugins = ["mypy_django_plugin.main", "mypy_drf_plugin.main"]
```
dblenkus, line 12: Add a GHA to check MyPy annotations.

resolwe_bio/variants/models.py
gregorjerse: I changed the length to 3. If is also easy to increase the size in the future if required.

resolwe_bio/variants/models.py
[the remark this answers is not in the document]
gregorjerse: I have synced the model data with the description.

resolwe_bio/variants/migrations/0001_initial.py
```diff
@@ -0,0 +1,192 @@
...
+                    models.AutoField(
+                        auto_created=True,
+                        primary_key=True,
+                        serialize=False,
+                        verbose_name="ID",
+                    ),
+                ),
+                ("quality", models.FloatField()),
+                ("depth", models.PositiveIntegerField()),
+                ("filter", models.CharField(max_length=20)),
```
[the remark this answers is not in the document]
gregorjerse: Migration file was updated to reflect the latest changes.

resolwe_bio/variants/models.py
```diff
@@ -0,0 +1,151 @@
...
+        ]
+
+
+class VariantAnnotation(models.Model):
+    """Describes an annotation of a variant."""
+
+    variant = models.OneToOneField(
+        Variant, on_delete=models.CASCADE, related_name="annotation"
+    )
+    type = models.CharField(max_length=TYPE_MAX_LENGTH, blank=True, null=True)
```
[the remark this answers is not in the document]
gregorjerse: Fixed. Please check if the descriptions of the choices are OK.

resolwe_bio/variants/models.py
```diff
@@ -0,0 +1,151 @@
...
+    )
+    clinical_diagnosis = models.CharField(
+        max_length=CLINICAL_DIAGNOSIS_MAX_LENGTH, blank=True, null=True
+    )
+    clinical_significance = models.CharField(
+        max_length=CLINICAL_SIGNIFICANCE_MAX_LENGTH, blank=True, null=True
+    )
+    # Optional references to the external databases.
+    dbsnp_id = models.CharField(max_length=DBSNP_ID_MAX_LENGTH, blank=True, null=True)
+    clinical_var_id = models.CharField(
```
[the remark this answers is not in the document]
gregorjerse: Done.

tests/urls.py
gregorjerse: Sure.

pyproject.toml
gregorjerse: OK.

resolwe_bio/tests/unit/test_api.py
gregorjerse: I have removed this file from the commit, since decriptor & descriptor_schema fields are already removed in the master branch.

resolwe_bio/variants/models.py
```diff
@@ -0,0 +1,151 @@
...
+GENE_MAX_LENGTH = 100
+PROTEIN_IMPACT_MAX_LENGTH = 100
+FEATURE_ID_MAX_LENGTH = 200
+FILTER_MAX_LENGTH = 20
+
+
+class Variant(models.Model):
+    """Describe a variant in the database."""
+
+    # Because Django ORM cannot handle composite primary keys, each feature is
```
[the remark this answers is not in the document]
gregorjerse: Fixed.

resolwe_bio/variants/models.py
gregorjerse: This has been changed: there is a new model for every "row" in the data linked to the `VariantCall` model. Please check if it meets your acceptance criteria.

jkokosar: Are individual fields in VariantAnnotationTranscript allowed to be empty? Looks good otherwise.

dblenkus (approved):

PR #1325, opened 2024-03-28
Rename beta and remove legacy Cut&Run workflow

romunov:
    before non-breaking changes.
    commit(s).
* [X] All output fields have a value assigned to them.


resolwe_bio/processes/workflows/cut_and_run.py
```diff
@@ -26,13 +26,13 @@ class WorkflowCUTnRUN(Process):
 
     """
 
-    slug = "workflow-cutnrun-beta"
+    slug = "workflow-cutnrun"
     name = "Beta Cut & Run workflow"
```
jkokosar, line 30: Update the name as well.

jkokosar (approved):

PR #1334, opened 2024-04-22
Remove sample annotation functionality  from ``geo-import`` process

marcellevstek:
in ``geo-import`` process

REID-2336

    before non-breaking changes.
    commit(s).

### DIFF
--- a/resolwe_bio/processes/workflows/geo_import.py
+++ b/resolwe_bio/processes/workflows/geo_import.py
@@ -46,141 +46,6 @@ def create_metadata(gse, run_info):
     return metadata.set_index(["Sample name"], drop=False)
 
 
-def construct_annotation(metadata, sample_name):
-    """Construct sample annotations from metadata.
-
-    Dictionary with GEO metadata that matches the sample annotation
-    schema is created. Attributes under general that have no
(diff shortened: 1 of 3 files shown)


romunov (commented):
I think this is a minor change, not a bugfix, so 2.7.2 -> 2.8.2.

resolwe_bio/processes/workflows/geo_import.py
```diff
@@ -224,7 +200,7 @@ class GeoImport(Process):
         },
     }
     data_name = "{{ gse_accession }}"
-    version = "2.7.2"
+    version = "2.7.3"
```
jkokosar, line 203: version = "2.8.0"

resolwe_bio/tests/workflows/test_geo_import.py
```diff
@@ -327,47 +294,6 @@ def test_geo_chipseq(self):
 
         self.assertEqual(sample.annotations.count(), 11)
```
AnzeLovse, line 295: Remove this line.

jkokosar (approved):
Please squash the commits.

PR #1336, opened 2024-05-13
Add ``pca-beta`` process

marcellevstek:
REID-2321

## Checklist

    before non-breaking changes.
    commit(s).


resolwe_bio/tools/pca.py
romunov, line 1: If you removed `pca.py` from tools, perhaps you should also remove the yaml file?
https://github.com/genialis/resolwe-bio/blob/master/resolwe_bio/processes/clustering/pca.yml

marcellevstek: I will add it back until we deprecate that yml process.

resolwe_bio/processes/clustering/pca.py
```diff
@@ -0,0 +1,325 @@
...
+    expressions = get_csv(exp_paths)
+    expressions = transform(
+        error=error,
+        expressions=expressions,
+        log2=log2_transform,
+        z_score=standard_scaler,
+    )
+
+    # Filter out genes with low expression values
+    if rc_paths:
```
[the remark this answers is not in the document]
marcellevstek: Would be better, thanks for noticing.

resolwe_bio/processes/clustering/pca.py
```diff
@@ -0,0 +1,334 @@
...
+
+        exps = ListField(DataField(data_type="expression"), label="Expressions")
+        genes = ListField(
+            StringField(),
+            label="Gene subset",
+            required=False,
+            placeholder="new gene id, e.g. ENSG00000185982 (ENSEMBL database)",
+            description="Specify at least two genes or leave this field empty.",
+        )
+        source = StringField(
```
romunov, line 223: We could infer this from expression objects (and made sure they all match)?

resolwe_bio/processes/clustering/pca.py
```diff
@@ -0,0 +1,334 @@
...
+                "Gene ID database and Species must be provided if gene subset is selected."
+            )
+
+        for exp in expressions:
+            if exp.output.source != expressions[0].output.source:
+                self.error(
+                    "Input samples are of different Gene ID databases: "
+                    f"{exp.output.source} and {expressions[0].output.source}."
+                )
+            if exp.output.species != expressions[0].output.species:
```
romunov, line 280: This can be inferred internally without bothering the user to specify species.

resolwe_bio/processes/clustering/pca.py
```diff
@@ -0,0 +1,334 @@
...
+
+        expressions = inputs.exps
+
+        if inputs.genes and not (inputs.source and inputs.species):
+            self.error(
+                "Gene ID database and Species must be provided if gene subset is selected."
+            )
+
+        for exp in expressions:
+            if exp.output.source != expressions[0].output.source:
```
romunov, line 275: This can be inferred internally without bothering the user to specify source.

resolwe_bio/processes/clustering/pca.py
```diff
@@ -0,0 +1,334 @@
...
+        },
+    }
+    data_name = "{{ alignment|name|default('?') }}"
+    version = "1.0.0"
+    process_type = "data:pca"
+    category = "Enrichment and Clustering"
+    scheduling_class = SchedulingClass.INTERACTIVE
+    persistence = Persistence.TEMP
+
+    description = "Principal component analysis (PCA)"
```
romunov, line 210: If we decide to infer species and source internally, it could be specified here that we are assuming they all match.

resolwe_bio/processes/clustering/pca.py
```diff
@@ -0,0 +1,334 @@
+"""Beta principal component analysis process."""
+
+import json
+
+import numpy as np
+import pandas as pd
+from sklearn.decomposition import PCA as plotPCA
```
romunov, line 7: If you decide to rename the class to `PrincipalComponentAnalysis` (see below), you can import this using the native name `PCA`.

romunov (commented):
@jkokosar please see my review and give feedback on my thoughts on handling the species and source fields internally.

resolwe_bio/processes/clustering/pca.py
```diff
@@ -47,8 +47,9 @@ def get_pca(expressions=pd.DataFrame(), gene_labels=[], error=None):
         coordinates = [
             t[:2].tolist() if len(t) > 1 else [t[0], 0.0] for t in pca_expressions
         ]
+        print(coordinates)
```
romunov, line 50: Should this print statement be here?

jkokosar:
@romunov @marcellevstek the ``species`` and ``source`` inputs in the original process relate and inform the ``genes`` input, which is usually provided by the FE module (based on the genes basket state). Even though appropriate controls are implemented elsewhere, It must not be assumed that the source/species of input ``genes`` will always be the same as those of expression objects on the process inputs. It is thus not ok to remove the process inputs and related checks within the process.

resolwe_bio/processes/clustering/pca.py
```diff
@@ -0,0 +1,286 @@
...
+            label="Gene subset",
+            required=False,
+            placeholder="new gene id, e.g. ENSG00000185982 (ENSEMBL database)",
+            description="To subset your analysis to only a subset of genes, specify them here. "
+            "Specify at least two genes or leave this field empty.",
+        )
+        low_expression_filter = BooleanField(
+            label="Low expression filter",
+            default=True,
+            description="Filter the input expression matrix to remove genes "
```
romunov, line 227: Please also specify that raw counts need to be available for the expression (shouldn't work for cases where only tpms have been uploaded, for example).

resolwe_bio/tests/processes/test_pca.py
```diff
@@ -4,6 +4,78 @@
 
 
 class PcaProcessorTestCase(KBBioProcessTestCase):
+    @with_resolwe_host
+    @tag_process("pca-beta")
+    def test_beta_pca(self):
+        with self.preparation_stage():
+            expression_1 = self.prepare_expression(
```
romunov, line 11: One test case could be uploaded tpms with no rc and then running the PCA (`low_expression_filter` turned on). It should fail since no raw counts are available.

romunov (commented):
Looks good. Squash and rebase, please.

jkokosar (approved):

PR #1337, opened 2024-05-16
Use the updated version of xengsort tool in processes xengsort-index and xengsort-classify

jkokosar:
## Checklist

    before non-breaking changes.
    commit(s).

### Processes ###

  can be performed).
  homepage and publication.

  tests).

### Tests ###


resolwe_bio/tests/files/xengsort/output/index_stats.txt
```diff
@@ -1,59 +1,44 @@
...
-
-# 2022-05-17 08:23:29: Done.
+
+# xengsort index [2024-05-14 15:14:40]
+
+- (c) 2019-2023 by Sven Rahmann, Jens Zentgraf, Algorithmic Bioinformatics, Saarland University
+- Licensed under the MIT License
+
+# Creating index index_dir/xengsort_index.zarr [2024-05-14 15:14:40]
+- memory for hash table: 0.000 GiB (with 3 subtables)
```
AnzeLovse, line 8: To get the same result on my computer I had to add `--subtables 3` parameter which was 5 by default for me. To have stable tests I suggest we expose the parameter and set the default to 9, which [they use](https://gitlab.com/genomeinformatics/xengsort/-/blob/master/Snakefile#L96) for the Homo sapiens index.

resolwe_bio/processes/reads_processing/xengsort.py
```diff
@@ -38,15 +38,6 @@ def concatenate_files(filenames, out_fname, decompress=False):
...
-    # Keep only alphanumeric characters and underscores.
-    basename = re.sub(pattern=r"[\W]", repl="_", string=basename)
-    # Replace consecutive underscores with a single one.
-    basename = re.sub(pattern=r"(_)\1+", repl=r"\1", string=basename)
-    return f"{basename}.{suffix}"
-
-
 def concatenate_reads(filenames, out_fasta, error):
     """Concatenate reads in FASTQ files."""
     try:
```
AnzeLovse, line 43: Concatenation could be made faster by disabling decompression which is no longer needed for the new version of Xengsort. `decompress=True` can be changed to False one line below. Mate file names should then be modified accordingly.

docs/CHANGELOG.rst
```diff
@@ -17,6 +17,8 @@ Added
 Changed
 -------
 - **BACKWARD INCOMPATIBLE:** Require Resolwe 40.x
+- Use the updated xengsort version in processes ``xengsort-index`` and
```
AnzeLovse, line 20: This is a `BACKWARD INCOMPATIBLE` change.

resolwe_bio/processes/reads_processing/xengsort.py
```diff
@@ -156,7 +147,7 @@ class Advanced:
     class Output:
         """Output fields to process XengsortIndex."""
 
-        index = FileField(label="Xengsort index")
+        index = DirField(label="Xengsort index")
```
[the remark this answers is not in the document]
jkokosar: I explored that initially, but decided to go with the DirField so that we can use a single input reference (pointing to the dir + common index file prefix). By having the two files stored in different output fields, we would need to use symlink in the classify process, which is not so nice.

AnzeLovse:
Looks good, the tests now also pass on my computer. Please squash and rebase.

AnzeLovse (approved):

PR #1339, opened 2024-05-28
Support writing variant information from mutations-table process to variants application

jkokosar:
## Checklist

    before non-breaking changes.
    commit(s).

### Processes ###

  can be performed).
  homepage and publication.

  tests).

### Tests ###


resolwe_bio/tests/processes/test_support_processors.py
```diff
@@ -9,6 +9,7 @@
 from resolwe_bio.models import Sample
 from resolwe_bio.utils.filter import filter_comment_lines, filter_rnaseqc_metrics
 from resolwe_bio.utils.test import KBBioProcessTestCase
+from resolwe_bio.variants.models import Variant, VariantCall, VariantExperiment
```
gregorjerse, line 12: This does not belong in this commit.

resolwe_bio/process/runtime.py
```diff
@@ -21,10 +21,10 @@ def __init__(self, data: Data):
...
 
-    def add_variants(self, variants: List[Dict[str, Any]]):
+    def add_variants(self, data_source: str, variants: List[Dict[str, Any]]):
         """Add variants to the database."""
-        return communicator.add_variants(variants)
+        return communicator.add_variants((data_source, variants))
 
     def add_variants_annotations(self, variants: List[Dict[str, Any]]):
         """Add variants annotations to the database."""
-        return communicator.add_variants_asnnotations(variants)
```
gregorjerse, line 30: This does not really belong into this commit. 

I suggest you merge the commit  `Initialize the listener plugin`  with this one and reword the title to something like `Fix bugs in the variants app`.

resolwe_bio/processes/support_processors/mutations_table.py
```diff
@@ -240,6 +241,40 @@ def get_depth(variants_table, bam):
...
+
+    Variants are given as <allele1>/<allele2>, e.g. T/T or C/T
+
+    Encode these variants as:
+        - 0/1 for heterozygous variants
+        - 1/1 for homozygous variants
+    """
+    try:
+        allele_line = row.get("genotype", np.nan)
+        print("allele_line", allele_line)
```
gregorjerse, line 255: Is this here for debugging purposes?

resolwe_bio/processes/support_processors/mutations_table.py
```diff
@@ -240,6 +241,40 @@ def get_depth(variants_table, bam):
...
+
+    if allele1 == allele2 == row["alternative"]:
+        return "1/1"
+    else:
+        return "0/1"
+
+
+def extract_alt_allele_depth(row):
+    """Extract the depth of the alternative allele from the AD field."""
+    ad_line = row.get("unfiltered_allele_depth")
```
gregorjerse, line 273: When key must be present, I suggest you use `row["..."]` instead of the `get` method.

resolwe_bio/processes/support_processors/mutations_table.py
```diff
@@ -240,6 +241,40 @@ def get_depth(variants_table, bam):
...
+    _, alt = ad_line.split(",")
```
gregorjerse, line 274: You could also use `alt = ad_line.split(",")[1]`.

resolwe_bio/processes/support_processors/mutations_table.py
```diff
@@ -498,3 +537,60 @@ def run(self, inputs, outputs):
...
+        columns_map = {
+            "species": "species",
+            "genome_assembly": "genome_assembly",
+            "CHROM": "chromosome",
+            "POS": "position",
+            "REF": "reference",
+            "ALT": "alternative",
+        }
+
+        if "QUAL" in requested_fields:
```
gregorjerse, line 562: Maybe something like this is more readable.

```
optional_mapping = {"DP": "depth", "FILTER": "filter"...}
for key, value in optional_mapping.items():
    if key in reequired_fields:
        columns_map[key] = value
```

resolwe_bio/processes/support_processors/mutations_table.py
jkokosar: Yes, will remove.

resolwe_bio/processes/support_processors/mutations_table.py
```diff
@@ -240,6 +241,39 @@ def get_depth(variants_table, bam):
...
+
+    Variants are given as <allele1>/<allele2>, e.g. T/T or C/T
+
+    Encode these variants as:
+        - 0/1 for heterozygous variants
+        - 1/1 for homozygous variants
+    """
+    try:
+        allele_line = row.get("genotype", np.nan)
+        allele_re = r"^([ACGT]+)/([ACGT]+)$"
```
AnzeLovse, line 255: I am aware this is an existing regex used by resdk but it fails to capture some of the variants. In particular `*/G` and similarly structured ones. We could add in the asterisk in the pattern `"^([ATGC*]+)/([ATGC*]+)$"`.

resolwe_bio/processes/support_processors/mutations_table.py
```diff
@@ -498,3 +536,62 @@ def run(self, inputs, outputs):
...
+
+        output_table = output_table.rename(columns=columns_map)
+
+        if "genotype" in output_table.columns:
+            output_table["genotype"] = output_table.apply(
+                encode_variant_genotype, axis=1, result_type="reduce"
+            )
+
+        if "unfiltered_allele_depth" in output_table.columns:
+            output_table["unfiltered_allele_depth"] = output_table.apply(
```
AnzeLovse, line 592: This step could be vectorised to speed up the processing `output_table["unfiltered_allele_depth"].str.split(",").str[1].astype(int)`.

resolwe_bio/processes/support_processors/mutations_table.py
```diff
@@ -498,3 +536,62 @@ def run(self, inputs, outputs):
         outputs.genes = genes
         outputs.species = inputs.variants.output.species
         outputs.build = inputs.variants.output.build
+
+        # Store the reported variants in the Variant Database
+        data_source = "RNA-Seq Variant Calling Pipeline"
```
AnzeLovse, line 541: Should this be an advanced parameter with the default value "RNA-Seq Variant Calling Pipeline" since any VCF file annotated with snpeff can be used as an input to this process?

resolwe_bio/tests/processes/test_support_processors.py
```diff
@@ -1801,3 +1827,12 @@ def test_report_variants(self):
...
         )
+
+        variant_calls = VariantCall.objects.filter(
+            sample_id=ensembl_mutations.entity.id
+        )
+        self.assertEqual(len(variant_calls), 3)
+
+        variant_call_1 = variant_calls[0]
+        self.assertEqual(variant_call_1.genotype, "1/1")
+        self.assertEqual(variant_call_1.unfiltered_allele_depth, 2)
```
AnzeLovse, line 1838: This could also test the `genotype_quality` field since it is `None` in the previous example.

resolwe_bio/processes/support_processors/mutations_table.py
jkokosar: Fixed as suggested, thanks!

resolwe_bio/processes/support_processors/mutations_table.py
```diff
@@ -498,3 +536,62 @@ def run(self, inputs, outputs):
...
+
+        output_table["species"] = inputs.variants.output.species
+        output_table["genome_assembly"] = (
+            "GRCh38"
+            if "GRCh38" in inputs.variants.output.build
+            else inputs.variants.output.build
+        )
+
+        sample_fields_map = {
```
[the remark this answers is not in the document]
jkokosar: Sure, its better this way!

resolwe_bio/processes/support_processors/mutations_table.py
jkokosar: Added as an (advanced) input field, thanks!

AnzeLovse (approved):
All my comments have been addressed. @gregorjerse would you have another pass and approve if everything looks good.

PR #1353, opened 2024-07-15
Use partial testing

gregorjerse:
## Checklist

    before non-breaking changes.
    commit(s).

### Processes ###

  can be performed).
  homepage and publication.

  tests).

### Tests ###

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
       RESOLWE_REDIS_PORT: 56380
     strategy:
       matrix:
-        toxenv: ["py311", "py312"]
+        toxenv: ["py312"]
 
     services:
       postgres:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,7 @@ jobs:
     env:
       RESOLWE_POSTGRESQL_PORT: 55433
       RESOLWE_REDIS_PORT: 56380
+      RESOLWE_TEST_ONLY_CHANGES_TO: "origin/master"
     strategy:
       matrix:
         toxenv: ["py312"]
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -64,7 +63,7 @@ jobs:
       # "master" branch, we allow installing pre-releases with the pip command.
       - name: Allow pre-releases for master branch
         if: github.ref == 'refs/heads/master'
-        run: echo "TOX_COMMAND=${{ env.TOX_COMMAND }} --pre" >> $GITHUB_ENV
+        run: echo "tox --pre" >> $GITHUB_ENV
```
acopar, line 66: I am not sure this will be working since you removed variable assignment. You are not setting any environment variable with this and after the change the echo would not do anything. The consequence of this not working is that QA deploys would not include prerelease packages.

gregorjerse: Nice catch. In tox 4+ it is recommended to set PIP_PRE environmental variable as --pre is deprecated and could even not work.

acopar (approved):

tox.ini
```diff
@@ -25,14 +25,6 @@ passenv =
         DOCKER_*
         DJANGO_TEST_PROCESSES
 
-commands_pre =
```
dblenkus, line 60: Improve the commit message to clarify that the tox file was changed.

docs/CHANGELOG.rst
```diff
@@ -16,6 +16,7 @@ Added
 
 Changed
 -------
+- Use partial testing when running unit tests
```
dblenkus, line 19: Improve the commit message to clarify when partial testing is used.

.github/workflows/ci.yml
```diff
@@ -23,9 +23,10 @@ jobs:
       TOX_COMMAND: "tox --skip-missing-interpreters false"
       RESOLWE_POSTGRESQL_PORT: 55433
       RESOLWE_REDIS_PORT: 56380
+      RESOLWE_TEST_ONLY_CHANGES_TO: "origin/master"
     strategy:
       matrix:
-        toxenv: ["py311", "py312"]
+        toxenv: ["py311-partial", "py312-partial"]
```
dblenkus, line 28: Why do we run tests with two versions of Python? Most of the tests are for processes and thus don't make any difference, except that they take double the amount of time.

gregorjerse: It's not exactly double time, but certainly double resources.

Not all tests are processes though. Some (variant tests for instance) are viewset tests. 

We could remove the `Python 3.11` tests since we use `Python 3.12` in the production.

docs/CHANGELOG.rst
gregorjerse: Done.

tox.ini
gregorjerse: Done.

dblenkus (approved):

PR #1361, opened 2024-08-08
Implement fusion calling with Arriba

zkuralt:
[REID-2491](https://genialis.atlassian.net/browse/REID-2491)

## Checklist

    before non-breaking changes.
    commit(s).

### Processes ###

  can be performed).
  homepage and publication.

  tests).

### Tests ###

[REID-2491]: https://genialis.atlassian.net/browse/REID-2491?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


resolwe_bio/processes/fusion_calling/arriba.py
```diff
@@ -0,0 +1,132 @@
...
+    version = "1.0.0"
+    category = "RNA-Seq"
+    scheduling_class = SchedulingClass.BATCH
+    entity = {"type": "sample"}
+    requirements = {
+        "expression-engine": "jinja",
+        "executor": {
+            "docker": {"image": "public.ecr.aws/genialis/resolwebio/rnaseq:6.2.0"}
+        },
+        "resources": {
```
jkokosar, line 40: Reduce the processing requirements to 1 thread and 12288 memory (from docs: "Arriba usually consumes less than 10 GB of RAM"). We'll be able to test this limits and potentially increase later, as needed.

resolwe_bio/processes/fusion_calling/arriba.py
```diff
@@ -0,0 +1,132 @@
...
+        if return_code:
+            self.error("Arriba process failed.")
+
+        # Save the output files
+        outputs.fusions = inputs.output_file
+        outputs.discarded_fusions = inputs.discarded_fusions_file
+        outputs.intergenic_fusions = inputs.intergenic_fusions_file
+        outputs.log = "arriba.log"
+
+        self.progress(1.0)
```
jkokosar, line 132: This is not needed, as the state is set to 1 automatically when finished.

resolwe_bio/processes/fusion_calling/arriba.py
```diff
@@ -0,0 +1,132 @@
...
+    name = "Arriba"
+    process_type = "data:genefusions:arriba"
+    version = "1.0.0"
+    category = "RNA-Seq"
+    scheduling_class = SchedulingClass.BATCH
+    entity = {"type": "sample"}
+    requirements = {
+        "expression-engine": "jinja",
+        "executor": {
+            "docker": {"image": "public.ecr.aws/genialis/resolwebio/rnaseq:6.2.0"}
```
jkokosar, line 38: Should be public.ecr.aws/genialis/resolwebio/rnaseq:6.3.0

resolwe_bio/processes/fusion_calling/arriba.py
```diff
@@ -0,0 +1,132 @@
...
+    [Arriba manual](https://arriba.readthedocs.io/en/latest/) and in
+    the [original paper](https://genome.cshlp.org/content/31/3/448)
+    The current version of Arriba is 2.4.0.
+    """
+
+    slug = "arriba"
+    name = "Arriba"
+    process_type = "data:genefusions:arriba"
+    version = "1.0.0"
+    category = "RNA-Seq"
```
jkokosar, line 32: Add a new category for gene fusions

resolwe_bio/processes/fusion_calling/arriba.py
```diff
@@ -0,0 +1,132 @@
...
+            description="Name of the output file for discarded fusions.",
+            default="discarded_fusions.tsv",
+        )
+        gtf_file = DataField(
+            "annotation",
+            label="GTF file",
+            description="Annotation file in GTF format.",
+        )
+        blacklist_file = DataField(
+            "genelist",
```
jkokosar, line 73: This should be an existing Data type. If a dedicated "upload" process that creates such type is not available or is not be introduced, a [general process](https://github.com/genialis/resolwe/blob/master/resolwe/toolkit/processes/files.py) to upload "files" should be used. Be advised that when using this process, there is no built-in file validation, and the uploaded file is stored "as is". The same applies bellow.

resolwe_bio/processes/fusion_calling/arriba.py
```diff
@@ -0,0 +1,132 @@
...
+        """Input fields for Arriba process."""
+
+        bam = DataField("alignment:bam:star", label="Input BAM file from STAR aligner")
+        chim_out_type = StringField(
+            label="Chimeric output type",
+            description="Specify the chimeric output type.",
+            choices=["WithinBam", "SAMSeparateOld"],
+            default="WithinBam",
+        )
+        output_file = StringField(
```
jkokosar, line 57: This can be omitted, as it is best to create output file name based on the sample slug, with `_fusions.tsv` suffix.

resolwe_bio/processes/fusion_calling/arriba.py
zkuralt: Hmm, blacklists are simple .tsv files with listed genomic coordinates (e.g. 14:1234-4321) that should be ignored by Arriba. Since these files are relatively big, the idea is to store them in an S3 bucket and put them into a collection in Expressions. Thus I don't think that anything will be actually uploaded at this point.

resolwe_bio/processes/fusion_calling/arriba.py
```diff
@@ -0,0 +1,131 @@
...
+        """Input fields for Arriba process."""
+
+        bam = DataField("alignment:bam:star", label="Input BAM file from STAR aligner")
+        gtf_file = DataField(
+            data_type="annotation",
+            label="GTF file",
+            description="Annotation file in GTF format.",
+        )
+        genome_file = DataField(
+            data_type="genome",
```
jkokosar, line 69: If the goal is to provide genome FASTA file, those are created by the `upload-fasta-nucl` process with the process type `process_type = "data:seq:nucleotide"`. Here, in this process, you can access such Data objects by specifying `data_type="seq:nucleotide"`.

resolwe_bio/processes/fusion_calling/arriba.py
```diff
@@ -0,0 +1,131 @@
...
+        },
+    }
+    data_name = '{{ bam|name|default("?") }}'
+
+    class Input:
+        """Input fields for Arriba process."""
+
+        bam = DataField("alignment:bam:star", label="Input BAM file from STAR aligner")
+        gtf_file = DataField(
+            data_type="annotation",
```
jkokosar, line 64: This will allow both "GTF" and "GFF3" inputs. To limit to just GTF, use `data_type="annotation:gtf"`

resolwe_bio/processes/fusion_calling/arriba.py
```diff
@@ -0,0 +1,131 @@
...
+
+import pandas as pd
+from plumbum import TEE
+from resolwe.process import (
+    Cmd,
+    DataField,
+    FileField,
+    Process,
+    SchedulingClass,
+)
```
jkokosar, line 11: According to linters, this should be:

```
import pandas as pd
from plumbum import TEE

from resolwe.process import Cmd, DataField, FileField, Process, SchedulingClass
```

resolwe_bio/tests/processes/test_arriba.py
```diff
@@ -0,0 +1,60 @@
...
+
+class ArribaProcessorTestCase(KBBioProcessTestCase):
+    @with_resolwe_host
+    @tag_process("arriba")
+    def test_arriba_fusion_detection(self):
+        input_folder = Path("arriba") / "input"
+        output_folder = Path("arriba") / "output"
+
+        with self.preparation_stage():
+            bam = self.run_process(
```
jkokosar, line 18: @zkuralt @romunov The test needs to be rewritten such that the reads are aligned by STAR. For this, you need to modify the test files such that you extract the reads from the "aligned_samples.bam" and align this data to a suitable STAR index generated by using a matching FASTA file. In this way, also the required `chimeric` output will be created and stored in the STAR-aligned BAM Data object.

resolwe_bio/processes/fusion_calling/arriba.py
```diff
@@ -0,0 +1,126 @@
...
+    gtf = pd.read_csv(gtf_file, sep="\t", header=None, usecols=[0])
+    contigs = set(gtf[0])
+    out = " ".join(list(contigs))
+    return out
+
+
+class Arriba(Process):
+    """Run Arriba for fusion detection from RNA-Seq data.
+
+    Arriba is a command-line tool for gene fusion detection from RNA-Seq data.
```
romunov, line 25: This line is superfluous.

resolwe_bio/processes/fusion_calling/arriba.py
```diff
@@ -0,0 +1,126 @@
...
+    This process detects gene fusions from RNA-Seq data using the Arriba tool.
+    The input can be a BAM file from the STAR aligner, and additional optional
+    inputs such as blacklist, protein domains, and known fusions files can be provided.
+    More information about Arriba can be found in the
+    [Arriba manual](https://arriba.readthedocs.io/en/latest/) and in
+    the [original paper](https://genome.cshlp.org/content/31/3/448)
+    The current version of Arriba is 2.4.0.
```
romunov, line 32: Let's see how long it takes us to have this text about version out of sync with the docker image. 😁

resolwe_bio/processes/fusion_calling/arriba.py
```diff
@@ -0,0 +1,126 @@
...
+            "cores": 1,
+            "memory": 12288,
+        },
+    }
+    data_name = '{{ bam|name|default("?") }}'
+
+    class Input:
+        """Input fields for Arriba process."""
+
+        bam = DataField("alignment:bam:star", label="Input BAM file from STAR aligner")
```
romunov, line 57: Label could go more in depth of what kind of bam file is expected (how it's processed in i.e. STAR).

resolwe_bio/processes/fusion_calling/arriba.py
```diff
@@ -0,0 +1,126 @@
...
+        gtf_file = DataField(
+            data_type="annotation:gtf",
+            label="GTF file",
+            description="Annotation file in GTF format.",
+        )
+        genome_file = DataField(
```
romunov, line 63: Could we, by default, use the one used to align the .bam file?

resolwe_bio/processes/fusion_calling/arriba.py
zkuralt: At least until the next rnaseq docker image update :D

resolwe_bio/processes/fusion_calling/arriba.py
zkuralt: I guess that means the code stays as it is?

romunov: Makes sense in a way, since .bam files could be stand-alone.

jkokosar (commented):
Output test files need to be committed as well. Please add.

jkokosar:
Also, add to changelog and fix linters

jkokosar (approved):

PR #1370, opened 2024-08-30
Add ``samtools-mpileup-single`` process

marcellevstek:
REID-2546

    before non-breaking changes.
    commit(s).


resolwe_bio/processes/samtools/samtools_mpileup.py
```diff
@@ -0,0 +1,346 @@
...
+    GroupField,
+    IntegerField,
+    ListField,
+    Process,
+    SchedulingClass,
+    StringField,
+)
+
+
+def parse_mpileup_entry(row):
```
marcellevstek, line 23: @jkokosar I can add additional tool parameters as input fields, that change the behaviour of the ``base_sequence`` column. For example, excluding the deletions in the output.

resolwe_bio/processes/samtools/samtools_mpileup.py
```diff
@@ -0,0 +1,367 @@
...
+                label="Exclude output ends in sequence column",
+                description="Removes the “^” (with mapping quality) and “$” markup from the sequence column [--no-output-ends].",
+                default=False,
+            )
+            output_base_counts = BooleanField(
+                label="Output per base counts",
+                description="Output base counts at each position. "
+                "This is a custom implementation and not part of the original samtools mpileup. "
+                "It outputs a file with per base counts at each position.",
+                default=False,
```
jkokosar, line 237: Should we make it a `default=True`?

resolwe_bio/processes/samtools/samtools_mpileup.py
```diff
@@ -0,0 +1,367 @@
...
+            if inputs.fasta.output.build != inputs.bam.output.build:
+                self.error(
+                    "Input BAM and FASTA files should have matching genome builds. "
+                    "BAM file has build "
+                    f"{inputs.bam.output.build}, while FASTA file has build "
+                    f"{inputs.fasta.output.build}."
+                )
+            input_options.extend(["-f", inputs.fasta.output.fasta.path])
+
+        if inputs.bedfile:
```
jkokosar, line 282: Begin with the check that either `bedfile` or `positions` is provided. ReSDK does not disable input fields in the same way as the FE can, so it is possible to provide both at the same time.

resolwe_bio/tests/files/samtools/outputs/samtools_in_counts.txt.gz
jkokosar, line 1: Please check the file header. The columns are just enumerated. This is probably not as intended.

resolwe_bio/processes/samtools/samtools_mpileup.py
marcellevstek: As this is a custom implementation, I thought it would be more appropriate to make it false.

resolwe_bio/tests/files/samtools/outputs/samtools_in_counts.txt.gz
marcellevstek: I fixed this locally but have yet to push it.

resolwe_bio/processes/samtools/samtools_mpileup.py
jkokosar: This is additional functionality, that is not changing the tool main output. I see no harm in enabling this by default. In this regard it is more problematic that the field `excl_flags` adds to the default filtering rules by including `UNMAP`. Perhaps its best to explicitly state this in tool help section.

marcellevstek: Will remove UMAP from defaults.

marcellevstek: It is enabled now by default.

jkokosar (approved):

PR #1390, opened 2024-09-25
Add known fusions file input option

zkuralt:
[REID-2567](https://genialis.atlassian.net/browse/REID-2567)

## Checklist

    before non-breaking changes.
    commit(s).

### Processes ###

  can be performed).
  homepage and publication.

  tests).

### Tests ###

[REID-2567]: https://genialis.atlassian.net/browse/REID-2567?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


resolwe_bio/processes/fusion_calling/arriba.py
```diff
@@ -40,7 +40,7 @@ class Arriba(Process):
     slug = "arriba"
     name = "Arriba"
     process_type = "data:genefusions:arriba"
-    version = "1.0.2"
+    version = "1.0.3"
```
jkokosar, line 43: Bump the middle version, to `1.1.0`. Use the same for workflow.

resolwe_bio/processes/fusion_calling/arriba.py
```diff
@@ -83,6 +83,13 @@ class Input:
             required=False,
         )
 
+        known_functions_file = DataField(
```
jkokosar, line 86: Shoud be `known_fusions_file`. Rename the input also below.

resolwe_bio/processes/workflows/gene-fusion-calling-arriba.py
```diff
@@ -246,6 +253,9 @@ def run(self, inputs, outputs):
         if inputs.blacklist_file:
             arriba_input["blacklist_file"] = inputs.blacklist_file.id
 
+        if inputs.known_fusions:
+            arriba_input["known_fusions"] = inputs.known_fusions.id
```
jkokosar, line 257: Should be `arriba_input[known_fusions_file]`. Consider making the variable name the same between the process and the workflow to avoid future confusions.

jkokosar: If possible, add an example fusion file to tests to test the working of this option.

resolwe_bio/processes/fusion_calling/arriba.py
zkuralt: Ah, autocomplete (aka copilot) at its best 🙈

resolwe_bio/processes/fusion_calling/arriba.py
```diff
@@ -40,7 +40,7 @@ class Arriba(Process):
     slug = "arriba"
     name = "Arriba"
     process_type = "data:genefusions:arriba"
-    version = "1.0.2"
+    version = "1.1.2"
```
jkokosar, line 43: Bump just the middle version: `version = "1.1.0"`

jkokosar (commented):
Please fix the arriba version and then squash commits and rebase.

jkokosar (approved):

PR #1409, opened 2024-12-19
Make changes to ``gtf-to-bed`` process

marcellevstek:
    before non-breaking changes.
    commit(s).

### DIFF
--- a/resolwe_bio/processes/support_processors/gtf_to_bed.py
+++ b/resolwe_bio/processes/support_processors/gtf_to_bed.py
@@ -15,7 +15,7 @@ from resolwe.process import (
 
 
 class GTFtoBED(Process):
-    """GTF to BED conversion.
+    """GTF to BED conversion for predefined genes and feature types.
 
     Note that this process only works with ENSEMBL annotations.
     """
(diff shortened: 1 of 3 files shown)


resolwe_bio/processes/support_processors/gtf_to_bed.py
```diff
@@ -23,9 +23,9 @@ class GTFtoBED(Process):
     slug = "gtf-to-bed"
     name = "GTF to BED"
     process_type = "data:bed"
-    version = "1.1.0"
+    version = "1.2.0"
     category = "Other"
-    data_name = "Converted GTF to BED file"
+    data_name = "{{ geneset|name|default('?') }}"
```
jkokosar, line 28: `geneset` is not a required input, thus the name variable will not always be constructed as intended. It is best to find some other informative way to set the Data object name.

marcellevstek: Thanks, looks like I missed it. We should consider making geneset a required input. I don't see a use case when you want to extract the whole exome or transcriptome, as you can obtain those files from other sources.

jkokosar: If this is the intended use case, then please make it so. Make sure to highlight this requirement in the process description as well.

jkokosar (approved):
LGTM. Please squash and rebase when the previous PR is merged.

PR #1419, opened 2025-02-11
Add an output field in alignment-star

marcellevstek:
indicating whether input reads were downsampled with Seqtk

    before non-breaking changes.
    commit(s).


resolwe_bio/processes/alignment/star.py
```diff
@@ -505,6 +505,9 @@ class Output:
         stats = FileField(label="Statistics")
         species = StringField(label="Species")
         build = StringField(label="Build")
+        seqtk_downsampled = BooleanField(
+            label="The input reads were downsampled with Seqtk"
```
marcellevstek, line 509: @jkokosar please comment on how to change this label.

resolwe_bio/processes/support_processors/multiqc.py
```diff
@@ -579,7 +579,10 @@ def run(self, inputs, outputs):
                     )
                 else:
                     count_report = "ReadsPerGene.out.tab.gz"
-                    report = f"{bam_name}.Log.final.out"
+                    if d.output.seqtk_downsampled == True:
```
jkokosar, line 582: This will brake for legacy STAR data objects without the `seqtk_downsampled` output field

resolwe_bio/processes/support_processors/multiqc.py
```diff
@@ -437,7 +437,7 @@ class MultiQC(Process):
     }
     category = "QC"
     data_name = "MultiQC report"
-    version = "1.25.0"
+    version = "1.26.0"
```
jkokosar, line 440: Docker image version will need to be bumped to use the updated MultiQC config file.

resolwe_bio/processes/alignment/star.py
jkokosar: I would simplify the output field name to `downsampled` so that we don't need to change the output field name in the case that the downsampling method changes in the future. The label can then be "BAM is derived from the downsampled input FASTQ file."

resolwe_bio/processes/alignment/star.py
```diff
@@ -59,7 +59,7 @@ class AlignmentStar(Process):
     slug = "alignment-star"
     name = "STAR"
     process_type = "data:alignment:bam:star"
-    version = "5.3.2"
+    version = "5.4.2"
```
jkokosar, line 62: version = "5.4.0"

resolwe_bio/processes/workflows/bbduk_star_featurecounts_qc.py
```diff
@@ -758,6 +758,7 @@ def run(self, inputs, outputs):
                 quantification,
                 alignment_qc_rrna,
                 alignment_qc_globin,
+                alignment_downsampled,
```
jkokosar, line 761: bump the workflow version

jkokosar (approved):

PR #1446, opened 2025-11-18
Add ATAC-Seq beta workflow

marcellevstek:
    before non-breaking changes.
    commit(s).

### DIFF
--- a/resolwe_bio/processes/workflows/atac_seq_beta.py
+++ b/resolwe_bio/processes/workflows/atac_seq_beta.py
@@ -1,0 +1,594 @@
+"""ATAC-Seq workflow."""
+
+from resolwe.process import (
+    BooleanField,
+    Data,
+    DataField,
+    FloatField,
+    GroupField,
+    IntegerField,
+    Process,
+    StringField,
+)
+from resolwe.process.models import Process as BioProcess
+
+
+class WorkflowATACSeq(Process):
+    """Beta ATAC-Seq workflow.
+
+    This ATAC-seq pipeline follows the official ENCODE DCC pipeline with additional modifications.
+
+    First, reads are optionally downsampled and then aligned to a genome using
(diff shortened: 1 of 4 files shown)


resolwe_bio/processes/workflows/atac_seq_beta.py
```diff
@@ -0,0 +1,584 @@
...
+        )
+        promoter = DataField(
+            data_type="bed",
+            label="Promoter regions BED file",
+            required=False,
+            description="BED file containing promoter regions (TSS+-1000 bp for example). "
+            "Needed to get the number of peaks and reads mapped to promoter regions.",
+        )
+
+        class DownsamplingOptions:
```
jkokosar, line 64: Please expose the `two_pass` advanced option for downsampling step that allows the reduction of memory usage for this operation.

resolwe_bio/processes/workflows/atac_seq_beta.py
```diff
@@ -0,0 +1,584 @@
...
+        chipqc = Data.create(
+            process=BioProcess.get_latest(
+                slug="chipqc",
+            ),
+            input=input_chipqc,
+        )
+
+        Data.create(
+            process=BioProcess.get_latest(slug="multiqc"),
+            input={"data": [alignment, macs2, chipqc]},
```
jkokosar: Missing in the MultiQC inputs is the `reads` object (original or downsampled)

resolwe_bio/tests/workflows/test_atac_seq.py
```diff
@@ -5,7 +5,7 @@
 
 
 class AtacSeqWorkflowTestCase(BioProcessTestCase):
-    @tag_process("workflow-atac-seq")
+    @tag_process("workflow-atac-seq-beta")
```
jkokosar: I think we should duplicate the test, so that both the existing and beta workflows are tested using the unit tests.

jkokosar (approved):
LGTM

